PR #4, opened 2022-04-30
feat: add initial ts gazelle plugin

jbedard:
Essentially https://github.com/aspect-build/rules_js/pull/21 without most of the features (no npm, named imports, `@types`, generated files, ...)


gazelle/tests/simple_import_subdir/BUILD.out
```diff
@@ -0,0 +1,9 @@
+load("@aspect_rules_ts//ts:defs.bzl", "ts_project")
+
+ts_project(
+    name = "simple_import_subdir",
+    srcs = [
+        "main.ts",
+        "subdir/lib.ts",
```
alexeagle, line 7: why doesn't this example generate a BUILD file under subdir/ ? Because there's no BUILD.in file?

gazelle/tests/tests_initial/BUILD.out
```diff
@@ -0,0 +1,12 @@
+load("@aspect_rules_ts//ts:defs.bzl", "ts_project")
+
+ts_project(
+    name = "tests_initial",
+    srcs = ["main.ts"],
+)
+
+ts_project(
+    name = "tests_initial_tests",
```
alexeagle, line 9: we should add `testonly = True` when generating a project containing tests

gazelle/logging.go
```diff
@@ -0,0 +1,13 @@
+package gazelle
+
+import "fmt"
+
+const (
+	DEBUG_ENABLED = false
+)
+
+func DEBUG(str string, args ...interface{}) {
```
alexeagle, line 9: isn't there something a bit more go-idiomatic in the stdlib that could do this logging?

alexeagle (commented):
This looks great! I went through all the tests, seems like the right stuff here for a first commit / MVP

gazelle/tests/simple_import_subdir/BUILD.out
jbedard: Currently it will never create a BUILD, only modify existing ones.

I think that is what we'd want? From what I've seen+created it's very common to have ts code spread across folders, things like angular components (ts, css, html) each having their own folder and a parent maybe containing the `@NgModule` and BUILD etc.

Is there a common gazelle convention/norm for *creating* BUILD files?

gazelle/logging.go
jbedard: Yeah I agree. I'll look into it a bit before merging.

gazelle/tests/simple_import_subdir/BUILD.out
alexeagle: Yeah it should work to run gazelle on a project with zero build files. Maybe this is just a quirk of the test runner though? Will try hooking up in a real example.

gazelle/BUILD.bazel
```diff
@@ -20,3 +21,47 @@ gazelle(
     ],
     command = "update-repos",
 )
+
+go_library(
+    name = "gazelle_library",
+    srcs = glob(
+        ["*.go"],
```
f0rmiga, line 28: Why use a glob by hand here? Just use the Gazelle extension for Go. Check how we have it setup in the rules_python repository.

gazelle/config.go
```diff
@@ -0,0 +1,253 @@
...
+	// to `$package_name$_my_lib` would render to `foo_my_lib`.
+	LibraryNamingConvention = "ts_project_naming_convention"
+	// TestsNamingConvention represents the directive that controls the ts_project test
+	// naming convention. See ts_project_naming_convention for more info on
+	// the package name interpolation.
+	TestsNamingConvention = "ts_tests_naming_convention"
+	// The glob for source files, excludes files matching TestsFileGlob
+	SourcesFileGlob = "ts_srcs_file_glob"
+	// The glob for test files
+	TestsFileGlob = "ts_tests_file_glob"
```
f0rmiga, line 36:
```suggestion
	// The glob for source files, excludes files matching TestsFileGlob.
	SourcesFileGlob = "ts_srcs_file_glob"
	// The glob for test files.
	TestsFileGlob = "ts_tests_file_glob"
```

f0rmiga: Please, use punctuation in the comments here and elsewhere.

gazelle/config.go
```diff
@@ -0,0 +1,253 @@
...
+	c.testsNamingConvention = testsNamingConvention
+}
+
+// RenderTestName returns the ts_project test target name by performing all
+// substitutions.
+func (c *TypeScriptConfig) RenderTestsLibraryName(packageName string) string {
+	return strings.ReplaceAll(c.testsNamingConvention, packageNameNamingConventionSubstitution, packageName)
+}
+
+func (c *TypeScriptConfig) SetSourceFileGlob(srcsFileGlob string) {
```
f0rmiga, line 213: Missing Godoc here and below for exported methods.

gazelle/config.go
```diff
@@ -0,0 +1,253 @@
...
+		return false
+	}
+	if c.srcsFileGlob == "" {
+		return true
+	}
+
+	m, e := doublestar.Match(c.srcsFileGlob, filePath)
+
+	if e != nil {
+		fmt.Println("ERROR: ", fmt.Errorf("srcs file glob error %e", e))
```
f0rmiga, line 227:
```suggestion
	m, err := doublestar.Match(c.srcsFileGlob, filePath)
	if err != nil {
		fmt.Printf("ERROR: srcs file glob error: %v\n", err)
```

f0rmiga: Interesting that you had `%e` for error, but it's meant for scientific notation of floating-point numbers.
https://pkg.go.dev/fmt

f0rmiga: Also, this may beg a bigger question: do you care about the error at all? From the `Match` comment:
```
// Match requires pattern to match all of name, not just a substring.
// The only possible returned error is ErrBadPattern, when pattern
// is malformed.
```

gazelle/config.go
```diff
@@ -0,0 +1,253 @@
...
+		return false
+	}
+	if c.testsFileGlob == "" {
+		return false
+	}
+
+	m, e := doublestar.Match(c.testsFileGlob, filePath)
+
+	if e != nil {
+		fmt.Println("ERROR: ", fmt.Errorf("tests file glob error %e", e))
```
f0rmiga, line 248: Same thing as above.

gazelle/generate.go
```diff
@@ -0,0 +1,309 @@
...
+
+	"github.com/bazelbuild/bazel-gazelle/label"
+	"github.com/bazelbuild/bazel-gazelle/language"
+	"github.com/bazelbuild/bazel-gazelle/rule"
+	"github.com/emirpasic/gods/lists/singlylinkedlist"
+	"github.com/emirpasic/gods/sets/treeset"
+)
+
+const (
+	// The filename (with any of the TS extensions) imported when importing a directory
```
f0rmiga, line 17:
```suggestion
	// The filename (with any of the TS extensions) imported when importing a directory.
```

gazelle/logging.go
f0rmiga: The first `DEBUG` I spotted above brought me here immediately. You can use either https://github.com/golang/glog or https://github.com/sirupsen/logrus.

gazelle/generate.go
```diff
@@ -0,0 +1,309 @@
...
+	// Build the GenerateResult with src and test rules
+	var result language.GenerateResult
+
+	addProjectRule(
+		cfg,
+		args,
+		cfg.RenderLibraryName(filepath.Base(args.Dir)),
+		libSourceFiles,
+		dataFiles,
+		&result,
```
f0rmiga, line 72: This is an OO anti-pattern. You should not modify an argument passed to a function/method.

gazelle/generate.go
```diff
@@ -0,0 +1,309 @@
...
+}
+
+func addProjectRule(cfg *TypeScriptConfig, args language.GenerateArgs, targetName string, sourceFiles, dataFiles *treeset.Set, result *language.GenerateResult) {
+	// Generate nothing if there are no source files
+	if sourceFiles.Empty() {
+		// If there is already a BUILD then potentially clean it up
+		if args.File != nil {
+			// Remove any exiting instance of this rule project rule
+			for _, r := range args.File.Rules {
+				if r.Name() == targetName && r.Kind() == tsProjectKind {
```
f0rmiga, line 94: Not sure cleaning up is the right thing to do...

gazelle/generate.go
```diff
@@ -0,0 +1,309 @@
...
+			collisionErrors.Add(err)
+		}
+	}
+
+	if !collisionErrors.Empty() {
+		it := collisionErrors.Iterator()
+		for it.Next() {
+			log.Printf("ERROR: %v\n", it.Value())
+		}
+		os.Exit(1)
```
f0rmiga, line 261: Busted... you should not interrupt the execution of a program in the middle of nowhere. Please, forward the error until when you have control. If the final function/method doesn't allow you to return the error, then interrupt the execution there.

gazelle/generate.go
```diff
@@ -0,0 +1,309 @@
...
+	importedFiles := treeset.NewWith(importStatementComparator)
+
+	// TODO(jbedard): parse files concurrently
+	sourceFileIt := sourceFiles.Iterator()
+	for sourceFileIt.Next() {
+		filePath := sourceFileIt.Value().(string)
+		fileImports, err := parseFile(filepath.Join(args.Dir, filePath))
+
+		if err != nil {
+			fmt.Println("Parse Error:", fmt.Errorf("%q: %v", filePath, err))
```
f0rmiga, line 121: A parsing error is non-fatal?

gazelle/kinds.go
```diff
@@ -0,0 +1,48 @@
...
+// Loads returns .bzl files and symbols they define. Every rule generated by
+// GenerateRules, now or in the past, should be loadable from one of these
+// files.
+func (ts *TypeScript) Loads() []rule.LoadInfo {
+	return tsLoads
+}
+
+var tsLoads = []rule.LoadInfo{
+	{
+		// TODO: @aspect_rules_ts is a variable. Get from a flag?
```
f0rmiga, line 42: You can see if you can get this from a directive. Not sure if the directives get processed before the `Loads()` method gets called.

gazelle/parser.go
```diff
@@ -0,0 +1,88 @@
+package gazelle
+
+import (
+	"log"
+
+	"path/filepath"
```
f0rmiga, line 6:
```suggestion
	"log"
	"path/filepath"
```

gazelle/parser.go
```diff
@@ -0,0 +1,88 @@
...
+
+	"github.com/evanw/esbuild/pkg/api"
+)
+
+type Parser struct {
```
f0rmiga, line 11: Missing Godoc for exported types.

gazelle/parser.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// ParseImports returns all the imports from a file
+// after parsing it.
+func (p *Parser) ParseImports(filePath, source string) []string {
+	imports := []string{}
+
+	// Construct an esbuild plugin that pulls out all the imports.
+	plugin := api.Plugin{
+		Name: "GetImports",
```
f0rmiga, line 44: `GetImports` is not a const somewhere upstream?

gazelle/parser.go
```diff
@@ -0,0 +1,88 @@
...
+		// Must set bundle to true so that esbuild actually does resolutions.
+		Bundle: true,
+		// Must include unused imports that would normally be tree-shacken
+		IgnoreAnnotations: true,
+	}
+	result := api.Build(options)
+
+	for _, e := range result.Errors {
+		// Inform users that some files couldn't be fully parsed.
+		// No need to crash the program though.
```
f0rmiga, line 83: I think the reasoning for parsing errors not being fatal should contain better reasoning other than `No need to crash the program`.

gazelle/resolve.go
```diff
@@ -0,0 +1,183 @@
...
+	deps := ts.ResolveModuleDeps(c, ix, modulesRaw.(*treeset.Set), from)
+
+	DEBUG("RESOLVED(%s): %s => %s", from.Name, modulesRaw.(*treeset.Set).Values(), deps.Values())
+
+	if !deps.Empty() {
+		r.SetAttr("deps", convertDependencySetToExpr(deps))
+	}
+}
+
+func (ts *TypeScript) ResolveModuleDeps(
```
f0rmiga, line 88: Godoc

gazelle/resolve.go
```diff
@@ -0,0 +1,183 @@
...
+					"\t2. Ignore the dependency with a comment '# gazelle:%[3]s %[1]s' in the BUILD file",
+				mod.Path, mod.SourcePath, IgnoreImportsDirective,
+			)
+			log.Printf("ERROR: Failed to validate dependencies for target %q. %v", from.String(), err)
+			hasFatalError = true
+		}
+	}
+
+	if hasFatalError {
+		os.Exit(1)
```
f0rmiga, line 157: Same as the comment before re interrupting the program.

f0rmiga (changes requested):

gazelle/logging.go
jbedard: I'm just deleting this for now, it was basically just debug junk

gazelle/BUILD.bazel
jbedard: Done 👍

gazelle/config.go
jbedard: > Also, this may beg a bigger question: do you care about the error at all? From the Match comment:

This means the user (developer) set an invalid directive so I think outputting an error and failing is a clear indication to the user. Where simply ignoring the error would add more files then necessary to the BUILD and could easily go unnoticed by the dev?

gazelle/parser.go
jbedard: I think it's just for debugging essentially. I assume esbuild will put this in logs if there is an error or plugins are listed somewhere etc. I think @alexeagle originally wrote this part.

gazelle/parser.go
jbedard: What is the normal approach when running gazelle on invalid source code which can't be parsed? What does python do?

gazelle/resolve.go
jbedard: Since we are just a gazelle-plugin I guess this means we are killing gazelle... do you know if there's a standard way to indicate an error to gazelle?

gazelle/generate.go
jbedard: I'm pushing it up a single function call to `GenerateRules` which is the entry into our gazelle plugin (if I'm remembering+understanding that correctly).

gazelle/resolve.go
jbedard: I'm moving this up to the plugin entrypoint `Resolve()` method

gazelle/resolve.go
jbedard: I've made this private for now

gazelle/parser.go
jbedard: I moved it up a few calls to the plugin entry point

gazelle/generate.go
jbedard: I've changed this to never remove a rule in case the user added `# keep`, but we won't generate a new rule unless we found source files.

gazelle/generate.go
jbedard: I've moved all `os.Exit` calls up to the gazelle plugin-API entry points

gazelle/generate.go
jbedard: We don't want to to `os.Exit` just because of a single file with bad syntax, just output an error/warning and continue generating BUILDs

gazelle/generate.go
jbedard: I've kept this for now. The `result` has multiple things in it and is passed into this method multiple times (one for source, one for tests etc).

aptenodytes-forsteri:
Why did this close?

jbedard:
I believe this was moved to https://github.com/aspect-build/gazelle-extensions/

aptenodytes-forsteri:
> I believe this was moved to https://github.com/aspect-build/gazelle-extensions/

Hmm that 404's for me. Is that a private repo?

alexeagle:
Yes, we haven't open-sourced that yet. We need to decide how gazelle extensions can help us pay our engineers. FWIW, the extension in that private repo isn't finished yet either and we're not currently working on it.

dudicoco:
what's the current status of a gazelle extension?
I think js/ts rules are not really user friendly for devs without a gazelle extension - no one wants to maintain build files.

jbedard:
The extension is part of the [aspect cli](https://www.aspect.build/cli), or you can checkout the extension itself in the repo if you want to setup gazelle yourself: https://github.com/aspect-build/aspect-cli/tree/main/gazelle/js

dudicoco:
@jbedard is there a documentation on the implementation of these options?

jbedard:
The CLI docs list them: https://github.com/aspect-build/aspect-cli/blob/main/docs/help/topics/directives.md

That should probably also be listed on the docs site 🤔

PR #5, opened 2022-05-02
feat: make typescript version selection easier

alexeagle:
- read from package.json in the typical case
- mirror version info from the registry so you can just give us a version number
- advanced users can still declare their own npm_typescript repo to bypass all of it


ts/repositories.bzl
```diff
@@ -6,24 +6,85 @@ See https://docs.bazel.build/versions/main/skylark/deploying.html#dependencies
...
     bazel_lib = "0.9.7",
     rules_nodejs = "5.4.0",
     rules_js = "1e97acf805b11d7b8716a5c2f4d4466f57d3fb3d",
-    # Users can easily override the typescript version just by declaring their own http_archive
-    # named "npm_typescript" prior to calling rules_js_dependencies.
-    typescript = "4.6.3",
 )
 
-# WARNING: any changes in this function may be BREAKING CHANGES for users
+LATEST_TS_VERSION = TS_VERSIONS.keys()[-1]
```
jbedard, line 17: I thought order in maps was not guaranteed? Is that not true for constants? Or not for `.keys()`?

ts/repositories.bzl
```diff
@@ -6,24 +6,85 @@ See https://docs.bazel.build/versions/main/skylark/deploying.html#dependencies
...
+    Args:
+        ts_version_from_package_json: path to a package.json file which declares an exact typescript version.
+            The version to fetch is read from the file.
+        ts_version: version of the TypeScript compiler.
+            Overrides ts_version_from_package_json.
+            Only needed if ts_version_from_package_json is not set,
+            or the version of typescript in package.json is a semver range.
+        ts_integrity: integrity hash for the npm package.
+            By default, uses values mirrored into rules_ts.
+            You can simply fetch this from the registry:
```
jbedard, line 82: Should we say "for example" or something since you can't just copy and paste this?

ts/repositories.bzl
```diff
@@ -6,24 +6,85 @@ See https://docs.bazel.build/versions/main/skylark/deploying.html#dependencies
...
+            `curl --silent https://registry.npmjs.org/typescript/4.6.3 | jq -r '.dist.integrity'`
+    """
+
+    if not ts_version_from_package_json and not ts_version:
```
jbedard, line 86: Should ensure they aren't both set? Would go allow `(not x) != (not y)`?

ts/repositories.bzl
```diff
@@ -6,24 +6,85 @@ See https://docs.bazel.build/versions/main/skylark/deploying.html#dependencies
...
 # and released only in semver majors.
-def rules_ts_dependencies():
-    "Dependencies needed by users of rules_ts"
+def rules_ts_dependencies(ts_version_from_package_json = None, ts_version = None, ts_integrity = None):
+    """Dependencies needed by users of rules_ts.
+
+    To skip fetching the typescript package, define repository called 'npm_typescript' before calling this.
+
+    Args:
+        ts_version_from_package_json: path to a package.json file which declares an exact typescript version.
```
jbedard, line 74: This seems like a weird name to me, with "from"...

ts/repositories.bzl
alexeagle: https://docs.bazel.build/versions/main/skylark/lib/dict.html
> Dictionaries are iterable; iteration yields the sequence of keys in insertion order.
so yeah, I'm pretty confident this gives the last key.

ts/repositories.bzl
alexeagle: do you have a different suggestion? I had just `package_json` but then I found myself wanting to comment on the usage site with "This is only used to get the version of typescript"

ts/repositories.bzl
jbedard: Oh, I was looking at go maps 🤦

ts/repositories.bzl
alexeagle: yeah good point

ts/repositories.bzl
alexeagle: hmm yeah it would probably be a mistake to specify both

jbedard: Actually it looks like specifying both is valid.. so nm?

ts/repositories.bzl
jbedard: Suggesting a name is a lot harder then complaining about a name 😅 

Anything I can think of has the same issue you mentioned (you need to add a comment saying "all we want is the version").

And just dropping the "from" makes the name more confusing?

I figured it's worth thinking about even if we leave it as is... 🤷

ts/repositories.bzl
jbedard: It seems like `ts_version` overrides the package.json, especially if the package.json doesn't have an explicit version (and we don't want to handle `^4.0`?).

ts/repositories.bzl
alexeagle: it's also more clear this way that `ts_version` and `ts_version_from_package_json` are mutually exclusive?
I added comments about that as well.

ts/repositories.bzl
alexeagle: I don't know what we would do with a semver range. Our goal is to have the same version. We'd need to parse some yarn.lock or package-lock.json or whatever if we want to know what version the editor will actually use.

I made them mutually exclusive.

ts/repositories.bzl
alexeagle: the other option is only one string-typed attribute `ts_version` with some special value, like

```
ts_version = "4.6.1"
ts_version = "from=//:package_json"
```
but that's inventing syntax and maybe less buildozer-readable

jbedard: Yeah I though of some magic syntax as well but that's not worth it.

> it's also more clear this way that ts_version and ts_version_from_package_json are mutually exclusive?
Those names do make it clearer 👍

jbedard (approved):
idk if others want to take a look but lgmt

ts/repositories.bzl
gregmagolan:
```
ts_version = "4.6.1"
ts_version = "//:package_json"
```
you could detect if the string is a "version", if not assume it is a label

gregmagolan: actually, it should be sufficient to check if the first char is a number since labels can't start with numbers

gregmagolan: you can also run into trouble if the version in package json is non-trivial
- it could be of the form `1.x.x` or `^1` or `~1.2`
- it could be in dependencies or devDependencies
- it could be resolved via "resolutions"
- it could be missing altogether defined as a transitive

gregmagolan: tho I suppose you'd just error out in those cases

jbedard: > you could detect if the string is a "version", if not assume it is a label

Do we want to avoid string attributes that might be labels though? That always feels odd to me, then bazel can't verify it up front, IDEs don't understand it etc

alexeagle: We'll do this in more rulesets like rules_terser etc. so I think it's worth getting what we want.

Advantage of two attributes:
- more machine-editable, you can tell whether a version or a label is specified
- error messages can say "I don't understand your version so please set the xx attribute"

We could make the second attribute name shorter:
```
ts_version = "4.6.1"
ts_version_from = "//:package.json"
```

gregmagolan: I like the short name, it reads well

gregmagolan: Suppose if we specify both we fail out?

gregmagolan (approved):
🌮

ts/repositories.bzl
alexeagle: Yes, exactly one of the two is required.

PR #8, opened 2022-05-06
feat: implement worker mode

thesayyn:


ts/defs.bzl
```diff
@@ -165,11 +166,10 @@ def ts_project(
         buildinfo_out = tsbuildinfo_path if composite or incremental else None,
         emit_declaration_only = emit_declaration_only,
         tsc = tsc,
+        tsc_worker = tsc_worker,
         # TODO: support transpiler attribute when we have a js_library equivalent
         # and can copy over the logic from rules_nodejs to set tsc_js_outs, ts_map_outs
         # and create all the targets for transpilation
         transpile = True,
-        # We don't support this feature at all from rules_nodejs yet
-        supports_workers = False,
```
alexeagle, line 173: I think we need to add the attribute above with all the others so users can pass True/False?

ts/private/ts_project.bzl
```diff
@@ -176,12 +178,12 @@ This is an error because Bazel does not run actions unless their outputs are nee
             run_action(
                 ctx,
                 link_workspace_root = ctx.attr.link_workspace_root,
-                executable = "tsc",
+                executable = executable,
```
alexeagle, line 181: oh this needs some cleanup, we still have the code from rules_nodejs which is no longer applicable based on rules_js.

alexeagle (commented):
why is this still a draft - what's left to merge it? (if it's the cancellation feature, can we defer that to a later PR?)

ts/defs.bzl
thesayyn: Not really. since we don't have an explicit parameter for it, it'll be passed via kwargs. all good

ts/private/ts_project.bzl
thesayyn: Thanks for doing this 😊

thesayyn:
I'll keep this PR just about worker mode to make it easier to review. I'll follow up with more PRs by adding tests and support for keeping cache per target to make it more scalable

package.json
```diff
@@ -1,6 +1,7 @@
 {
     "private": true,
     "dependencies": {
-        "typescript": "4.6.3"
+        "typescript": "4.6.3",
+        "@bazel/worker": "^5.4.2"
```
thesayyn, line 5: This might be unnecessary given that this file is only used to pick the typescript version.

alexeagle:
Is this ready for review and merge now?

thesayyn:
> Is this ready for review and merge now?

Yes, this is ready to be reviewed and landed if there are no review changes.  I will flip the worker mode as default in my emit result cache PR.

ts/private/ts_project.bzl
```diff
@@ -44,6 +45,7 @@ def _ts_project_impl(ctx):
         execution_requirements["supports-workers"] = "1"
         execution_requirements["worker-key-mnemonic"] = "TsProject"
         progress_prefix = "Compiling TypeScript project (worker mode)"
```
gregmagolan, line 47: Seems wrong to hard-code "(worker mode)" since a worker will only be used if the a worker spawn strategy is used. All we're saying here is that the rule support workers.

ts/private/ts_project_worker.js
```diff
@@ -0,0 +1,168 @@
...
+
+    return succeded
+}
+
+function main() {
+    if (worker.runAsWorker(process.argv)) {
+        worker.log(`Running ${MNEMONIC} as a Bazel worker`)
+        worker.runWorkerLoop(emit)
+    } else {
+        worker.log(`Running ${MNEMONIC} as a standalone process`)
```
gregmagolan, line 159: This code path is broken looks like. Something to do with parsing the args from the params file:
```
bazel build //examples/worker:ts --spawn_strategy=darwin-sandbox

ERROR: /Users/greg/aspect/rules/js/rules_ts/examples/worker/BUILD.bazel:14:11: Compiling TypeScript project (worker mode) //examples/worker:ts [tsc -p examples/worker/tsconfig.json] failed: (Exit 1): _tsc_worker_launcher.sh failed: error executing command bazel-out/darwin-opt-exec-2B5CBBC6/bin/external/npm_typescript/_tsc_worker_launcher.sh @bazel-out/darwin-fastbuild/bin/examples/worker/foo.js-0.params


Use --sandbox_debug to see verbose messages from the sandbox
Running TsProject as a standalone process
Started a new process to perform this action. Your build might be misconfigured, try 
 build --strategy=TsProject=worker
node:internal/fs/utils:344
    throw err;
    ^

Error: ENOENT: no such file or directory, open 'bazel-out/darwin-fastbuild/bin/examples/worker/foo.js-0.params'
    at Object.openSync (node:fs:585:3)
    at Object.readFileSync (node:fs:453:35)
```

ts/private/ts_project_worker.js
```diff
@@ -0,0 +1,168 @@
...
+        )
+
+        lastRequestArgHash = args.join(' ')
+        program = ts.createWatchProgram(host)
+    }
+
+    return program
+}
+
+async function emit(args, inputs, once = false) {
```
gregmagolan, line 91: once is not used but guessing it should be so we can do minimal amount of work if the spawn strategy is not worker and this process will be thrown away

ts/private/ts_project_worker.js
```diff
@@ -0,0 +1,168 @@
...
+        const host = ts.createWatchCompilerHost(
+            tsconfigPath,
+            parsedArgs.options,
+            sys,
+            ts.createSemanticDiagnosticsBuilderProgram,
+            noop,
+            noop
+        )
+
+        lastRequestArgHash = args.join(' ')
```
gregmagolan, line 84: should we also add the contents of the tsconfig to the hash so that the program is re-created whenever tsconfig changes? suppose to be correct you'd also need the contents of any transitive tsconfigs that are extended from?

ts/private/ts_project.bzl
thesayyn: Agreed.

ts/private/ts_project_worker.js
thesayyn: This is not relevant anymore. It is completely different on #26

thesayyn: It didn't make sense to use `tsc --watch` for standalone processes as you would be paying the cost of setting up watchers, cache etc..

ts/private/ts_project_worker.js
```diff
@@ -0,0 +1,168 @@
...
+               return setImmediate(callback)
+            },
+            clearTimeout: clearImmediate,
+            watchFile: (filePath, callback) => {
+                const relativeFilePath = path.relative(execRoot, filePath)
+                invalidate.set(relativeFilePath, (kind) => callback(filePath, kind))
+                return {close: () => invalidate.delete(relativeFilePath)}
+            },
+            watchDirectory: (path, callback, recursive, options) => {
+                // TODO: Figure out what to do with directory watchers
```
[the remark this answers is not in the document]
thesayyn: I have this figured out in #26

ts/private/ts_project_worker.js
thesayyn: Also, this logic has been changed as well in #26

thesayyn:
@gregmagolan i wonder if we could discard this PR and continue over at #26

thesayyn:
superseded by https://github.com/aspect-build/rules_ts/pull/26

Issue #15, opened 2022-05-13
Incompatible with rules_js@0.6.2


jfirebaugh:
```
$ bazel run @npm_typescript//:tsc -- --version
ERROR: Skipping '@npm_typescript//:tsc': error loading package '@npm_typescript//': cannot load '@aspect_rules_js//js:node_package.bzl': no such file
```

rules_ts has:

https://github.com/aspect-build/rules_ts/blob/a866c0e9e577ad3ac27961526a4f7df4ec6330e5/ts/BUILD.typescript#L3

But `node_package` went through a series of renames, and no longer exists.

PR #16, opened 2022-05-13
update rules_js

alexeagle:


thesayyn:
Ah. I already had this already covered in #8. Did you mean to land it separately?

alexeagle:
yes I totally borrowed your fixes from there, we do want to land this separately because we can :)

alexeagle:
because it fixes https://github.com/aspect-build/rules_ts/issues/15

thesayyn (approved):

PR #21, opened 2022-05-18
add example coverage for using 1p/3p deps with types

alexeagle:


alexeagle:
@jfirebaugh are you okay with signing our CLA for your commit I've included here? I want to give credit/attribution :)

examples/simple/tsconfig.json
```diff
@@ -1,5 +1,10 @@
 {
     "compilerOptions": {
-        "declaration": true
+        "declaration": true,
+        "baseUrl": ".",
+        "paths": {
+            "@myorg/*": ["../*"]
```
thesayyn, line 6: It feels like this shouldn't be needed since tsc will find it inside node_modules.

examples/pnpm-lock.yaml
```diff
@@ -0,0 +1,26 @@
+lockfileVersion: 5.3
```
thesayyn, line 1: nit: we should tell prettier give up on formatting lock files.

thesayyn (approved):

examples/pnpm-lock.yaml
alexeagle: agreed, we've done that in some other repos

examples/simple/tsconfig.json
thesayyn: Oh. its probably for the IDE to locate these properly/

jbedard (approved):
😄 

EDIT: I was only looking at the fix commit

examples/simple/tsconfig.json
alexeagle: it's needed for the editor to avoid red squiggles. I'll add a comment

PR #24, opened 2022-05-19
Example with ts_project depending on another ts_project

jfirebaugh:
Is this the right way to have a `ts_project` depend on another `ts_project`? It seems like it should work, but:

```
$ bazel build //examples/ts_project_dep/b:ts
ERROR: /Users/john/figma/rules_ts/examples/ts_project_dep/b/BUILD.bazel:3:11: in _ts_project rule //examples/ts_project_dep/b:ts: 
Traceback (most recent call last):
	File "/Users/john/figma/rules_ts/ts/private/ts_project.bzl", line 156, column 47, in _ts_project_impl
		inputs = copy_files_to_bin_actions(ctx, inputs),
	File "/private/var/tmp/_bazel_johnfirebaugh/21a27cee0c79a1509a094219b285fcfa/external/aspect_bazel_lib/lib/private/copy_to_bin.bzl", line 59, column 36, in copy_files_to_bin_actions
		return [copy_file_to_bin_action(ctx, file, is_windows = is_windows) for file in files]
	File "/private/var/tmp/_bazel_johnfirebaugh/21a27cee0c79a1509a094219b285fcfa/external/aspect_bazel_lib/lib/private/copy_to_bin.bzl", line 38, column 35, in copy_file_to_bin_action
		dst = ctx.actions.declare_file(file.basename, sibling = file)
Error in declare_file: the output artifact 'examples/ts_project_dep/a/tsconfig.json' is not under package directory 'examples/ts_project_dep/b' for target '//examples/ts_project_dep/b:ts'
```


alexeagle:
Could you try fixing this so I can push commits to your branches? It will be a lot faster to iterate through these if you send a red PR and I can add a commit making it green and then land.

I think you either need to change settings on your figma org, or send these PRs from a personal fork. Does that work?

alexeagle:
https://github.com/figma/rules_ts/pull/1 has the fix

alexeagle (approved):

PR #25, opened 2022-05-26
fix: set declaration dir to "." when it is empty

thesayyn:
ts_project fails when it is placed at workspace root with `declaration=True` due to empty declaration dir argument.


thesayyn:
This is identical to `outdir` logic we have right above the lines of the changes.

alexeagle (approved):

PR #26, opened 2022-05-27
feat: add worker support to ts_project

thesayyn:
Differences from #8 

- No more `args` hash to find a suitable worker for a given target.
- Keep a warm worker for each target (tsconfig)
- Every `ts.Program` instance has only access to its inputs. 
- Keeping a cache of emitted files to shorten emit time. `createEmitCacheAndDiagnosticsProgram`


thesayyn:
The only quirk that I still have to work on is when you change tsconfig.ts and remove a file from `srcs` worker fails. Apart from that ready to land.

e2e/worker/.bazelrc
```diff
@@ -0,0 +1 @@
+build:verbose --define=VERBOSE_LOGS=1
```
gregmagolan, line 1: VERBOSE_LOGS?

e2e/worker/test.sh
```diff
@@ -0,0 +1,68 @@
...
+
+# just run it in case there is leftovers
+cleanup
+# set up trap for graceful cleanup
+trap cleanup EXIT
+
+
+message "# Case 0; try to build a target that will never succeed"
+message="error TS2322: Type 'number' is not assignable to type 'string'."
+bazel build :should_fail 2>&1 | grep "$message" || exit_with_message "Case 0: expected worker to report \"$message\" but it didn't"
```
jbedard, line 29: I think we can drop the "but it didn't"? Maybe put "FAIL: " at the start?

ts/BUILD.typescript
```diff
@@ -17,19 +18,65 @@ link_npm_package(
     visibility = ["//visibility:public"],
 )
 
+npm_package(
```
jbedard, line 21: These are all worker dependencies? Should we put them down below under ` # tsc worker`?

ts/BUILD.typescript
```diff
@@ -17,19 +18,65 @@ link_npm_package(
...
+
+
+
+# tsc
 directory_path(
     name = "tsc_entrypoint",
     directory = ":node_modules/typescript/dir",
     path = "bin/tsc",
 )
-
```
jbedard, line 25: Undo? 🤷

ts/private/ts_project_worker.js
```diff
@@ -0,0 +1,384 @@
...
+}
+
+function main() {
+    if (worker.runAsWorker(process.argv)) {
+        worker.log(`Running ${MNEMONIC} as a Bazel worker`)
+        worker.runWorkerLoop(emit)
+    } else {
+        worker.log(`WARNING: Running ${MNEMONIC} as a standalone process`)
+        worker.log(`Your build might be misconfigured, try putting "build --strategy=${MNEMONIC}=worker" into your .bazelrc or add "supports_workers = False" attribute into this ts_project target.`)
+        worker.log(`Started a standalone process to perform this action but this might lead to some unexpected behavior with tsc due to being run non-sandboxed.`)
```
jbedard, line 374: Should this be one line up before the suggestion?

ts/private/ts_project_worker.js
```diff
@@ -0,0 +1,384 @@
...
+}
+function timingEnd(label) {
+    debuglog(`${label} ${performance.now() - timing[label]}ms`)
+}
+
+
+function getArgsFromParamFile() {
+    let paramFilePath = process.argv.pop()
+    if (paramFilePath.startsWith('@')) {
+        paramFilePath = path.resolve("..", "..", "..", paramFilePath.slice(1))
```
jbedard, line 26: Why 3 `..`? Worth a comment?

ts/private/ts_project_worker.js
```diff
@@ -0,0 +1,384 @@
...
+function timingStart(label) {
+    timing[label] = performance.now();
+}
+function timingEnd(label) {
+    debuglog(`${label} ${performance.now() - timing[label]}ms`)
+}
+
+
+function getArgsFromParamFile() {
+    let paramFilePath = process.argv.pop()
```
jbedard, line 24: Why `pop()` and modify the `argv`?

.github/workflows/ci.yaml
```diff
@@ -51,3 +52,7 @@ jobs:
                   XDG_CACHE_HOME: ~/.cache/bazel-repo
               working-directory: ${{ matrix.folder }}
               run: bazel --bazelrc=$GITHUB_WORKSPACE/.github/workflows/ci.bazelrc --bazelrc=.bazelrc test //...
+            - name: ./test.sh (e2e/worker)
```
gregmagolan, line 55: hmm; can you make this generic so it runs test.sh if it exists in the directory instead of making it specific to e2e/worker?

ts/repositories.bzl
```diff
@@ -112,6 +127,7 @@ def rules_ts_dependencies(ts_version_from = None, ts_version = None, ts_integrit
         url = "https://github.com/aspect-build/rules_js/archive/refs/tags/v0.10.0.tar.gz",
     )
 
+
```
jbedard, line 130: Undo?

ts/BUILD.package
```diff
@@ -0,0 +1 @@
+exports_files(["package"])
```
jbedard, line 1: What is the "package" file?

ts/private/ts_project_worker.js
```diff
@@ -0,0 +1,384 @@
...
+        worker.log(`Your build might be misconfigured, try putting "build --strategy=${MNEMONIC}=worker" into your .bazelrc or add "supports_workers = False" attribute into this ts_project target.`)
+        worker.log(`Started a standalone process to perform this action but this might lead to some unexpected behavior with tsc due to being run non-sandboxed.`)
+        const args = getArgsFromParamFile()
+        if (!emitOnce(args)) {
+            process.exit(1)
+        }
+    }
+}
+
+if (require.main === module) {
```
jbedard, line 382: When would this not be true?

ts/private/ts_project_worker.js
```diff
@@ -0,0 +1,384 @@
...
+
+    const directoryWatchers = new Map();
+    const fileWatchers = new Map();
+
+    const knownInputs = new Set(initialInputs);
+
+    /** @type {ts.System} */
+    const sys = {
+        ...ts.sys,
+        write: worker.log,
```
jbedard, line 112: Passing `worker.log` assumes `log` doesn't require any context. Should we put a wrapper (or `.bind`) at the top of the file?

gregmagolan (approved):
🧙

ts/private/ts_project_worker.js
jbedard, line 112: Can this be a constant put outside the function?

`const SYS_OVERRIDES = { ... }` or something like that, then `const sys = {...ts.sys, ...SYS_OVERRIDES}` here?

ts/private/ts_project_worker.js
```diff
@@ -0,0 +1,384 @@
...
+        parsedArgs.options,
+        sys,
+        createEmitCacheAndDiagnosticsProgram,
+        noop,
+        noop
+    )
+
+    host.invalidate = invalidate;
+
+    program = ts.createWatchProgram(host)
```
jbedard, line 137: Where is `program` defined?

ts/private/ts_project_worker.js
```diff
@@ -0,0 +1,384 @@
+const fs = require('fs')
```
jbedard, line 1: Was there a reason for cjs instead of es6? We're using es6 throughout the file...

ts/private/ts_project_worker.js
```diff
@@ -0,0 +1,384 @@
...
+    function writeFile(path, data, writeByteOrderMark) {
+        debuglog(`writeFile ${path}`);
+        ts.sys.writeFile(path, data, writeByteOrderMark);
+    }
+
+    function readFile(filePath, encoding) {
+        debuglog(`readFile ${filePath}`);
+        const relative = path.relative(execRoot, filePath)
+        /** if it is under node_modules just allow file reads as we don't have a list of deps */
+        if (filePath.indexOf("node_modules") == -1 && !knownInputs.has(relative)) {
```
jbedard, line 157: Can use `includes()` instead of `indexOf() == -1`

ts/private/ts_project_worker.js
```diff
@@ -0,0 +1,384 @@
...
+        }
+        return ts.sys.readFile(filePath, encoding);
+    }
+
+    function directoryExists(directory) {
+        if (!directory.startsWith(execRoot)) {
+            return false;
+        }
+        const exists = ts.sys.directoryExists(directory)
+        directory.indexOf("node_modules") == -1 && debuglog(`directoryExists ${directory} ${exists}`);
```
jbedard, line 168: `if` to avoid this long line? `.includes` instead of `indexOf`...

jbedard:
I know we don't really have any standard/enforced styling in this repo but I think it would be nice if we were at least consistent within this file? Or is there no point and that's unmaintainable? But I'd vote to just do this once before committing:
```
npx prettier --semi --print-width 120 --write ts/private/ts_project_worker.js
```

ts/private/ts_project_worker.js
```diff
@@ -0,0 +1,384 @@
...
+        debuglog(`readDirectory ${directory} ${strictView}`)
+
+        return strictView;
+    }
+
+    function invalidate(filePath, kind) {
+        if (filePath.endsWith(".params")) return;
+        debuglog(`invalidate ${filePath} : ${ts.FileWatcherEventKind[kind]}`)
+
+        if (kind == ts.FileWatcherEventKind.Created) {
```
jbedard, line 210: Normally `===` is recommended everywhere. Can we just do a search and replace and always use the triple-equals?

ts/private/ts_project_worker.js
```diff
@@ -0,0 +1,384 @@
...
+    function getDirectories(directory) {
+        const dirs = ts.sys.getDirectories(directory);
+        debuglog(`getDirectories ${directory}`, dirs)
+        return dirs;
+    }
+
+    function fileExists(filePath) {
+        const relative = path.relative(execRoot, filePath)
+        debuglog(`fileExists ${filePath} ${knownInputs.has(relative)}`)
+        if (!knownInputs.has(relative)) {
```
jbedard, line 181: can just do `return knownInputs.has(relative)`?

ts/private/ts_project_worker.js
jbedard: Looks like it's not defined anywhere (it's a lint error if you run a linter), so this is creating + modifying a global. If this worker is ever used concurrently and there's async callbacks this would very likely cause bugs.

ts/private/ts_project_worker.js
jbedard: Actually maybe that doesn't work since all the methods are defined within this closure because they consume the `Map`s...?

ts/private/ts_project_worker.js
```diff
@@ -0,0 +1,384 @@
...
+    const parsedArgs = ts.parseCommandLine(args)
+
+    const directoryWatchers = new Map();
+    const fileWatchers = new Map();
+
+    const knownInputs = new Set(initialInputs);
+
+    /** @type {ts.System} */
+    const sys = {
+        ...ts.sys,
```
jbedard, line 111: FWIW for anyone curious (such as myself)... and @thesayyn let me know if any of this is wrong or you learnt anything else interesting...

https://github.com/microsoft/TypeScript/blob/v4.7.3/src/compiler/sys.ts#L1253 is where this global `sys` object is defined. It seems like there's one and only one instance of this object and that single instance is passed around to almost every method? The global only gets accessed in a couple odd places so passing it into a method essentially overrides it for that unique invocation like what we're doing here.

Here we are cloning it per program to override some methods with caching based on that project.

It also seems to have no prototype or use of `this.` so it's ok to simply clone it instead of something such as `Object.create` to inherit from it?

.github/workflows/ci.yaml
thesayyn: Yes. I'll do that

ts/private/ts_project_worker.js
thesayyn: I am not sure.  Leftover from rnjs.

thesayyn: I'll remove this.

ts/BUILD.package
thesayyn: each npm package we download and extract through `http_archive` has a single directory at the root called `package`.

thesayyn: We want that directory to be exposed so we can copy it via `npm_package`

ts/private/ts_project_worker.js
thesayyn: When you require this file. since there isn't any exported symbols, we should remove this logic.  Good catch. Thanks!

ts/private/ts_project_worker.js
thesayyn: It doesn't actually. That'd be unnecessary since `worker` is an object that we get via cjs require. `worker = require(...)` 

Also, `log` is a top-level function that doesn't have a context. https://github.com/bazelbuild/rules_nodejs/blob/621c75f1c2daac39ee7a0896938faf6ca4fcbbb9/packages/worker/src/worker.ts#L17

ts/private/ts_project_worker.js
thesayyn: Ah.. good catch!

ts/private/ts_project_worker.js
thesayyn: Haven't thought about it tbh. I guess that'd force users some node version and up? Nevertheless, I'll convert it to typescript when it gets stable enough

ts/private/ts_project_worker.js
thesayyn: yeap.

ts/private/ts_project_worker.js
thesayyn: Agreed.

ts/private/ts_project_worker.js
thesayyn: That'd help readability. Agreed!

thesayyn: Ah, you mean we only construct the `sys` object once?  That wouldn't work since we read some state variables such as `knownInputs` `directoryWatchers` `fileWatchers`. I don't think that'd help reducing the memory usage significantly either.

ts/private/ts_project_worker.js
thesayyn: Done. sometimes I hate javascript so much.

ts/private/ts_project_worker.js
thesayyn: Oh yes. Done

ts/private/ts_project_worker.js
thesayyn: Your findings are accurate. I'd somewhat not complicate things by having only one instance passed around with different binds. I am not sure how performant that would be.  

However, my opinion is not the only one that counts.

ts/private/ts_project_worker.js
thesayyn: Done. That's better.

ts/private/ts_project_worker.js
thesayyn: Done

ts/BUILD.typescript
thesayyn: Yes. Done

e2e/worker/test.sh
thesayyn: Done. just changed `exit_with_message` func

ts/private/ts_project_worker.js
thesayyn: https://www.npmjs.com/package/@typescript/vfs  is one example that creates more than one sys instances.

e2e/worker/.bazelrc
thesayyn: https://github.com/aspect-build/rules_ts/blob/b55b8608bcd4255c6b13459dea557e63ef6bf605/ts/private/ts_project.bzl#L74 
 
 tsc will spit out some helpful information that we use for debugging purposes.

PR #28, opened 2022-05-31
chore: update to rules_js 0.9.0

gregmagolan:


ts/BUILD.typescript
```diff
@@ -1,6 +1,7 @@
 "BUILD file inserted into @npm_typescript repository"
 load("@aspect_rules_js//js:defs.bzl", "link_js_package", "js_package", "js_binary")
-load("@aspect_bazel_lib//lib:directory_path.bzl", directory_path = "directory_path")
+load("@aspect_rules_js//js/private:pnpm_utils.bzl", "pnpm_utils")
```
alexeagle, line 3: this private load will trigger a buildifier warning for the next person - we should make these constants public API

alexeagle (approved):

PR #32, opened 2022-06-02
Transpiler

alexeagle:


WORKSPACE
```diff
@@ -48,9 +48,16 @@ load("@aspect_rules_js//js:npm_import.bzl", "translate_pnpm_lock")
...
     name = "npm",
-    pnpm_lock = "//examples:pnpm-lock.yaml",
+    pnpm_lock = "//:pnpm-lock.yaml",
 )
 
 load("@npm//:repositories.bzl", "npm_repositories")
 
 npm_repositories()
+
+load("@aspect_rules_swc//swc:repositories.bzl", "LATEST_VERSION", "swc_register_toolchains")
```
jbedard, line 58: We don't want to separate repo vs example dependencies?

alexeagle: We do have https://github.com/aspect-build/bazel-examples/tree/main/ts_project_transpiler already.

I think our goal in each ruleset is minimal, standalone examples, while the bazel-examples repo is for larger ones that combine things.

I do agree that this is somewhere in the middle. What would you suggest for a minimal example of the transpiler attribute?

jbedard: I was mainly thinking about mixing repo vs example dependencies. Seeing this in the WORKSPACE makes me assume it is needed by rules_ts.

alexeagle: yeah we've decided to keep the examples simple (and easily testable) by not making them nested workspaces.

ts/test/BUILD.bazel
```diff
@@ -0,0 +1,74 @@
...
+    srcs = ["big.ts"],
+    transpiler = _transpiler,
+    tsconfig = _TSCONFIG,
+)
+
+ts_project(
+    name = "transpile_with_typeerror",
+    srcs = ["typeerror.ts"],
+    # The transpile_with_typeerror.check target will have a build failure
+    # But the default transpile_with_typeerror target should still produce JS outs
```
gregmagolan, line 57: anything downstream that consumes typeerror.js to make sure it is emitted? a js_binary perhaps?

gregmagolan: oh I see, it is checked in transpiler_test_suite(). maybe plumb it through to a js_test as well to be extra clear and add additional coverage?

alexeagle: it's in the transpiler_tests.bzl, you'll get there in a moment :)

alexeagle: sure can add build_test

examples/transpiler/BUILD.bazel
```diff
@@ -0,0 +1,91 @@
...
+#
+# But the _typecheck target fails to build:
+#
+# $ bazel build examples/swc:transpile_with_typeerror_typecheck
+# INFO: Analyzed target //examples/swc:transpile_with_typeerror_typecheck (0 packages loaded, 1 target configured).
+# ERROR: /home/alexeagle/Projects/rules_ts/examples/swc/BUILD.bazel:30:11: Compiling TypeScript project //examples/swc:transpile_with_typeerror_typings
+# examples/swc/typeerror.ts(1,14): error TS2322: Type 'number' is not assignable to type 'string'
+ts_project(
+    name = "transpile_with_typeerror",
+    srcs = ["typeerror.ts"],
```
gregmagolan, line 86: this is duplicate to what is in ts/test/BUILD.bazel; consolidate the two or intentional so that there is an example as well as a test

alexeagle: I could remove from examples, it wasn't in the bazel-examples repo I borrowed this dir from

WORKSPACE
jbedard: I'd just worry if we keep adding examples the WORKSPACE will become a mess.

Would a transpiler implemented in the example be more or less complicated? It means the example is self contained and doesn't pollute anything outside it such as this WORKSPACE, but means the example needs to somehow have it's own transpiler? Can it have a local package.json with a transpiler instead? Or a bzl with a dumb transpiler? 🤷

examples/transpiler/babel.bzl
```diff
@@ -0,0 +1,37 @@
...
+    # You might instead use a single babel_cli call to compile
+    # a directory of sources into an output directory,
+    # but you'll need to:
+    # - make sure the input directory only contains files listed in srcs
+    # - make sure the js_outs are actually created in the expected path
+    for idx, src in enumerate(srcs):
+        # see https://babeljs.io/docs/en/babel-cli
+        args = [
+            "../../../../$(location %s)" % src,
+            "--presets=@babel/preset-typescript",
```
gregmagolan, line 17: hmm. if you did
```
--presets=examples/node_modules/@babel/preset-typescript
```
could you drop the chdir?

alexeagle: maybe with a `./` on the front so nodejs doesnt try to resolve an npm package named "examples"
but do you think that's an improvement?

gregmagolan: although in its current shape I guess it would work as well... since the four `../` segments get you to the root

alexeagle: execroot="../../../.."
...
        # see https://babeljs.io/docs/en/babel-cli
        args = [
            "{}/$(location {})".format(execroot, src),
            "--presets=@babel/preset-typescript",
            "--out-file",
            "{}/$(location {})".format(execroot, js_outs[idx]),
        ]

gregmagolan: maybe it would just be be a little clearer to read

gregmagolan (approved):
🌮 ^2

PR #35, opened 2022-06-03
chore: update to aspect_rules_swc 0.9.1

gregmagolan:


alexeagle (commented):
I think we should just remove the dep instead... I'll make a PR

PR #37, opened 2022-06-03
Add bzlmod e2e

kormide:
Not quite ready, but wanted initial feedback. I need to add a new entry for rules_js to our registry.


e2e/bzlmod/MODULE.bazel
```diff
@@ -0,0 +1,28 @@
...
+
+bazel_dep(name = "aspect_rules_ts", version = "0.0.0")
+bazel_dep(name = "bazel_skylib", version = "1.1.1")
+
+local_path_override(
+    module_name = "aspect_rules_ts",
+    path = "../..",
+)
+
+# TODO: remove this after rules_js 0.10.0 pushed to aspect bcr
```
alexeagle, line 15: Just publish one to our BCR fork so this can be green?

ts/extensions.bzl
```diff
@@ -0,0 +1,11 @@
+load("//ts/private:npm_typescript_repository.bzl", "npm_typescript_repository")
+
+def _extension_impl(module_ctx):
+    npm_typescript_repository(ts_version = "4.6.3")
```
alexeagle, line 4: Fix me?

e2e/bzlmod/MODULE.bazel
```diff
@@ -0,0 +1,28 @@
...
+    module_name = "aspect_rules_js",
+    path = "../../../rules_js",
+)
+
+rules_ts_ext = use_extension("@aspect_rules_ts//ts:extensions.bzl", "ext", dev_dependency = True)
+
+rules_ts_ext.npm_typescript(
+    name = "npm_typescript",
+    ts_version = "4.6.3",
+)
```
kormide, line 26: Any opinions on the naming for the extension/method?

ts/repositories.bzl
```diff
@@ -6,60 +6,20 @@ See https://docs.bazel.build/versions/main/skylark/deploying.html#dependencies
 
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")
-load("//ts/private:versions.bzl", TS_VERSIONS = "VERSIONS")
+load("//ts/private:npm_typescript_repository.bzl", "npm_typescript_repository", _http_archive_version = "http_archive_version")
+load("//ts/private:versions.bzl", _LATEST_VERSION = "LATEST_VERSION", _TS_VERSIONS = "VERSIONS")
+
+TS_VERSIONS = _TS_VERSIONS
+LATEST_VERSION = _LATEST_VERSION
```
kormide, line 13: Should probably call this `LATEST_TS_VERSION`. Will be a breaking change.

e2e/bzlmod/MODULE.bazel
kormide: Something feels off about requiring downstream users to declare this extension, since it's something that's needed internally.

ts/repositories.bzl
alexeagle: We call it `LATEST_VERSION` in all the rules - I think I changed this the other way to be consistent. Users can (and I have done) `load("@rules_ts", typescript_version = "LATEST_VERSION")`

e2e/bzlmod/MODULE.bazel
alexeagle: Users have to tell us the version of typescript to use, or point to the package.json with `ts_version_from`.
BUT it feels simpler that either
- user should just call the `npm` extension from `rules_js` and tell it to use our BUILD.typescript file
- we could put npm_typescript itself into the BCR so it's just a `bazel_dep` to get it

kormide: Do you have a preference?

alexeagle: the second one doesn't seem like it would scale, so I think we should see what the first one looks like

PR #40, opened 2022-06-06
fix: tsc tries to read files that is no longer a src

thesayyn:
This fixes the flakiness that worker mode had when a file was removed from srcs.


e2e/worker/test.sh
```diff
@@ -53,15 +54,33 @@ cleanup
...
-echo '{"compilerOptions": {"noImplicitAny": true}}' > tsconfig.json
+echo '{"compilerOptions": {"noImplicitAny": true, "module": "ES2020", "moduleResolution": "node"}}' > tsconfig.json
 
 message="error TS7006: Parameter 'should_i' implicitly has an 'any' type."
 bazel build :ts 2>&1 | grep "$message" || exit_with_message "Case 3: expected worker to report \"$message\""
 cleanup
+
+
+message "# Case 4; assert that tsc does not read a file that's been removed from srcs"
+for i in $(seq 0 9)
```
thesayyn, line 65: brute force 🤫

ts/private/ts_project_worker.js
```diff
@@ -243,10 +244,22 @@ function createProgram(args, initialInputs) {
...
+
         return {
             close: () => {
-                directoryWatchers.delete(directoryPath);
+                const callbacks = directoryWatchers.get(directoryPath);
+                callbacks.delete(cb);
+                if (callbacks.size == 0) {
+                    directoryWatchers.delete(directoryPath)
+                } else {
+                    directoryWatchers.set(directoryPath, callbacks)
```
jbedard, line 261: I don't think we need to set it again?

ts/private/ts_project_worker.js
```diff
@@ -222,11 +225,9 @@ function createProgram(args, initialInputs) {
...
         if (kind === ts.FileWatcherEventKind.Created) {
             knownInputs.add(filePath);
-
-            const dirname = path.dirname(filePath);
-            getDirectoryWatcherForPath(filePath)?.(dirname);
-
-            // signal that the directory has been created.
+            const directoryWatcher = getDirectoryWatcherForPath(filePath)
+            directoryWatcher?.(path.dirname(filePath));
+            directoryWatcher?.(filePath);
```
jbedard, line 230: Why do we need to invoke the directory watcher with both the directory and file?

e2e/worker/test.sh
```diff
@@ -14,7 +14,8 @@ exit_with_message() {
 }
 
 cleanup() {
-    rm -f evil.ts included.ts
+    rm -f evil.ts
+    find . -type f -name '_addendum_*.ts' -exec rm {} \;
```
jbedard, line 18: Can we use `git clean`? Then this method doesn't need all the generated files listed in it...

ts/private/ts_project_worker.js
thesayyn: to signal that there have been some changes in that directory.

e2e/worker/test.sh
thesayyn: I am afraid that it might be destructive since you could have some local changes and invoke test.sh then it would take back local changes which isn't something we desire. so I wanted to be explicit about what's been generated and what's not.

ts/private/ts_project_worker.js
thesayyn: You are right. but wouldn't it help you reason about the whole directory watcher logic?

ts/private/ts_project_worker.js
jbedard: We have to signal it twice though? Does that fix something?

ts/private/ts_project_worker.js
jbedard: I don't think so, it's just confusing imo 🤷

ts/private/ts_project_worker.js
```diff
@@ -243,10 +244,22 @@ function createProgram(args, initialInputs) {
...
+        const callbacks = directoryWatchers.get(directoryPath) || new Set();
+        callbacks.add(cb)
+        directoryWatchers.set(directoryPath, callbacks);
+
         return {
             close: () => {
-                directoryWatchers.delete(directoryPath);
+                const callbacks = directoryWatchers.get(directoryPath);
+                callbacks.delete(cb);
+                if (callbacks.size == 0) {
```
jbedard, line 258: Should be `===`. Basically never use `==`, it's essentially a mistake from the 90s that can never be undone :(

jbedard:
Why aren't we writing this file in typescript? I think half the things I'm pointing out and probably the bug you're fixing would have been compiler errors...

ts/private/ts_project_worker.js
thesayyn: given filePath `/tmp/bin/feat/1.ts`  first we have to signal there is been some changes in `/tmp/bin/feat` which is what line 229 does by getting dirname from filepath. the second one actually specifies which file has changed.

thesayyn:
> Why aren't we writing this file in typescript? I think half the things I'm pointing out and probably the bug you're fixing would have been compiler errors...

I'll make a separate PR for that.

ts/private/ts_project_worker.js
thesayyn: I'll remove then

ts/private/ts_project_worker.js
jbedard: Is that part of a contract or API that we are following? Why do we have to do that?

thesayyn: Yes without this logic tsc does not pick up changes properly.

jbedard: As long as there's a test that fails when either of those two lines is removed. Or a link to something explaining it. Is this a typescript API with docs somewhere saying this?

jbedard (approved):

ts/private/ts_project_worker.js
thesayyn: I couldn't come across a proper implementation `ts.sys`. my assumption was that it's how a real fs would behave when tsc calls `watchDirectory` with `recursive = true`

a simple example 

```ts
const ts = require("typescript");
ts.sys.watchDirectory("./dir", console.log, true);
```

will yield 

```bash
dir/bin
dir/bin/1.ts
```

adding a test for this is gonna be a tricky one but I'll follow with another PR

PR #41, opened 2022-06-07
fix: flip supports_workers True by default

thesayyn:


alexeagle (commented):
I think we should shake out some bugs before landing this

alexeagle:
was included in #42

PR #49, opened 2022-06-10
chore: update to rules_js 0.12.1

gregmagolan:


ts/BUILD.typescript
```diff
@@ -70,9 +70,9 @@ npm_package(
 npm_link_package(
     name = "node_modules/@bazel/worker",
     src = "npm_at_bazel_worker",
-    deps = [
-        ":node_modules/google-protobuf"
-    ],
+    deps = {
+        ":node_modules/google-protobuf/store/3.20.1": "",
```
alexeagle, line 74: I totally tried that...

alexeagle (approved):

ts/BUILD.typescript
gregmagolan: Hah. This was my first try as well and is wrong. It should actually be /0.0.0 in this case since its linking there with the vanilla npm_link_package as if it was a 1p dep and not the generated one from the __links repo

ts/BUILD.typescript
```diff
@@ -70,9 +70,9 @@ npm_package(
 npm_link_package(
     name = "node_modules/@bazel/worker",
     src = "npm_at_bazel_worker",
-    deps = [
-        ":node_modules/google-protobuf"
-    ],
+    deps = {
+        ":node_modules/google-protobuf/store/0.0.0": "",
```
alexeagle, line 74: So confusing. Maybe the version shouldn't appear in this case?

alexeagle (approved):

ts/BUILD.typescript
gregmagolan: Maybe, but I like the consistency of always having a version with 0.0.0 being a default for 1p deps; you can override to any version you want when you link

PR #54, opened 2022-06-17
feat: make rules_ts compatible with bzlmod

kormide:
Originally I wanted the e2e MODULE.bazel to use the single package npm import extension from `rules_js` to fetch the npm dependencies. However, the cascading effect is that we need to use `npm_import` for the `WORKSPACE` deps as well. To do that, we have have two `rules_ts_dependencies()` macros because we can't run `npm_import` in the same file that we fetch `rules_js`, and that creates a less desirable DX for those using WORKSPACE.

This depends on https://github.com/aspect-build/bazel-central-registry/pull/10.


ts/extensions.bzl
```diff
@@ -0,0 +1,18 @@
+"""Define module extensions for using rules_ts with bzlmod.
+See https://bazel.build/docs/bzlmod#extension-definition
+"""
+
+load("//ts/private:npm_repositories.bzl", "npm_dependencies")
+load("//ts:repositories.bzl", LATEST_TS_VERSION = "LATEST_VERSION")
```
alexeagle, line 6: nit: no need for rename on load

e2e/bzlmod/MODULE.bazel
```diff
@@ -0,0 +1,27 @@
...
+
+rules_ts_ext = use_extension(
+    "@aspect_rules_ts//ts:extensions.bzl",
+    "rules_ts_ext",
+    dev_dependency = True,
+)
+
+rules_ts_ext.deps()
+
+use_repo(rules_ts_ext, "npm_typescript")
```
alexeagle, line 23: seems pretty reasonable here.

e2e/bzlmod/MODULE.bazel
```diff
@@ -0,0 +1,27 @@
+module(
+    name = "e2e_bzlmod",
+    compatibility_level = 1,
+    version = "0.0.0",
+)
+
+bazel_dep(name = "aspect_rules_ts", dev_dependency = True, version = "0.0.0")
```
alexeagle, line 7: OMG I had an emotional moment when aspect_rules_js was missing here

alexeagle (approved):
sweet

Issue #85, opened 2022-07-11
ts_validate_options should be a validation action


alexeagle:
once we force users to upgrade to bazel 6.0 we should adopt
https://docs.bazel.build/versions/main/skylark/rules.html#validation-actions

(note, the feature doesn't exist in Bazel 5 afaict)

alexeagle:
Also, validation actions give us a way to type-check `.d.ts` inputs even if no output files are produced. Maybe useful for a "ts_declaration" rule ?

alexeagle:
Fixed by @dzbarsky on 2.x branch - thanks!

PR #86, opened 2022-07-11
fix: give better error when ts_project should be js_library

### LINKED ISSUE #72: Cannot depend on ts_project that only contains declaration files
Consider the following modification to `examples/ts_project_dep`:

```ts
// a/a.d.ts
export interface Foo { }
```

```starlark
# a/BUILD.bazel

ts_project(
    name = "ts",
    srcs = ["a.d.ts"],
    declaration = True,
)
```

```ts
// b/b.ts
import { Foo } from '../a/a'
```

```starlark
# a/BUILD.bazel

ts_project(
    name = "ts",
    srcs = ["b.ts"],
    deps = ["//a:ts"],
)
```

Running `bzl build //examples/ts_project_dep/...` results in the following error:

```
ERROR: /Users/duarte.nunes/code/rules_ts/examples/ts_project_dep/b/BUILD.bazel:3:11: in ts_project_rule rule //examples/ts_project_dep/b:ts:
Traceback (most recent call last):
	File "/Users/duarte.nunes/code/rules_ts/ts/private/ts_project.bzl", line 158, column 47, in _ts_project_impl
		inputs = copy_files_to_bin_actions(ctx, inputs),
	File "/private/var/tmp/_bazel_duarte.nunes/f58cdaf222cf7d2c1c0920674e48c308/external/aspect_bazel_lib/lib/private/copy_to_bin.bzl", line 109, column 36, in copy_files_to_bin_actions
		return [copy_file_to_bin_action(ctx, file, is_windows = is_windows) for file in files]
	File "/private/var/tmp/_bazel_duarte.nunes/f58cdaf222cf7d2c1c0920674e48c308/external/aspect_bazel_lib/lib/private/copy_to_bin.bzl", line 42, column 13, in copy_file_to_bin_action
		fail(_file_in_different_package_error_msg(file, ctx.label))
Error in fail:
Expected to find file a.d.ts in //examples/ts_project_dep/b, but instead it is in //examples/ts_project_dep/a.

To use copy_to_bin, either move a.d.ts to //examples/ts_project_dep/b, or move the copy_to_bin
target to //examples/ts_project_dep/a using:

    buildozer 'new copy_to_bin a.d' //examples/ts_project_dep/a:__pkg__
    buildozer 'add srcs a.d.ts' //examples/ts_project_dep/a:a.d
    buildozer 'new_load @aspect_bazel_lib//lib:copy_to_bin.bzl copy_to_bin' //examples/ts_project_dep/a:__pkg__
    buildozer 'add visibility //examples/ts_project_dep/b:__subpackages__' //examples/ts_project_dep/a:a.d
```

alexeagle:
Fixes #72

### DIFF
--- a/ts/private/ts_project.bzl
+++ b/ts/private/ts_project.bzl
@@ -124,11 +124,27 @@ def _ts_project_impl(ctx):
         ])
         outputs.append(ctx.outputs.buildinfo_out)
     runtime_outputs = json_outs + js_outs + map_outs
-    typings_outputs = typings_outs + typing_maps_outs + [s for s in ctx.files.srcs if s.path.endswith(".d.ts")]
+    typings_srcs = [s for s in ctx.files.srcs if s.path.endswith(".d.ts")]
 
-    if len(js_outs) + len(typings_outputs) < 1:
+    if len(js_outs) + len(typings_outs) < 1:
         label = "//{}:{}".format(ctx.label.package, ctx.label.name)
-        if ctx.attr.transpile:
+        if len(typings_srcs) > 0:
+            no_outs_msg = """ts_project target {target} only has typings in srcs.
+Since there is no `tsc` action to perform, there are no generated outputs.
+
+> ts_project doesn't support "typecheck-only"; see https://github.com/aspect-build/rules_ts/issues/88
+
+This should be changed to js_library, which can be done by running:
+
+    buildozer 'new_load @aspect_rules_js//js:defs.bzl js_library' //{pkg}:__pkg__
+    buildozer 'set kind js_library' {target}
+    buildozer 'remove declaration' {target}
+
+""".format(
+                target = label,
+                pkg = ctx.label.package,
+            )
+        elif ctx.attr.transpile:
             no_outs_msg = """ts_project target %s is configured to produce no outputs.
 
 This might be because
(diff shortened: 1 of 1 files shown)


ts/private/ts_project.bzl
```diff
@@ -124,11 +124,25 @@ def _ts_project_impl(ctx):
...
     runtime_outputs = json_outs + js_outs + map_outs
-    typings_outputs = typings_outs + typing_maps_outs + [s for s in ctx.files.srcs if s.path.endswith(".d.ts")]
+    typings_srcs = [s for s in ctx.files.srcs if s.path.endswith(".d.ts")]
 
-    if len(js_outs) + len(typings_outputs) < 1:
+    if len(js_outs) + len(typings_outs) < 1:
         label = "//{}:{}".format(ctx.label.package, ctx.label.name)
-        if ctx.attr.transpile:
+        if len(typings_srcs) > 0:
+            no_outs_msg = """ts_project target {target} has no outputs, but it has typings in srcs.
```
thesayyn, line 132: this makes me wonder whether tsc is supposed to type check .d.ts sources as well.

alexeagle: I think it could - but it doesn't produce any outputs in that case, so it would have to be run as a Bazel "validation action" whose outputs aren't required in the dep graph. I'll note that in https://github.com/aspect-build/rules_ts/issues/85#issuecomment-1180520106

thesayyn (approved):

ts/private/ts_project.bzl
thesayyn: Ah cool. I'd expect tsc to at least type check and copy it over if instructed. At least that is what I'd expect.

Issue #23, opened 2022-05-19
Add remaining test cases


alexeagle:
There are like 20 folders under packages/typescript/test/ts_project in rules_nodejs that have coverage we still want:

  - [x] a
  - [x] allow_js
  - [x] b
  - [x] c
  - [x] data
  - [x] declarationdir
  - [x] declarationdir_with_value
  - [x] declaration_only
  - [x] directory_declarations
  - [x] empty_intermediate
  - [x] esm
  - [x] extends_chain
  - [x] filegroup
  - [x] generated_tsconfig
  - [x] import_package_by_name
  - [x] js_library
  - [x] json
  - [x] jsx
  - [x] ngc
  - [x] outdir
  - [x] output_group
  - [x] rootdir
  - [x] rootdir_with_value
  - [x] simple
  - [x] swc
  - [x] tsbuildinfofile
  - [x] ts_config
  - [x] validation
  - [x] worker

alexeagle:
Note for this one: in https://github.com/aspect-build/rules_ts/pull/176 I ported over the tests for composite project references, however I dropped a test case that we'll need to add back: there's a use of `args=["--emitBOM"]` that we were using to verify arbitrary args get passed to tsc, and then the jasmine test in https://github.com/bazelbuild/rules_nodejs/blob/stable/packages/typescript/test/ts_project/b/b.spec.ts verified that the BOM was written. I couldn't get that green in the same PR, but also there's no particular reason to mix that assertion with the composite project references.
We should make sure to restore that coverage as part of this issue.

alexeagle:
Need to check:

  - [ ] a
  - [ ] allow_js
  - [ ] b
  - [ ] c
  - [ ] data
  - [ ] declarationdir
  - [ ] declarationdir_with_value
  - [ ] declaration_only
  - [ ] directory_declarations
  - [ ] empty_intermediate
  - [ ] esm
  - [ ] extends_chain
  - [ ] filegroup
  - [ ] generated_tsconfig
  - [ ] import_package_by_name
  - [ ] js_library
  - [ ] json
  - [ ] jsx
  - [ ] ngc
  - [ ] outdir
  - [ ] output_group
  - [ ] rootdir
  - [ ] rootdir_with_value
  - [ ] simple
  - [ ] swc
  - [ ] tsbuildinfofile
  - [ ] ts_config
  - [ ] tsconfig-base.json
  - [ ] validation
  - [ ] worker

PR #87, opened 2022-07-11
test: add case for declaration_dir, out_dir and root_dir

thesayyn:
Part of #23


examples/declarationdir_with_value/BUILD.bazel
```diff
@@ -0,0 +1,40 @@
...
+        "$(rootpaths :types)",
+        "$(rootpaths :transpile)",
+    ],
+    out = "params.txt"
+)
+
+write_source_files(
+    name = "write_params",
+    files = {
+        "params.out": "params.txt",
```
alexeagle, line 38: nit: params.out isn't a very useful name here, when reading the folder I'd rather see `expected_outputs.txt` or sth

PR #91, opened 2022-07-14
fix: correctly invalidate failed lookups

### LINKED ISSUE #80: Successive runs with differing/missing types don't fail
Super hard to find a good summary for this one! Anyway, picking up from #79, let's see if I can summarize.

The repro is in this repo (a repro repo!): https://github.com/paullewis/repro-rules-ts-issue-79

Essentially if you have the following:
```bazel

ts_project(
    name = "src",
    srcs = [
        "src.ts",
    ],
    declaration = True,
    tsconfig = "//:tsconfig",
    deps = [
        "//:node_modules/@types/long",
    ],
)
```

And here we can assume that `src.ts` uses the types in the `deps`. It builds just fine.

However, if we comment out the deps, rather than failing the build it still succeeds. @alexeagle mentioned in #79 that might be something to do with the worker mode, in case that's of use.

thesayyn:
Fixes #80

### DIFF
--- a/ts/private/ts_project_worker.js
+++ b/ts/private/ts_project_worker.js
@@ -342,18 +342,6 @@ function createProgram(args, initialInputs) {
         compilerOptions.extendedDiagnostics && host.trace?.(message)
     }
 
-    function getDirectoryWatcherForPath(fileName) {
-        const p = path.dirname(fileName); 
-        const callbacks = directoryWatchers.get(p);
-        if (callbacks) {
-            return (filePath) => {
-                for (const callback of callbacks) {
-                    callback(filePath)
-                }
-            };
-        }
-    }
-
     function readFile(filePath, encoding) {
         const relative = path.relative(execRoot, filePath);
         // external lib are transitive sources thus not listed in the inputs map reported by bazel.
@@ -378,11 +366,24 @@ function createProgram(args, initialInputs) {
         return filesystemTree.fileExists(path.relative(execRoot, filePath));
     }
 
+    
     function readDirectory(directoryPath, extensions, exclude, include, depth) {
         return filesystemTree.readDirectory(
             path.relative(execRoot, directoryPath), 
             extensions, exclude, include, depth
         ).map(p => path.isAbsolute(p) ? p : path.join(execRoot, p))
+    }
+
+    function getDirectoryWatcherForPath(fileName) {
+        const p = path.dirname(fileName); 
+        const callbacks = directoryWatchers.get(p);
+        if (callbacks) {
+            return (filePath) => {
+                for (const callback of callbacks) {
+                    callback(filePath)
+                }
+            };
+        }
     }
 
     function invalidate(filePath, kind, digest) {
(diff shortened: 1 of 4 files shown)


e2e/worker/BUILD
```diff
@@ -5,31 +5,34 @@ load("@aspect_bazel_lib//lib:copy_to_bin.bzl", "copy_to_bin")
 ts_project(
     name = "ts",
     srcs = glob(
-        ["*.ts"],
```
jbedard, line 8: Looks like buildozer ran on the first commit as well? Maybe put the buildozer commit first or in another PR before the actual fix-commit?

ts/private/ts_project_worker.js
```diff
@@ -330,16 +342,15 @@ function createProgram(args, initialInputs) {
...
-        for (const [directory, callbacks] of directoryWatchers.entries()) {
-            if (isAncestorDirectory(directory, path)) {
-                debuglog(`found a directory watcher for ${path} ${directory}`);
-                return (filePath) => {
-                    for (const callback of callbacks) {
-                        callback(filePath)
-                    }
-                };
-            }
+    function getDirectoryWatcherForPath(fileName) {
```
jbedard, line 345: Seems like the only time we call this we just invoke the return method right away... should we just do `invokeDirectoryWatcherForPath`?

e2e/worker/BUILD
thesayyn: Sure.

ts/private/ts_project_worker.js
jbedard: Right now we're doing 3 things though: collect the callbacks, create a wrapper function, invoke the wrapper function. Why not just do one thing: invoke the callbacks?

ts/private/ts_project_worker.js
```diff
@@ -286,6 +281,23 @@ function createProgram(args, initialInputs) {
...
+
+    function clearTimeout(cb) {
+        taskQueue.delete(cb)
+    }
+
+    function applyChanges() {
+        debuglog("Applying changes");
+        for (const task of taskQueue) {
+            taskQueue.delete(task);
+            task();
```
jbedard, line 297: If this throws we want to stop the loop? Or are we certain enough it won't throw so it doesn't matter?

ts/private/ts_project_worker.js
thesayyn: Yes. I want to hide the fact that there could be more than one callback for directory watchers. Invoking is the job for the invalidate function and keep it consistent with fileWatchers.

ts/private/ts_project_worker.js
thesayyn: That's fatal and the worker should die. There is no right way to handle the error that could be raised by the underlying function. Think of this as setTimeout with a callback that throws.

thesayyn: It shouldn't throw for user errors. If it throws then either worker does something bad or tsc does.

ts/private/ts_project_worker.js
jbedard: Yeah that is clearer to me, and while only a minor difference I find it easier to ready 👍 🤷

jbedard (approved):

PR #92, opened 2022-07-14
feat: add support for .mts and .cts source files

jbedard:
The file extensions allow individual files run by node to override the `type` in `package.json`. 

Added in typescript 4.7: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-7.html#new-file-extensions


alexeagle (approved):
Nice

jbedard:
The failure was me removing the preserve/jsx from the tsconfig but not BUILD, since it's not relevant for this test.

@thesayyn you can press merge if you think it looks good

thesayyn (approved):

PR #94, opened 2022-07-15
fix: support multiple typescript versions

thesayyn:
This will allow us to assert that worker mode doesn't rely on typescript apis that are only present in recent versions.


e2e/multiple_version/versions.bzl
```diff
@@ -0,0 +1,86 @@
...
+    "4.5.2",
+    "4.5.3",
+    "4.5.4",
+    "4.5.5",
+    "4.6.2",
+    "4.6.3",
+    "4.6.4",
+    "4.7.2",
+    "4.7.3",
+    "4.7.4"
```
alexeagle, line 85: this seems pretty bad for maintenance, can we use semver ranges instead?

ts/BUILD.typescript
```diff
@@ -10,7 +10,7 @@ npm_package_internal(
     name = "npm_typescript",
```
alexeagle, line 10: does this line need a replacement as well?

ts/defs.bzl
```diff
@@ -260,7 +260,7 @@ def ts_project(
     """
 
     # Disable workers if a custom tsc was provided but not a custom tsc_worker.
-    if tsc != _tsc and tsc_worker == _tsc_worker:
+    if tsc != _tsc and tsc_worker == _tsc_worker and supports_workers == None:
```
jbedard, line 263: What's this? Where would it get set to `None` and wouldn't we still want to disable it with the current condition (so it would be an `or`)?

ts/BUILD.typescript
thesayyn: not necessary since this target represents the npm package.

e2e/multiple_version/versions.bzl
thesayyn: sure. I'll change it.

ts/defs.bzl
thesayyn: we want supports_workers to be whatever value was passed to ts_project. Now that there could be more than one version of tsc, users will still want to benefit from workers even when they provide a non-default tsc.

Though this was the intention, this change is faulty since supports_workers is never `None` implicitly.

jbedard: Can the same worker process be used even though a different typescript version might be used by different `ts_project` targets?

If yes, does this work...?
```
# default param values
tsc = _tsc,
tsc_worker = None
supports_workers = None

# Use _tsc_worker by default if workers not explicitly disabled
if tsc_worker == None and supports_workers != False:
   tsc_worker = _tsc_worker
   supports_workers = True
# Remove any worker if explicitly disabled
elif supports_workers == False
   tsc_worker = None

# If a worker is set then use it
if tsc_worker != None:
   ... use the worker ...
```

thesayyn: Bazel will spawn new workers for different versions of typescript. Though bazel is not aware of typescript, it only cares what goes into `tools` of `ctx.actions.run`.  Transitive files that comes with `ctx.actions.run#executable` goes into `tools` implicitly. Thus every time a distinct `Target` is passed to `ts_project#tsc_worker`, bazel calculates a new hash to see if any of the already spawned workers has the same set `tools` which won't be the case for `ts_project` targets that has distinct `tsc_worker` attributes.

See: https://github.com/bazelbuild/bazel/blob/8a195443c629f94d31502ff0bab3b91908b37e46/src/main/java/com/google/devtools/build/lib/worker/WorkerFilesHash.java#L69 and https://github.com/bazelbuild/bazel/blob/cdb77503765eac5936be4b02201fc2c7f259d84c/src/main/java/com/google/devtools/build/lib/worker/WorkerKey.java#L107

thesayyn: > Can the same worker process be used even though a different typescript version might be used by different `ts_project` targets?
> 
> If yes, does this work...?
> 
> ```
> # default param values
> tsc = _tsc,
> tsc_worker = None
> supports_workers = None
> 
> # Use _tsc_worker by default if workers not explicitly disabled
> if tsc_worker == None and supports_workers != False:
>    tsc_worker = _tsc_worker
>    supports_workers = True
> # Remove any worker if explicitly disabled
> elif supports_workers == False
>    tsc_worker = None
> 
> # If a worker is set then use it
> if tsc_worker != None:
>    ... use the worker ...
> ```
It might. I'll try this out.

cgrindel:
@thesayyn Is this PR still relevant? If so, can get it over the finish line without too much effort?

thesayyn:
I still want to land this but let's convert it to draft for now.

PR #98, opened 2022-07-20
test: add example of ts_project .d.ts source

jbedard:
A test for `ts_project` containing `.d.ts` source files and another `ts_project` importing via relative + named imports (`js_library` + linked `npm_package`).

It seems like this is not possible with rules_nodejs: https://github.com/bazelbuild/rules_nodejs/commit/7b42331ac0bfc962e58067f4f0ce7dea5ffa7148


gregmagolan (approved):
🌮

PR #108, opened 2022-07-29
refactor: make fstree handle watchers

thesayyn:


ts/private/ts_project_worker.js
```diff
@@ -29,85 +29,138 @@ function getArgsFromParamFile() {
 function createFilesystemTree(root, inputs) {
     const hashes = new Map();
     const tree = {};
+    const watchingTree = {};
+
+    const TYPE = {
```
jbedard, line 34: Do these get exposed in the resulting `FilesystemTree` at all? Does it matter they are recreated for each `createFilesystemTree`?

I assume if/when this is an es6 class they will not be re-declared on every instance.

jbedard, line 34: Should we call this `NODE_TYPE`?

ts/private/ts_project_worker.js
```diff
@@ -29,85 +29,138 @@ function getArgsFromParamFile() {
...
     }
 
     function add(p, hash) {
-        const parts = p.split(path.sep);
-        let backtrack = tree;
-        for (const part of parts) {
-            if (typeof backtrack[part] != "object") {
-                backtrack[part] = {};
+        const parts = path.parse(p);        
+        const dirs = parts.dir.split(path.sep).filter(p => p != "");
```
jbedard, line 71: Do we actually need the `.filter`? I'd think if the path is normalized it wouldn't be necessary.

ts/private/ts_project_worker.js
```diff
@@ -29,85 +29,138 @@ function getArgsFromParamFile() {
...
 
+    function update(p, hash) {
+        let node = getNode(p);
+        const parts = p.split(path.sep);
+        const base = parts.pop();
+        notifyWatchers(parts, base, node[Type], EVENT_TYPE.UPDATED);
+        hashes.set(p, hash);
+    }
+
+    // TODO: file an issue to typescript repo?
```
jbedard, line 145: Why?

gregmagolan (approved):
🌮

ts/private/ts_project_worker.js
```diff
@@ -120,14 +173,60 @@ function createFilesystemTree(root, inputs) {
...
+            if (!(part in node)) {
+                node[part] = {};
+            }
+            node = node[part];
+        }
+        if (!(Watcher in node)) {
+            node[Watcher] = new Array();
+        }
+        let i = node[Watcher].push(callback) - 1;
+        return () => node[Watcher].splice(i, 1);
```
jbedard, line 226: Can't the index change? Unless the nodes are added/removed in a fifo order I think the indexes will change, and in that case just use push/pop instead of indexes.

I assume this has to be `node[Watcher] = node[Watcher].filter(...)`?

ts/private/ts_project_worker.js
```diff
@@ -374,82 +468,44 @@ function createProgram(args, initialInputs) {
...
-        };
+        return {close};
     }
 
     function watchFile(filePath, callback, interval) {
-        debuglog(`watchFile ${filePath} ${interval}`);
-        const relativeFilePath = path.relative(execRoot, filePath);
-        fileWatchers.set(relativeFilePath, (kind) => callback(filePath, kind));
-        return { close: () => fileWatchers.delete(relativeFilePath) };
+        if (!filePath.startsWith(filePath)) {
```
jbedard, line 501: What's this?

ts/private/ts_project_worker.js
thesayyn: Exactly. these would have been `const enum` if we were to use typescript. I filed #114 to come back for this.

jbedard: I assume `const enum` would be at the top of the file (outside this function). Idk if you'd want to do that here? 🤷

thesayyn: Nodes represent the filesystem nodes so I am not sure whether renaming would make this more ambiguous. Also, since these are only attached to nodes, makes the prefix a bit redundant?

ts/private/ts_project_worker.js
thesayyn: oh shoot. I'll fix it!

ts/private/ts_project_worker.js
jbedard: I guess if it's always `node[TYPE.*]` it makes sense. Somewhere down below I saw `TYPE.*` and I had to scroll up to understand what was going on so I thought I'd mention it. You're probably right though 👍

ts/private/ts_project_worker.js
thesayyn: Good catch. you are absolutely right.

jbedard: Can you think of an easy test to add that reproduces it?

ts/private/ts_project_worker.js
thesayyn: Yes. it would. or a static property of the class.

ts/private/ts_project_worker.js
thesayyn: problem is that `.split` will return the whole string when there is no match for the `sep`. 

for instance
```
const path = require("path");
const p = path.parse("dir");
const dirs = p.dir.split(path.sep);
console.log(p, dirs);
```

`p.dir.split(path.sep)` will yield `[""]`

thesayyn: https://codesandbox.io/s/node-playground-forked-9dyrpc

ts/private/ts_project_worker.js
thesayyn: I think so

ts/private/ts_project_worker.js
jbedard: Ah I see, I didn't notice what the `parts.dir` was and assumed it was `p.split("/")`...

ts/private/ts_project_worker.js
thesayyn: Added a new test case and fixed!

ts/private/ts_project_worker.js
```diff
@@ -474,7 +474,7 @@ function createProgram(args, initialInputs) {
 
         if (kind === ts.FileWatcherEventKind.Created) {
             filesystemTree.add(filePath, digest);
-            // TODO: explain why tsc (i hate you) decides not to invalidate failed lookups based
```
jbedard: 🤣

jbedard (approved):

thesayyn:
green again.

ts/private/ts_project_worker.js
```diff
@@ -374,82 +465,46 @@ function createProgram(args, initialInputs) {
...
     }
 
     function watchFile(filePath, callback, interval) {
-        debuglog(`watchFile ${filePath} ${interval}`);
-        const relativeFilePath = path.relative(execRoot, filePath);
-        fileWatchers.set(relativeFilePath, (kind) => callback(filePath, kind));
-        return { close: () => fileWatchers.delete(relativeFilePath) };
+        if (!path.isAbsolute(filePath)) {
+            filePath = path.resolve(filePath);
+        } else if (!filePath.startsWith(execRoot)) {
```
jbedard, line 500: If the path was relative we never need to do this? Should it be `if` instead of `else if`?

thesayyn: The path is sometimes relative. The only instance I came across was file watcher opened against tsconfig.json. the reason why https://github.com/aspect-build/rules_ts/runs/7643046681?check_suite_focus=true#step:7:72 failed at https://github.com/aspect-build/rules_ts/pull/108/commits/83b946e3ff2016ebf88bc228732b7c8e407dce0e

jbedard: and `path.relative(filePath)` won't ever start with `execRoot`?

thesayyn: ah, you are right. if typescript ever passed something like "../../../../" it might be out of execRoot when normalized. will fix it!

thesayyn: fixed.

PR #113, opened 2022-08-01
refactor: update to rules_js 1.0.0-rc.4

gregmagolan:
rc4 is not out yet but this update to what the derivate rules API will look like when it is released


.github/workflows/ci.yaml
```diff
@@ -20,7 +20,7 @@ jobs:
             matrix:
                 folder:
                     - '.'
-                    - 'e2e/bzlmod'
+                    # - 'e2e/bzlmod' # FIXME
```
alexeagle, line 23: the fact that we can't fix it now is an indicator that the changes were naughty to make during rc :)

examples/declarationdir_with_value/BUILD.bazel
```diff
@@ -14,27 +14,27 @@ ts_project(
 )
 
 filegroup(
-    name = "types",
+    name = "declarations",
     srcs = [":transpile"],
-    output_group = "types",
+    output_group = "declarations",
```
alexeagle, line 19: hm, this breaking change feels unnecessary to me. Even if you prefer the name "declarations" to "types" it's going to cause a lot of churn in userland code

examples/deps_lib/BUILD.bazel
```diff
@@ -23,10 +23,6 @@ npm_package(
     name = "deps_lib",
     srcs = [
         ":deps_ts",
-        # TODO: This seems undesirable to have to specify the transitive closure of ts_project
```
alexeagle, line 26: yay

ts/private/ts_lib.bzl
```diff
@@ -12,7 +12,7 @@ ValidOptionsInfo = provider(
 
 # Targets in deps must provide one or the other of these
 DEPS_PROVIDERS = [
-    [DeclarationInfo],
```
alexeagle, line 15: https://en.wikipedia.org/wiki/Robustness_principle suggests that we should consider keeping DeclarationInfo here as an option. I think it's pretty trivial in the implementation to check it if JsInfo isn't present on a dep

ts/private/ts_lib.bzl
```diff
@@ -22,7 +22,16 @@ STD_ATTRS = {
         doc = "https://www.typescriptlang.org/docs/handbook/compiler-options.html",
     ),
     "data": attr.label_list(
-        doc = "Runtime dependencies to include in binaries/tests that depend on this TS code",
+        doc = """Runtime dependencies to include in binaries/tests that depend on this library.
+
+        If this list contains linked npm packages, npm package store targets or other targets that provide 'JsInfo',
```
alexeagle, line 27: why so much single quotes instead of backticks? won't render properly as code

ts/private/ts_project.bzl
```diff
@@ -123,7 +127,7 @@ def _ts_project_impl(ctx):
             ctx.outputs.buildinfo_out.short_path,
         ])
         outputs.append(ctx.outputs.buildinfo_out)
-    runtime_outputs = json_outs + js_outs + map_outs
+    sources = json_outs + js_outs + map_outs
```
alexeagle, line 130: this is a confusing rename

ts/private/ts_project.bzl
```diff
@@ -158,17 +162,23 @@ This might be because
...
+        default_outputs = sources[:] if len(sources) else declarations[:]
     else:
         # We must avoid tsc writing any JS files in this case, as tsc was only run for typings, and some other
         # action will try to write the JS files. We must avoid collisions where two actions write the same file.
         arguments.add("--emitDeclarationOnly")
 
         # We don't produce any DefaultInfo outputs in this case, because we avoid running the tsc action
-        # unless the DeclarationInfo is requested.
-        default_outputs_depset = depset([])
+        # unless the declarations are requested.
```
alexeagle, line 180: did you verify if this continues to work? If some downstream action asks for the JsInfo but doesn't need declarations, are we now going to trigger typechecks where we didn't previously?

(that may be a critical reason to keep the DeclarationInfo separate from JsInfo)

.github/workflows/ci.yaml
gregmagolan: The error message was for `js/libs.bzl.
```
ERROR: error loading package '': at /private/var/tmp/_bazel_greg/77e9eabe4dbb96758b221045a8e291a6/external/aspect_rules_ts/ts/defs.bzl:14:6: at /private/var/tmp/_bazel_greg/77e9eabe4dbb96758b221045a8e291a6/external/aspect_rules_ts/ts/private/ts_project.bzl:6:6: cannot load '@aspect_rules_js.0.13.0//js:libs.bzl': no such file
```
Sad we can't add a new entry in bzlmod because of the rc suffix

ts/private/ts_lib.bzl
gregmagolan: One problem with that is that we'd be advertising interop when we wouldn't be picking up the transitive npm dependencies from any ts_projects the produce DeclarationInfo

ts/private/ts_lib.bzl
gregmagolan: VsCode :(

I should figure out the regex to fix them all

ts/private/ts_project.bzl
gregmagolan: meant to align with what is produced in JsInfo. `output_sources` clearer?

ts/private/ts_project.bzl
gregmagolan: I'll double check but yes it should work. A declaration file is only created if it ends up as an input to an action. Having File objects passed around in `declarations` and `transitive_declarations` in JsInfo doesn't trigger the actions unless they are passed as inputs to actions that need to run or as runfiles to binary targets that are run.

examples/declarationdir_with_value/BUILD.bazel
gregmagolan: Went back to just `types` in the output group. If users want transitive_declarations or transitive_sources they can use js_filegroup instead

examples/deps_lib/BUILD.bazel
gregmagolan: yup. that one was a very big rabbit hole

ts/private/ts_project.bzl
gregmagolan: updated to `output_sources`

ts/private/ts_lib.bzl
alexeagle: yes that will be confusing for users that some deps have different behavior. I'm okay with this like it is

ts/private/ts_project.bzl
alexeagle: I verified this locally and pretty sure it does work correctly, since `include_declarations` is False by default

ts/private/ts_lib.bzl
```diff
@@ -22,7 +22,16 @@ STD_ATTRS = {
         doc = "https://www.typescriptlang.org/docs/handbook/compiler-options.html",
     ),
     "data": attr.label_list(
-        doc = "Runtime dependencies to include in binaries/tests that depend on this TS code",
+        doc = """Runtime dependencies to include in binaries/tests that depend on this library.
```
jbedard, line 25: Is this essentially a copy+paste of the `data` of other rules_js rules? Is there a common place in rules_js that could be linked to for a full explanation of the providers here?

Anyone without more understanding of rules_js won't understand anything beyond the first sentence here...

gregmagolan: Good point. We could expose the docstring in js_library_helpers... or a default data attribute that you can override with dicts add.

jbedard: Can we just have this first sentence followed by "See: <url to full explanation>"? Except that URL probably has to be version specific?

examples/deps_lib/BUILD.bazel
jbedard: Can the TODO in examples/BUILD.bazel also be removed?

gregmagolan: Oh right. Thanks for pointing it out. Forgot about that one and it is a different use case that is also now fixed. Updated.

ts/private/ts_lib.bzl
gregmagolan: Cleaned up with regex.

ts/private/ts_lib.bzl
gregmagolan: I ended up exposing the attr.label_list() from js_library for downstream rules to use.

ts/private/npm_repositories.bzl
```diff
@@ -26,10 +26,10 @@ def _http_archive_version_impl(rctx):
         elif "dependencies" in p.keys() and "typescript" in p["dependencies"]:
             ts = p["dependencies"]["typescript"]
         else:
-            fail("key `typescript` not found in either dependencies or devDependencies of %s" % json_path)
```
jbedard, line 29: Are the quote changes in this file correct?

gregmagolan: They are intentional. Backticks for anything that goes into docs and single quotes for CLI messages.

ts/private/ts_project.bzl
```diff
@@ -153,22 +154,28 @@ This might be because
 - `srcs` has elements producing non-ts outputs
 """ % label
         else:
-            no_outs_msg = "ts_project target %s with custom transpiler needs `declaration = True`." % label
+            no_outs_msg = "ts_project target %s with custom transpiler needs 'declaration = True'." % label
```
jbedard, line 157: Should this not use backticks?

gregmagolan: This is intentional. Backticks for anything that goes into docs and single quotes for CLI messages.

jbedard (approved):

.github/workflows/ci.yaml
gregmagolan: Fixed with 0.14.0 in BCR on same commit that will be rc4

PR #122, opened 2022-08-07
fix: add .d.mts/.d.cts in declaration info

jstuder-gh:
Update ts_project rule to include both .d.mts and .d.cts declaration
files in DeclarationInfo provider

Also add test within existing dts_lib example


gregmagolan (approved):
Thanks for the fix

PR #125, opened 2022-08-08
fix: follow symlinks in getDirectories

thesayyn:


gregmagolan (approved):
🌮

ts/test/ts_project_worker.test.js
```diff
@@ -86,6 +78,19 @@ assert.ok(!tree.directoryExists("symlinked"));
 assert.ok(!tree.directoryExists("symlinked/to"));
 
 
+fs.mkdirSync(path.join(root, "dir"));
```
thesayyn, line 81: this is the test addition. others are test refactor to ease reading

gregmagolan: 👍

PR #129, opened 2022-08-16
fix: add file info to error for `tsc` overreach

vpanta:
This data was part of this error previously but was removed in https://github.com/aspect-build/rules_ts/pull/77

I'm not sure if that was intentional, but I found it very helpful to have the data when trying to understand what was going wrong in my migration. So just a thought to add it back?

### DIFF
--- a/ts/private/ts_project_worker.js
+++ b/ts/private/ts_project_worker.js
@@ -458,7 +458,7 @@ function createProgram(args, initialInputs) {
         const relative = path.relative(execRoot, filePath);
         // external lib are transitive sources thus not listed in the inputs map reported by bazel.
         if (!filesystemTree.fileExists(path.relative(execRoot, filePath)) && !isExternalLib(filePath) && !outputs.has(relative)) {
-            throw new Error(`tsc tried to read file that wasn't an input to it.`);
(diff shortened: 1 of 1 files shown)


jbedard (approved):

thesayyn (approved):
LGTM. we removed it as tsc was yielding redundant information for path in some errors.

thesayyn:
FYI. you shouldn't be bumping into this. I'd like to know more about use cases where you bump into this error. we can improve `vfs` to direct `tsc` in the right direction.

vpanta:
I've been trying to get a POC working and this definitely helped me debug one of my issues, but I'm not positive of the specific case. I think, maybe, it was a breakage from a rule accidentally set up like this:
```
# BUILD.bazel
ts_project(
  name = "ts",
  config = ":tsconfig.json",
)

ts_config(
  name = "tsconfig",
  src = "tsconfig.json",
  deps = ["//:other_tsconfig"],
)
```
and so `tsc` wanted to load the file referenced by `//:other_tsconfig`, but the rule was misconfigured. 

It showed me a real configuration failure, but without it telling me which file was overreach, it would've been a lot harder to track down.

vpanta:
Sorry about the delay, but it's settled and the CLA is signed.

PR #130, opened 2022-08-16
fix: `ts_project` fails if `root_dir` used deep in source tree

vpanta:
Partner to https://github.com/bazelbuild/rules_nodejs/pull/3535, applying that fix here as well.

The issue is that when calculating the output, it's overzealously removing the root_dir matches. I'm happy to change this to be whatever we think is clearest, because right now it is a bit magic-number-y.

### DIFF
--- a/ts/private/ts_lib.bzl
+++ b/ts/private/ts_lib.bzl
@@ -179,7 +179,7 @@ def _out_paths(srcs, out_dir, root_dir, allow_js, ext_map):
     outs = []
     for f in srcs:
         if _is_ts_src(f, allow_js):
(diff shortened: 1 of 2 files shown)


jbedard (approved):

jbedard:
Thanks! 👍 

The CLA is still a WIP?

thesayyn (approved):

vpanta:
> Thanks! 👍
> 
> The CLA is still a WIP?

Yeah, I just sent it over to them today. Hopefully I'll hear back soon.

vpanta:
Sorry about the delay, but the CLA is signed and so this is good to go!

PR #133, opened 2022-08-26
chore: update to rules_js 1.1.0

gregmagolan:


.github/workflows/ci.yaml
```diff
@@ -17,6 +17,7 @@ jobs:
         runs-on: ubuntu-latest
 
         strategy:
+            fail-fast: false
```
jbedard, line 20: Are the changes in this file necessary?

e2e/worker/WORKSPACE
```diff
@@ -22,9 +22,10 @@ load("@aspect_rules_js//npm:npm_import.bzl", "npm_translate_lock")
 
 npm_translate_lock(
     name = "npm",
+    lifecycle_hooks_no_sandbox = False,
```
jbedard, line 25: Why was this required?

jbedard (approved):

e2e/worker/WORKSPACE
gregmagolan: It was. One of the @nestjs lifecycle hooks fails outside of the sandbox. Didn't look further into why. I'll comment there.

.github/workflows/ci.yaml
gregmagolan: bzlmod was failing on CI; we should add fast-fail throughout the repos actually. I'll prefactor it.

PR #135, opened 2022-08-27
chore: update to bazel 5.2.0

gregmagolan:
5.3.0 breaks e2e/bzlmod so we also pin the e2e tests to 5.2.0 in this PR

### DIFF
--- a/.bazelversion
+++ b/.bazelversion
@@ -1,4 +1,4 @@
-5.1.0
+5.2.0
 # The first line of this file is used by Bazelisk and Bazel to be sure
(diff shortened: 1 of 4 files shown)


.bazelversion
```diff
@@ -1,4 +1,4 @@
-5.1.0
+5.2.0
```
jbedard, line 1: Why not 5.3?

jbedard (approved):

.bazelversion
gregmagolan: 5.3.0 breaks e2e/bzlmod in a strange way

PR #138, opened 2022-08-29
Move TS config validation to validation output group

dzbarsky:
Avoid blocking the critical path (typechecking actions) on the validation. Not sure if we have a minimum bazel version we want to support?

### DIFF
--- a/ts/defs.bzl
+++ b/ts/defs.bzl
@@ -325,7 +325,6 @@ def ts_project(
             validator = validator,
             **common_kwargs
         )
-        tsc_deps = tsc_deps + ["_validate_%s_options" % name]
 
     typings_out_dir = declaration_dir if declaration_dir else out_dir
     tsbuildinfo_path = ts_build_info_file if ts_build_info_file else name + ".tsbuildinfo"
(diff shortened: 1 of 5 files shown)


alexeagle:
According to https://github.com/aspect-build/rules_ts/issues/85 we can't do this until most users can be on Bazel 6

gregmagolan:
Thanks for the PR @dzbarsky. I'll add this to the 1.1 milestone so we don't forget and mark it as DRAFT for now so it doesn't show up in our review notifications.

dzbarsky:
Sounds good! I had checked the docs but it wasn't obvious which version of Bazel had added the feature :)

alexeagle:
I think we can do this now - it's time to branch 1.x from 2.x and start landing things for 2.0

dzbarsky:
@alexeagle Finally found a few cycles to update this, PTAL

alexeagle:
Cool, could you target the 2.x branch with this, since it's a breaking change to drop Bazel 5 support?

dzbarsky:
Sure, done

alexeagle:
Nice, having your help here inspires me to get to a beta of 2.0 soon!

alexeagle (approved):

PR #146, opened 2022-09-08
fix: check for dangling symlinks

thesayyn:
CC @jfirebaugh


jbedard (approved):

jbedard:
Is there an issue or something describing the issue this actually causes? Something we can reference or maybe in the commit message?

thesayyn:
Repro is https://github.com/aspect-build/rules_ts/compare/main...jfirebaugh:rules_ts:jfirebaugh/worker_bug. We don't have an issue for this.

PR #169, opened 2022-10-07
fix: report failedLookups twice

thesayyn:


ts/private/ts_project_worker.js
```diff
@@ -588,6 +591,14 @@ function emit(args, inputs) {
 
     timingStart('applyChanges')
     host.applyChanges();
+    if (failedLookups.size) {
+        timingStart('applyChanges for failedLookups');
+        for (const [input, digest] of failedLookups) {
+            host.invalidate(input, ts.FileWatcherEventKind.Changed, digest);
+        }
+        host.applyChanges();
+        timingEnd('applyChanges for failedLookups');
```
jbedard, line 600: Should start/end things like this always be within a try/finally in case something between throws?

jbedard: I guess we don't elsewhere right now though so maybe not? Either none or do them all in a separate PR...

ts/private/ts_project_worker.js
```diff
@@ -353,7 +353,8 @@ function createProgram(args, initialInputs) {
     const filesystemTree = createFilesystemTree(execRoot, initialInputs);
     const outputs = new Set();
 
-    const taskQueue = new Array();
+    // NOTE: tsc will never clearTimeout if the index is 0 hence array needs to have at least one item.
```
jbedard, line 356: Why? Who does that?

ts/private/ts_project_worker.js
thesayyn: do we need to know how long did it take to fail? nothing in this file is supposed to fail. if it does 99% of the time it is a bug we'd have to fix.

ts/private/ts_project_worker.js
```diff
@@ -403,13 +404,13 @@ function createProgram(args, initialInputs) {
...
+        delete taskQueue[i];
     }
 
     function applyChanges() {
-        debuglog(`Applying changes ${taskQueue.length}`);
-        while(taskQueue.length) {
-            const task = taskQueue.shift();
+        for (let i = 0; i < taskQueue.length; i++) {
+            const task = taskQueue[i];
+            delete taskQueue[i]
```
jbedard, line 413: Why delete instead of `pop` or `shift` like before? This means the array length won't change and it will grow forever?

ts/private/ts_project_worker.js
thesayyn: Right. I should put a link there. https://github.com/microsoft/TypeScript/blob/d0bfd8caed521bfd24fc44960d9936a891744bb7/src/compiler/watchPublic.ts#L681 
if we return zero from `setTimeout` then this condition will never be met thus leading to incorrect behavior.

jbedard: But this is the array index, not the timeout id..?

jbedard: Ohh nm, I see it now that I look at the full file 👍

jbedard: Can we just change `setTimeout` to not have the `-1`?

ts/private/ts_project_worker.js
thesayyn: if we pop or shift we'd be changing the index of the task array. so when a new task is scheduled (setTimeout) we push the task into the array and return back the index to be used later by tsc to clean it up (clearTimeout). we throw out the task array after each apply

jbedard: I see now that I look around more... I do wonder if we should do this differently though...

ts/private/ts_project_worker.js
thesayyn: it won't have `-1` it'll have `0` which is the problem here.

ts/private/ts_project_worker.js
jbedard: Depends how the timer works. Normally if one never ends then a) it never ends and essentially leaks and b) it often messes up the next one (but that depends how it's implemented).

ts/private/ts_project_worker.js
thesayyn: > But this is the array index, not the timeout id..?

for the `setTimeout`, it is the index but for tsc, it is a number.

ts/private/ts_project_worker.js
jbedard: If the extra callback isn't a performance issue normally I'd do:

```
timer("name", () => { ... the work ... })

function timer(name, work) {
  start(name);
  try { return work() }
  finally { end(name); }
}
```

ts/private/ts_project_worker.js
thesayyn: I'd be interested to see different fake scheduling algorithms. we don't need anything fancy though. there are only two things that tsc schedules and both of them have the same `timeout` value hence no need to worry about task urgency.

ts/private/ts_project_worker.js
thesayyn: we just delegate it to `ts.performance` api which in turn delegates to node `perf` module. https://github.com/microsoft/TypeScript/blob/main/src/compiler/performance.ts

ts/private/ts_project_worker.js
jbedard: First thing that comes to mind would be:
```
let nextTimerId = 0;
function setTimeout(cb) { taskQueue[++nextTimerId] = cb; return nextTimerId; }
function clearTimeout(id) { delete taskQueue[id] }
function applyChanges() {
   for (const id of taskQueue) {
        taskQueue[id]?();
        delete taskQueue[id]
    }
}
```

However that looses the order which I assume is important, so... 🤔 

```
let nextTimerId = 1;
let lastRun = 0;
function setTimeout(cb) { taskQueue[++nextTimerId] = cb; return nextTimerId; }
function clearTimeout(id) { delete taskQueue[id] }
function applyChanges() {
   while (lastRun++ < nextTimerId)
        taskQueue[lastRun]?();
        delete taskQueue[lastRun]
    }
}
```

ts/private/ts_project_worker.js
jbedard: I'd think just returning `+1` and deleting `-1` within the set/clear would be easier, then all the "id" logic is within those methods and this array doesn't care. Those methods could change to persist `{id: Math.random(), callback}` and this array wouldn't care etc...

ts/private/ts_project_worker.js
jbedard: Interesting, those methods don't clear the data anyway so failing to call `end()` doesn't really matter 👍

ts/private/ts_project_worker.js
thesayyn: the second one could work. do you believe this is better in terms of readability? I didn't like the use of lastRun. kind of makes it harder to understand.

jbedard: You're essentially doing the exact same thing today except all the data is hidden in the array. The `nextTimerId` is the `taskQueue.length` and the `lastRun` is `taskQueue.findIndex(e => !!e)`.

Note that today the array grows forever so when you do the loop the first `lastRun-1` entries will be undefined. The array also grows forever.

thesayyn: maybe we should just put `taskQueue = new Array(1)` to the last line of applyChanges so it doesn't grow indefinitely?

jbedard: Doesn't that just continue making it more complicated? Then the `new Array(1)` thing is in multiple places. I was hoping to remove it...

jbedard: We can think about all this stuff later though, don't have to do it in this PR...

jbedard:
Can you explain how the test works? Could do it on a call later if that's better...

thesayyn:
> Can you explain how the test works? Could do it on a call later if that's better...
I'll try. btw it comes from; https://github.com/aspect-build/bazel-examples/tree/js_build_file_generation

ts/private/ts_project_worker.js
thesayyn: added logic to clear the array after each apply.

e2e/worker_workspace/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "js_build_file_generation",
```
gregmagolan, line 2: nit: remove "name" and "description" here that came in from bazel-examples

e2e/worker_workspace/test.sh
```diff
@@ -0,0 +1,9 @@
+set -x
+bazel clean
+
+buildozer "remove deps //apps/triad/pivot //apps/triad/vibe" //apps/triad:triad_ts
```
gregmagolan, line 4: nit: add a comment here that explains what this is asserting; link back to the issue and/or PR as well for future reference

gregmagolan (approved):
🌮

ts/private/ts_project_worker.js
```diff
@@ -399,21 +399,24 @@ function createProgram(args, initialInputs) {
...
             if (task) {
                 task();
             }  
         }
+        taskQueue.splice(0, taskQueue.length);
```
jbedard, line 419: How about...
```
while (0 < taskQueue.length) {
   const task = taskQueue.shift();
   ...
}
```
Then I think we can drop this line, drop the `delete` etc

jbedard: Or maybe changing the size of the array should be avoided until the end since the indexes are IDs? Maybe just leave it as is.

Could just do `taskQueue.length = 0` though.

thesayyn: tasks might schedule new tasks while being executed so the array has to stay the same until the end.

thesayyn: changed with `taskQueue.length = 0`

jbedard (approved):

PR #171, opened 2022-10-11
fix: validate preserveSymlinks tsconfig option is not set

jbedard:
I'm finding this a bit awkward to test because typescript is only fetched from `BUILD.typescript` which isn't really accessible from tests. But every test out there ensures it at least works with valid config, and I've tested the case this is catching manually. @thesayyn I don't understand how you tested workers with such minimal mocking of typescript?

### DIFF
--- a/ts/private/ts_project_options_validator.js
+++ b/ts/private/ts_project_options_validator.js
@@ -150,6 +150,17 @@ function main(_a) {
         )
         return 1
     }
+    if (options.preserveSymlinks) {
+        console.error(
+            'ERROR: ts_project rule ' +
+                target +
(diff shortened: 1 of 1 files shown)


alexeagle (approved):
Yeah I had the same issue in https://github.com/bazelbuild/rules_nodejs/tree/stable/packages/typescript/test/ts_project/validation - it's a challenge to write a negative test case for validation failures.

The logic is so simple to read and understand that I think it's fine to skip automated tests for this one.

jbedard:
@thesayyn that's what I was looking at, but the number of functions needing to be mocked was just way too high

Issue #30, opened 2022-06-02
Give better error when a ts_project dep didn't produce declarations


alexeagle:
TypeScript's default for `declaration` is false, but in Bazel we typically have dependencies on a library and want to type-check usages ~~so maybe we want to change that default?~~

gregmagolan:
Sounds reasonable if there is no extends

PR #173, opened 2022-10-12
fix: validate tsconfig dictionary values

jbedard:
CURRENT PR: Just validate to ensure nothing is invalid, tsconfig dict or not: https://github.com/aspect-build/rules_ts/pull/173/commits/0fa1c8f916309923c9adcef3a63430cb83aa9627

Ensure tsconfig dict() and ts_project attributes don't collide: https://github.com/aspect-build/rules_ts/commit/a510a3c29e52566ad6359bbd073dfec3dc3dad7d

### DIFF
--- a/ts/defs.bzl
+++ b/ts/defs.bzl
@@ -325,7 +325,7 @@ def ts_project(
         # user supplied a tsconfig.json InputArtifact
         tsconfig = "tsconfig_%s.json" % name
 
-    elif validate:
+    if validate:
         validate_options(
             name = "_validate_%s_options" % name,
             target = "//%s:%s" % (native.package_name(), name),


alexeagle (commented):
Did you test it out?

I don't think this is right, the validator here is to check that the options in the tsconfig.json file match the attributes passed to ts_project. What specifically are you trying to validate?

alexeagle:
Err, sorry, going too fast. I think you're just fixing the TODO you've commented out. Yes, if you tested this then it makes sense to me.

jbedard:
Actually it might not even fully fulfill what the TODO is mentioning... when there's a `sourceMap: true` within a tsconfig dict it overwrites the `ts_project(source_map)`. Should it fail if a `ts_project(source_map)` was passed and doesn't match the tsconfig dict? I think that's the original TODO. But that sounds awkward... things like `def ts_project(source_map = False)` would have to change to `def ts_project(source_map = None)` so you can determine if it was passed and invalid vs default that should be overridden. IDK if that's worth it... @thesayyn any opinion or preference?

Either way I think validating the config is a good idea, generated tsconfig or not, may as well be consistent and validate it imo 🤷

jbedard:
Note that I was looking at https://github.com/aspect-build/rules_ts/issues/30 when I did this which might require the ugly `= None` anyway 🤷

thesayyn:
`attr#tsconfig.sourceMap` and `attr#sourceMap` should be canonical IMHO and users shall not provide both. `attr#sourceMap` is there to help us declare outputs correctly. I say `attr#tsconfig.sourceMap` should take precedence over `attr#sourceMap`.

jbedard:
I've reverted back to the original idea of just validating the options. Validating tsconfig vs attributes having no conflicts and informing the user via `fail` can be done another time. See the PR description...

alexeagle (approved):

PR #174, opened 2022-10-12
Chore/export ts config

sallustfire:
This adds the TsConfigInfo as provider to the public rules interface, which allows downstream rules to reuse the output of `ts_config`.

For reference there is a parallel PR in rules_nodejs: https://github.com/bazelbuild/rules_nodejs/pull/3578#issue-1406456330


alexeagle (approved):

jbedard:
@sallustfire can you update the docs as well? See the CI error...

sallustfire:
@jbedard Ok, regenerated the docs, passes locally for me.

PR #176, opened 2022-10-12
fix: allow composite projects

### LINKED ISSUE #163: Request for project-references examples (if supported)
It's unclear to me whether https://www.typescriptlang.org/docs/handbook/project-references.html (incremental builds) are supported by just looking at the examples https://github.com/aspect-build/rules_ts/tree/d5842963bae0c1f0e83a77dabee595784418917b/examples.

I would have expected an example with multiple tsconfig.json's, with some tsconfig.json's referring to other tsconfig.json's via the references config (https://www.typescriptlang.org/tsconfig#references)

My main goal is to get yarn workspaces working with typescript and bazel

Edit: TLDR, my question is whether aspect ts rules support project-references

alexeagle:
We need to propagate the tsconfig.json file to composite project dependents as an input. This was in rules_nodejs but we lost it.

Also correct the pathing for the tsbuildinfo output, which hadn't been tested in rules_ts yet.

Copied the three composite projects from rules_nodejs tests, so we have that coverage now.

Fixes #163

### DIFF
--- a/ts/private/ts_project.bzl
+++ b/ts/private/ts_project.bzl
@@ -89,6 +89,13 @@ def _ts_project_impl(ctx):
 
     inputs = srcs_inputs[:]
     for dep in ctx.attr.deps:
+        # When TypeScript builds a composite project, our compilation will want to read the tsconfig.json of
+        # composite projects we reference.
+        # Otherwise we'd get an error like
+        # examples/project_references/lib_b/tsconfig.json(5,9): error TS6053:
+        # File 'execroot/aspect_rules_ts/bazel-out/k8-fastbuild/bin/examples/project_references/lib_a/tsconfig.json' not found.
+        if ctx.attr.composite and TsConfigInfo in dep:
+            inputs.extend(dep[TsConfigInfo].deps.to_list())
         if ValidOptionsInfo in dep:
             inputs.append(dep[ValidOptionsInfo].marker)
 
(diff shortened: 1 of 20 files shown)


ts/private/ts_project.bzl
```diff
@@ -89,6 +89,13 @@ def _ts_project_impl(ctx):
 
     inputs = srcs_inputs[:]
     for dep in ctx.attr.deps:
+        # When TypeScript builds a composite project, our compilation will want to read the tsconfig.json of
+        # composite projects we reference.
+        # Otherwise we'd get an error like
+        # examples/project_references/lib_b/tsconfig.json(5,9): error TS6053:
+        # File 'execroot/aspect_rules_ts/bazel-out/k8-fastbuild/bin/examples/project_references/lib_a/tsconfig.json' not found.
+        if ctx.attr.composite and TsConfigInfo in dep:
+            inputs.extend(dep[TsConfigInfo].deps.to_list())
```
thesayyn, line 98: nit: we might want to have another pass on ts_project to remove `to_list` usages.

thesayyn (approved):
lgtm

ts/private/ts_project.bzl
alexeagle: yes, that's https://github.com/aspect-build/rules_ts/issues/177, I'll pick it up next :)

thesayyn: oh... great!

PR #178, opened 2022-10-12
refactor: TsConfigInfo uses depsets

### LINKED ISSUE #177: port tsconfig memory leak fix
from https://github.com/bazelbuild/rules_nodejs/pull/3431

alexeagle:
When we pass to copy_files_to_bin_actions, we're required to convert these to a list, which is too bad. I don't have a big repo handy to run a bazel dump to profile how much memory this actually saves, unlike the contribution to rules_nodejs this is based on: https://github.com/bazelbuild/rules_nodejs/pull/3431

Closes #177

### DIFF
--- a/ts/private/ts_config.bzl
+++ b/ts/private/ts_config.bzl
@@ -29,13 +29,13 @@ TsConfigInfo = provider(
 
 def _ts_config_impl(ctx):
     files = [copy_file_to_bin_action(ctx, ctx.file.src)]
-    transitive_deps = []
+    transitive_deps = [depset(copy_files_to_bin_actions(ctx, ctx.files.deps))]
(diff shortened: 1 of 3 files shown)


thesayyn (approved):
LGTM. looks like it would be hard to change copy_files_to_bin_actions since it calls `declare_file` for each source.

alexeagle:
Leaving open for a bit in case @gregmagolan comes to look

gregmagolan:
> LGTM. looks like it would be hard to change copy_files_to_bin_actions since it calls `declare_file` for each source.

Yeah. You need to flatten the depset in copy_files_to_bin_actions to iterator of the File objects.

gregmagolan (approved):

PR #181, opened 2022-10-17
feat: enable declarations by default when unspecified in tsconfig dict

### LINKED ISSUE #30: Give better error when a ts_project dep didn't produce declarations

jbedard:
Fix #30


ts/defs.bzl
```diff
@@ -291,6 +291,9 @@ def ts_project(
             tsconfig = "tsconfig.json"
 
     if type(tsconfig) == type(dict()):
+        # Output declarations by default if unspecified and not extending a config which may specify
+        declaration = False if declaration == False or extends else True
```
thesayyn, line 295: had to read four times to understand the outcome 🗡️

jbedard: Have a recommendation or preference to use instead?

jbedard:
We decided to create a better error message instead of sometimes overriding the tsconfig, see the change to https://github.com/aspect-build/rules_ts/issues/30

I'll open a new PR...

Issue #114, opened 2022-08-03
migrate ts_project_worker.js to typescript


thesayyn:
We started off with js to keep it simple but this decision makes it harder to review changes

alexeagle:
We are moving away from supporting the Persistent Worker in the next major 2.0 release of rules_ts, and likely will never fix this, sorry!

PR #184, opened 2022-10-21
fix: prefix host call

thesayyn:


jbedard (approved):

jbedard:
Was this crashing?

thesayyn:
> Was this crashing?

unfortunately, yes. and it went unnoticed. I guess it's time for #114

jbedard:
> unfortunately, yes. and it went unnoticed. I guess it's time for https://github.com/aspect-build/rules_ts/issues/114

I think we can even write plain js but still typecheck it by putting types in comments and just type inference on its own is useful.

thesayyn:
> > unfortunately, yes. and it went unnoticed. I guess it's time for #114
> 
> I think we can even write plain js but still typecheck it by putting types in comments and just type inference on its own is useful.

i agree.

PR #192, opened 2022-10-26
fix: allow customizing the npm registry of the core typescript npm packages

### LINKED ISSUE #191: can not customize npm_dependencies URLs
With rules_js `.npmrc` can be used to customize registry URLs. Customizing the typescript URLs in rules_ts is difficult and undocumented, see ts/private/npm_repositories.bzl

jbedard:
Fixes #191

When a custom npm registry is used the same must be done for the rules_ts npm packages as the rules_js packages. Normally it's done by reading the npmrc file but that isn't available when fetching in the WORKSPACE. Would this be the ideal solution? Ideas?

### DIFF
--- a/docs/repositories.md
+++ b/docs/repositories.md
@@ -11,7 +11,7 @@ See https://docs.bazel.build/versions/main/skylark/deploying.html#dependencies
 ## rules_ts_dependencies
 
 <pre>
-rules_ts_dependencies(<a href="#rules_ts_dependencies-ts_version_from">ts_version_from</a>, <a href="#rules_ts_dependencies-ts_version">ts_version</a>, <a href="#rules_ts_dependencies-ts_integrity">ts_integrity</a>)
+rules_ts_dependencies(<a href="#rules_ts_dependencies-ts_version_from">ts_version_from</a>, <a href="#rules_ts_dependencies-ts_version">ts_version</a>, <a href="#rules_ts_dependencies-ts_integrity">ts_integrity</a>, <a href="#rules_ts_dependencies-npm_registry">npm_registry</a>)
 </pre>
 
 Dependencies needed by users of rules_ts.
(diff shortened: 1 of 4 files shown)


kormide (approved):
LGTM

kormide:
> Normally it's done by reading the npmrc file but that isn't available when fetching in the WORKSPACE

Could you elaborate on that? Why isn't npmrc available?

jbedard:
We can't read files during the loading phase when the WORKSPACE/BUILD files are essentially just being parsed and put into a dep graph.

The information we need is normally read from a npmrc file when yarn/npm/pnpm run, we parse that npmrc at the same time as the lockfile: https://github.com/aspect-build/rules_js/blob/c6c257ac82f9030ee4443f7ed1db775f056815b6/npm/private/npm_translate_lock.bzl#L480-L486

This PR is the best way I could think of customizing the npm registry URL in the loading phase. I'd be curious what other ideas there are though...

kormide:
> We can't read files during the loading phase when the WORKSPACE/BUILD files are essentially just being parsed and put into a dep graph.
> 
> The information we need is normally read from a npmrc file when yarn/npm/pnpm run, we parse that npmrc at the same time as the lockfile: https://github.com/aspect-build/rules_js/blob/c6c257ac82f9030ee4443f7ed1db775f056815b6/npm/private/npm_translate_lock.bzl#L480-L486
> 
> This PR is the best way I could think of customizing the npm registry URL in the loading phase. I'd be curious what other ideas there are though...

But it can be loaded in a repository rule or module_ctx before anything is downloaded? (ignore me if that doesn't apply here, I don't have a deep understanding of what's going on).

jbedard:
> But it can be loaded in a repository rule or module_ctx before anything is downloaded?

Yeah I think you're correct, we can pass the npmrc just like we do for `npm_translate_lock`. That means we'd have to fetch `@bazel/worker` and `google-protobuf` with our own repo rule as well instead of using the normal `http_archive` (today only the typescript npm package is downloaded with our custom [http_archive_version](https://github.com/aspect-build/rules_ts/blob/1cbe54029c00241a0ea399b97d0b4d4175cee8a8/ts/private/npm_repositories.bzl#L18-L68).

@gregmagolan WDYT?

thesayyn (approved):

thesayyn:
I wish we had a mechanism to configure npm registry in rules_js and all the other rules follow the same pattern. 
As a user, I'd expect that one configuration `.npmrc` is sufficient to configure `rules_js` and derivative rules.

jbedard:
After talking with @gregmagolan we've decided we should handle this differently like @kormide and @thesayyn are both suggesting.

The npmrc can have various other config in it that might effect downloading npm packages such as CA certs, registry URLs, registry auth etc, so we should come up with a solution that respects all of those.

PR #198, opened 2022-10-27
chore: update to rules_js 1.6.2

gregmagolan:


alexeagle:
maybe moved and I have #197

PR #199, opened 2022-10-27
feat: add ts_project(deps) validation ensuring all deps of declarations

jbedard:
Currently causes some errors which we should discuss...


ts/private/ts_validate_options.bzl
```diff
@@ -23,6 +24,13 @@ def _validate_options_impl(ctx):
     # We make it a .d.ts file so we can plumb it to the deps of the ts_project compile.
     marker = ctx.actions.declare_file("%s.optionsvalid.d.ts" % ctx.label.name)
 
+    # Provider validation
+    if not ctx.attr.allow_js:
+        for d in ctx.attr.deps:
+            if not d[JsInfo].declarations:
```
thesayyn, line 30: I am not sure what would happen if there were an empty depset. would this condition still be true?

ts/private/ts_validate_options.bzl
```diff
@@ -56,6 +64,7 @@ def _validate_options_impl(ctx):
     ]
 
 _ATTRS = dict(COMPILER_OPTION_ATTRS, **{
+    "deps": attr.label_list(mandatory = True, providers = [JsInfo]),
```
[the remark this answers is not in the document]
jbedard: Yes. Today the `ts_project(deps)` also has that, although `ts_project` also has `ValidOptionsInfo` so maybe we need to either add that here as well or filter it out?

ts/private/ts_validate_options.bzl
jbedard: An empty `depset` will evaluate to false. This is where the CI errors are failing atm.

https://github.com/aspect-build/rules_js/pull/563 fixes the CI errors failing due to npm packages missing `declarations`, but there's other failures we should discuss.

thesayyn: gotcha. Afaik rules_js does not put declarations into `declarations`. It would have to do this at analysis time but there is no way to know what the package contains during analysis time.

jbedard: rules_js does it for `js_library` if there are dts, json, or directories with unknown content. I think that's an issue because if that changes (no more dts for example) then it will suddenly cause errors in `ts_project`s elsewhere. Maybe that's not an issue though?

rules_js will also do it for linked npm packages after https://github.com/aspect-build/rules_js/pull/563

I think that's all the places rules_js sets `JsInfo` so other then my concern with `js_library` it might be ok?

ts/private/ts_validate_options.bzl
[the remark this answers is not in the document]
alexeagle: ValidOptionsInfo is what's produced from this rule, so you have it right, only JsInfo is required here (it's the only one you read from, as well)

ts/private/ts_validate_options.bzl
```diff
@@ -23,6 +24,20 @@ def _validate_options_impl(ctx):
     # We make it a .d.ts file so we can plumb it to the deps of the ts_project compile.
     marker = ctx.actions.declare_file("%s.optionsvalid.d.ts" % ctx.label.name)
 
+    # Provider validation
+    if not ctx.attr.allow_js:
```
alexeagle, line 28: I think checkJs is the feature we ought to make this conditional on, see the definition of the two of them.

However that would mean adding an attribute to ts_project, and maybe allowJs is always true when checkJs is true (you could verify that's what `tsc` does) so then allow_js here just makes our disabling a bit broader and it's fine

alexeagle:
There are some cases where the right answer to green this up is to turn off the validator. Possibly just do it in all cases as long as they don't look like a bug.

ts/private/ts_validate_options.bzl
```diff
@@ -23,6 +24,20 @@ def _validate_options_impl(ctx):
...
+        for d in ctx.attr.deps:
+            if not d[JsInfo].declarations:
+                fail("""\
+ts_project '{0}' does not contain any declarations (.d.ts or other type-check files).
```
jbedard, line 32: @alexeagle I think assuming this dep is a `ts_project` might be confusing in the cases it is not such as the CI failure I was getting (depending on a `js_library` that has no declarations).

Maybe... "ts_project {1} dependency {0} does not contain ..."?

examples/path_mappings/BUILD.bazel
```diff
@@ -18,6 +18,7 @@ ts_project(
     srcs = ["foo.ts"],
     declaration = True,
     tsconfig = "tsconfig.json",
+    validate = False,
```
alexeagle, line 21: could we change the deps here to `["//examples/dts_lib"]` instead?

ts/private/ts_validate_options.bzl
jbedard: I'm fixing things up atm and will do it

examples/path_mappings/BUILD.bazel
alexeagle: err no, I guess the point of this example is to use path mapping rather than dep on a linked npm_package rule, so it should be `["//examples/dts_lib:lib"]` - but still doesn't seem like the js_lib should actually be needed in the deps so the validator is telling us a correct thing here?

jbedard: The `js_lib` is a compile-time dependency so no. This is the case that I think will bother people...

alexeagle: okay, tried it locally and I see now, `tsc` wants to do the resolution even though the type-checker gives up and makes it an `any`. I think validate = False is correct then 👍🏻

ts/private/ts_validate_options.bzl
```diff
@@ -23,6 +24,20 @@ def _validate_options_impl(ctx):
...
+        for d in ctx.attr.deps:
+            if not d[JsInfo].declarations:
+                fail("""\
+ts_project '{1}' dependency '{0}' does does not contain any declarations (.d.ts or other type-check files).
+Generally, targets which produce no declarations aren't useful as dependencies to the TypeScript type-checker.
+This likely means you forgot to set 'declaration = true' in the compilerOptions for that target.
+
+To disable this check, set the validate attribute to False:
+  npx @bazel/buildozer 'set validate False' {1}
+""".format(d.label, ctx.attr.target))
```
alexeagle, line 38: could you just post a comment of what this error looks like now when a user hits it, and test that the buildozer command works to repair if you remove the validate = False from that example?

jbedard:
```
Error in fail: ts_project '//examples/path_mappings:path_mappings' dependency '//examples/js_lib:lib' does does not contain any declarations (.d.ts or other type-check files).
Generally, targets which produce no declarations aren't useful as dependencies to the TypeScript type-checker.
This likely means you forgot to set 'declaration = true' in the compilerOptions for that target.

To disable this check, set the validate attribute to False:
  npx @bazel/buildozer 'set validate False' //examples/path_mappings:path_mappings
```

Note originally this was `d` instead of `d.label` so the message had `<target //examples/path_mappings:path_mappings>`

alexeagle: nice

ts/private/ts_validate_options.bzl
alexeagle: any thought on this one before we merge?

alexeagle (approved):

PR #211, opened 2022-10-31
test: add case for filegroup of srcs

alexeagle:
Part of #23


jbedard:
How about generated files? A `genrule` that just copies the ts files?

alexeagle:
We have that in `examples/simple` already. I'll move it over here since generated code isn't the simple case.

jbedard:
In this PR or another?

jbedard (approved):

alexeagle:
moved it here, PTAL

jbedard (approved):
👍

PR #213, opened 2022-10-31
docs: add custom compiler example

alexeagle:
part of #23

### DIFF
--- a/examples/package.json
+++ b/examples/package.json
@@ -1,6 +1,9 @@
 {
     "private": true,
     "dependencies": {
+        "@angular/core": "14",
+        "@angular/compiler": "14",
+        "@angular/compiler-cli": "14",
         "@babel/cli": "7.18.9",
         "@babel/core": "7.18.9",
         "@babel/parser": "7.18.9",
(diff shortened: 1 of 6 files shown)


examples/package.json
```diff
@@ -1,6 +1,9 @@
 {
     "private": true,
     "dependencies": {
+        "@angular/core": "14",
```
jbedard, line 4: Aren't these at 16 something now?

examples/package.json
```diff
@@ -1,6 +1,9 @@
...
+        "@angular/compiler": "14",
```
jbedard, line 5: Is this used?

jbedard (approved):

examples/package.json
alexeagle: well... it's a peerDep of angular/compiler-cli so it complains, and I assume it would error since the compiler-cli would be missing somethings

examples/package.json
alexeagle: nope, not according to https://www.npmjs.com/package/@angular/compiler-cli

PR #216, opened 2022-11-01
chore: fixes to doc presentation

alexeagle:


gregmagolan (approved):

ts/defs.bzl
```diff
@@ -163,24 +163,28 @@ def ts_project(
...
+
+            By default the out_dir is the package's folder under bazel-out.
 
         tsc: Label of the TypeScript compiler binary to run.
             This allows you to use a custom compiler.
         tsc_worker: Label of a custom TypeScript compiler binary which understands Bazel's persistent worker protocol.
         validator: Label of the tsconfig validator to run when `validate = True`.
         allow_js: Whether TypeScript will read .js and .jsx files.
             When used with `declaration`, TypeScript will generate `.d.ts` files from `.js` files.
-        resolve_json_module: None | Boolean; Specifies whether TypeScript will read .json files. Defaults to None.
```
jbedard, line 177: This makes me wonder if this param is actually required? 🤔 

It's essentially only used for determining the default `srcs`...

jbedard (approved):

ts/defs.bzl
alexeagle: It's not required, default is None

jbedard: I mean required at all. Can we just delete the param and code for it? The default `srcs` can not longer use it, but that's just the defaults and imo not worth a param like this...

Issue #228, opened 2022-11-04
[Bug]: Workers broken on Windows


alexeagle:
### What happened?

With workers
https://buildkite.com/bazel/bcr-presubmit/builds/674#018443a2-60ec-4c83-b51b-021568de0748

```
(17:15:06) ERROR: C:/b/bk-windows-r2x7/bazel-downstream-projects/output/rules_ts-1.0.0-rc5/e2e/bzlmod/BUILD:4:11: Compiling TypeScript project @@//:ts [tsc -p tsconfig.json] failed: Worker process did not return a WorkResponse:
  |  
  | ---8<---8<--- Start of log, file at C:/b/kvdsqqpu/bazel-workers/worker-1-TsProject.log ---8<---8<---
  | node:internal/modules/cjs/loader:936
  | throw err;
  | ^
  |  
  | Error: Cannot find module 'google-protobuf'
  | Require stack:
  | - C:\b\kvdsqqpu\execroot\_main\bazel-out\x64_windows-opt-exec-2B5CBBC6\bin\external\aspect_rules_ts~override~ext~npm_typescript\tsc_worker.bat.runfiles\aspect_rules_ts~override~ext~npm_typescript\node_modules\@bazel\worker\src\worker_protocol.js
  | - C:\b\kvdsqqpu\execroot\_main\bazel-out\x64_windows-opt-exec-2B5CBBC6\bin\external\aspect_rules_ts~override~ext~npm_typescript\tsc_worker.bat.runfiles\aspect_rules_ts~override~ext~npm_typescript\node_modules\@bazel\worker\src\worker.js
  | - C:\b\kvdsqqpu\execroot\_main\bazel-out\x64_windows-opt-exec-2B5CBBC6\bin\external\aspect_rules_ts~override~ext~npm_typescript\tsc_worker.bat.runfiles\aspect_rules_ts~override~ext~npm_typescript\node_modules\@bazel\worker\index.js
  | - C:\b\kvdsqqpu\execroot\_main\bazel-out\x64_windows-opt-exec-2B5CBBC6\bin\external\aspect_rules_ts~override~ext~npm_typescript\tsc_worker.bat.runfiles\aspect_rules_ts~override~ext~npm_typescript\ts_project_worker.js
 
```

Without workers
https://buildkite.com/bazel/bcr-presubmit/builds/681#018443d0-fe66-4fd5-aa90-c872f24458f7

### Version

rules_ts 1.0.0-rc5

### Fund our work

[...]

PR #229, opened 2022-11-04
chore: updates for bzlmod launch

alexeagle:
- we are broken on windows, https://github.com/aspect-build/rules_ts/issues/228
- add the snippet so our next release shows how to use from BCR


kormide:
Few things left to do:
* Move `bcr` folder to root, change to `.bcr`
* Delete bcr github actions scripts
* Set a fixedReleaser in `.bcr/config.yml`

alexeagle:
okay I'll do those here while I'm in these files

.bcr/config.yaml
```diff
@@ -0,0 +1 @@
+fixedReleaser: alexeagle
```
kormide, line 1: The schema for this changed. `fixedReleaser` now takes `login` and `email` fields.

kormide: Also, I think the `.yml` extension is assumed. I should make that more flexible..

kormide:
Just need to rename config.yaml -> config.yml.

kormide (approved):

PR #230, opened 2022-11-04
fix: document tsc with more heap

alexeagle:


ts/defs.bzl
```diff
@@ -171,7 +171,11 @@ def ts_project(
             By default the out_dir is the package's folder under bazel-out.
 
         tsc: Label of the TypeScript compiler binary to run.
-            This allows you to use a custom compiler.
+            This allows you to use a different js_binary rule to invoke `tsc`, or run some other compiler which is API-
```
jbedard, line 174:
```suggestion
            This allows you to use a custom API-compatible compiler in place of the regular `tsc` such as a custom `js_binary` or Angular's `ngc`.
```

Just not sure about the "or some other" 🤷

jbedard (approved):

Issue #220, opened 2022-11-02
[Bug]: ts_project can not have a `select` statement in srcs or deps


mattem:
### What happened?

It's not possible to use a `select` in the `srcs` or `deps` of a `ts_project` rule. 

Apply the patch below and run `bazel build //examples/transpiler:babel`, not the error:

```
ERROR: Traceback (most recent call last):
	File "/Users/matt/workspace/rules_ts/examples/transpiler/BUILD.bazel", line 28, column 11, in <toplevel>
		ts_project(
	File "/Users/matt/workspace/rules_ts/ts/defs.bzl", line 307, column 37, in ts_project
		js_outs = _lib.calculate_js_outs(srcs, out_dir, root_dir, allow_js, preserve_jsx, emit_declaration_only)
	File "/Users/matt/workspace/rules_ts/ts/private/ts_lib.bzl", line 206, column 22, in _calculate_js_outs
		return _out_paths(srcs, out_dir, root_dir, allow_js, exts)
	File "/Users/matt/workspace/rules_ts/ts/private/ts_lib.bzl", line 180, column 14, in _out_paths
		for f in srcs:
Error: type 'select' is not iterable
```

This is made double tricky to hold as #219 prevents placing the `select` on a `filegroup` and adding that to `ts_project`

### Version

HEAD

### How to reproduce

```shell
diff --git a/examples/transpiler/BUILD.bazel b/examples/transpiler/BUILD.bazel
index d9e1ede..74e6fc9 100644
--- a/examples/transpiler/BUILD.bazel
+++ b/examples/transpiler/BUILD.bazel
@@ -18,11 +18,18 @@ write_file(
     ],
 )

+config_setting(
+    name = "stamp",
+    values = {"stamp": "true"},
+)
+
 # Runs babel to transpile ts -> js
 # and tsc to type-check
 ts_project(
     name = "babel",
-    srcs = ["big.ts"],
+    srcs = select({
+        ":stamp": ["big.ts"],
+    }),
     declaration = True,
     out_dir = "build-babel",
     source_map = True,
```

### Fund our work

- [ ] Sponsor our open source work by donating a [bug bounty](https://opencollective.com/aspect-build/)

arrdem:
Closing since there appears to be a workaround available.

PR #236, opened 2022-11-09
fix: support non-file transpiler srcs

### LINKED ISSUE #219: [Bug]: can't pass srcs to ts_project via another rule when transpiler is used
### What happened?

When using the `transpiler` attr on `ts_project`, it's not possible to pass `srcs` to the rule that are an output of another without referencing them by named output.

This is caused by the assumptions made in the `_is_ts_src` (and related methods) that it is working directly with labels that match source file names, https://github.com/aspect-build/rules_ts/blob/main/ts/private/ts_lib.bzl#L152-L162

### Version

HEAD

### How to reproduce
Apply the following patch:

```shell
diff --git a/examples/transpiler/BUILD.bazel b/examples/transpiler/BUILD.bazel
index d9e1ede..694333b 100644
--- a/examples/transpiler/BUILD.bazel
+++ b/examples/transpiler/BUILD.bazel
@@ -18,11 +18,16 @@ write_file(
     ],
 )

+filegroup(
+    name = "srcs",
+    srcs = ["big.ts"],
+)
+
 # Runs babel to transpile ts -> js
 # and tsc to type-check
 ts_project(
     name = "babel",
-    srcs = ["big.ts"],
+    srcs = [":srcs"],
     declaration = True,
     out_dir = "build-babel",
     source_map = True,
```

Then run:

```
$ bazel build //examples/transpiler:babel

INFO: Analyzed target //examples/transpiler:babel (1 packages loaded, 7 targets configured).
INFO: Found 1 target...
Target //examples/transpiler:babel up-to-date (nothing to build)
INFO: Elapsed time: 0.223s, Critical Path: 0.00s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
```

Note how bazel reports there is nothing to build.

### Any other information?

_No response_

### Fund our work

- [ ] Sponsor our open source work by donating a [bug bounty](https://opencollective.com/aspect-build/)

jbedard:
The current transpiler which takes 3 arrays with aligning indexes is a bad API, and to fix #219 one of those arrays will no longer align with the other 2 making it unusable.

Proposed solution in this PR: https://github.com/aspect-build/rules_ts/issues/219#issuecomment-1309178614

This change makes the transpiler API simply `transpiler(srcs)`. Any additional information required should be added via `partial.make` such as `out_dir`, `source_map` etc. Often these extra params will be required in order to align with the `ts_project` attributes. A macro will almost always be ideal to configure the transpiler including things like aligning these attributes.

swc change to do its own output declarations: https://github.com/aspect-build/rules_swc/commit/f7169623c180bd55aac4f60bc11d879bb83d81d4

[ts/test/mock_transpiler.bzl](https://github.com/aspect-build/rules_ts/pull/236/files#diff-75f4efb350a10f1cde0b94a92e8fea9b83bc39bfad099ccfa9ca31527778c9df) is an example implementing a full rule including support for non-file srcs and predeclaring the .js and .js.map outputs.

[examples/transpiler/babel.bzl](https://github.com/aspect-build/rules_ts/pull/236/files#diff-c4d1b455609d9c2374c64765b2de93cd9a34e615a04603c24bb2ffc22ad00f03) is an example of a macro keeping the same functionality as today where it only supports file inputs. This also predeclares the .js and .js.map outputs as part of the babel invocation.

Alternatives:
* ~~changing `js/map_outs` to maps is a lot nicer then 3 arrays where the indexes align, especially since fixing targets (non-files) in `srcs` breaks the index-aligning. See https://github.com/aspect-build/rules_ts/commit/17a80d5082ca297de23197e7f7ffdb0018eb29fb~~

### DIFF
--- a/docs/transpiler.md
+++ b/docs/transpiler.md
@@ -20,13 +20,16 @@ Read more: https://blog.aspect.dev/typescript-speedup
 The `transpiler` attribute of `ts_project` lets you select which tool produces the JavaScript outputs. The default value of `None` means that `tsc` should do transpiling along with type-checking, as this is the simplest configuration without additional dependencies. However as noted above, it's also the slowest.
 
 The `transpiler` attribute accepts a rule or macro with this signature:
-`name, srcs, js_outs, map_outs, **kwargs`
+`name, srcs, **kwargs`
 where the `**kwargs` attribute propagates the tags, visibility, and testonly attributes from `ts_project`.
 
-If you need to pass additional attributes to the transpiler rule, you can use a
+If you need to pass additional attributes to the transpiler rule such as `out_dir`, you can use a
 [partial](https://github.com/bazelbuild/bazel-skylib/blob/main/lib/partial.bzl)
 to bind those arguments at the "make site", then pass that partial to this attribute where it will be called with the remaining arguments.
-
+The transpiler rule or macro is responsible for predicting and declaring outputs.
+If you want to pre-declare the outputs (so that they can be addressed by a Bazel label, like `//path/to:index.js`)
+then you should use a macro which calculates the predicted outputs and supplies them to a `ctx.attr.outputs` attribute
+on the rule.
 See the examples/transpiler directory for a simple example using Babel, or
 <https://github.com/aspect-build/bazel-examples/tree/main/ts_project_transpiler>
 for a more complete example that also shows usage of SWC.
--- a/examples/transpiler/babel.bzl
+++ b/examples/transpiler/babel.bzl
@@ -1,6 +1,5 @@
 "Adapter from the Babel CLI to the ts_project#transpiler interface"
 
-load("@aspect_rules_js//js:defs.bzl", "js_library")
 load("@npm//examples:@babel/cli/package_json.bzl", "bin")
 
 # buildifier: disable=function-docstring
@@ -24,6 +23,7 @@ def babel(name, srcs, **kwargs):
         if not src.endswith(".ts"):
             fail("babel example transpiler only supports source .ts files")
 
+        # Predict the output paths where babel will write
         js_out = src.replace(".ts", ".js")
         map_out = src.replace(".ts", ".js.map")
 
@@ -51,8 +51,8 @@ def babel(name, srcs, **kwargs):
         outs.append(js_out)
         outs.append(map_out)
 
-    # The target producing the js files which ts_project() will reference
-    js_library(
+    # The target whose default outputs are the js files which ts_project() will reference
+    native.filegroup(
         name = name,
         srcs = outs,
     )
--- a/examples/transpiler/babel.bzl
+++ b/examples/transpiler/babel.bzl
@@ -1,32 +1,40 @@
 "Adapter from the Babel CLI to the ts_project#transpiler interface"
 
+load("@aspect_rules_js//js:defs.bzl", "js_library")
 load("@npm//examples:@babel/cli/package_json.bzl", "bin")
 
 # buildifier: disable=function-docstring
-def babel(name, srcs, js_outs, map_outs, **kwargs):
+def babel(name, srcs, **kwargs):
     # rules_js runs under the output tree in bazel-out/[arch]/bin
     # and we additionally chdir to the examples/ folder beneath that.
     execroot = "../../../.."
 
-    # In this example we compile each file individually.
+    outs = []
+
+    # In this example we compile each file individually on .ts src files.
+    # The src files must be .ts files known at the loading phase in order
+    # to setup the babel compilation for each .ts file.
+    #
     # You might instead use a single babel_cli call to compile
     # a directory of sources into an output directory,
     # but you'll need to:
     # - make sure the input directory only contains files listed in srcs
     # - make sure the js_outs are actually created in the expected path
     for idx, src in enumerate(srcs):
+        if not src.endswith(".ts"):
+            fail("babel example transpiler only supports source .ts files")
+
+        js_out = src.replace(".ts", ".js")
+        map_out = src.replace(".ts", ".js.map")
+
         # see https://babeljs.io/docs/en/babel-cli
         args = [
             "{}/$(location {})".format(execroot, src),
             "--presets=@babel/preset-typescript",
             "--out-file",
-            "{}/$(location {})".format(execroot, js_outs[idx]),
+            "{}/$(location {})".format(execroot, js_out),
+            "--source-maps",
         ]
-        outs = [js_outs[idx]]
-
-        if len(map_outs) > 0:
-            args.append("--source-maps")
-            outs.append(map_outs[idx])
 
         bin.babel(
             name = "{}_{}".format(name, idx),
@@ -35,7 +43,16 @@ def babel(name, srcs, js_outs, map_outs, **kwargs):
                 "//examples:node_modules/@babel/preset-typescript",
             ],
             chdir = "examples",
-            outs = outs,
+            outs = [js_out, map_out],
             args = args,
             **kwargs
         )
+
+        outs.append(js_out)
+        outs.append(map_out)
+
+    # The target producing the js files which ts_project() will reference
+    js_library(
+        name = name,
+        srcs = outs,
+    )
--- a/ts/private/ts_lib.bzl
+++ b/ts/private/ts_lib.bzl
@@ -187,7 +187,7 @@ def _out_paths(srcs, out_dir, root_dir, allow_js, ext_map):
                 outs.append(out)
     return outs
 
-def _calculate_js_outs(srcs, out_dir, root_dir, allow_js, preserve_jsx, emit_declaration_only):
+def _calculate_js_outs(srcs, out_dir = ".", root_dir = ".", allow_js = False, preserve_jsx = False, emit_declaration_only = False):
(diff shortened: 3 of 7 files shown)


ts/private/ts_lib.bzl
```diff
@@ -176,20 +176,20 @@ def _replace_ext(f, ext_map):
...
+    outs = {}
     for f in srcs:
         if _is_ts_src(f, allow_js):
             out = _join(out_dir, f[:f.rindex(".")].replace(rootdir_replace_pattern, "", 1) + _replace_ext(f, ext_map))
 
             # Don't declare outputs that collide with inputs
             # for example, a.js -> a.js
             if out != f:
-                outs.append(out)
+                outs.setdefault(f, out)
```
alexeagle, line 187: this seems to permit collisions, is that intentional?

alexeagle (commented):
Could you add the documentation change as well, anywhere that the transpiler function signature is referenced? (the transpiler attribute and the separate transpiler.md file)

will help to understand if it's the right shape based on how we explain to users.

examples/transpiler/babel.bzl
```diff
@@ -15,18 +15,21 @@ def babel(name, srcs, js_outs, map_outs, **kwargs):
     # - make sure the input directory only contains files listed in srcs
     # - make sure the js_outs are actually created in the expected path
     for idx, src in enumerate(srcs):
```
thesayyn, line 17:
```suggestion
    for src in srcs:
```

examples/transpiler/babel.bzl
```diff
@@ -1,32 +1,37 @@
 "Adapter from the Babel CLI to the ts_project#transpiler interface"
 
+load("@aspect_rules_js//js:defs.bzl", "js_library")
```
alexeagle, line 3: since this PR claims to fix an issue, I think you should add another `ts_project` call to this example that demonstrates the issue is fixed, like a `filegroup` in the `srcs` that goes through this same `babel` rule and assert on the output files (see the new testing.bzl#assert_outputs in bazel-lib)

docs/transpiler.md
```diff
@@ -20,10 +20,10 @@ Read more: https://blog.aspect.dev/typescript-speedup
...
 
 The `transpiler` attribute accepts a rule or macro with this signature:
-`name, srcs, js_outs, map_outs, **kwargs`
+`name, srcs, **kwargs`
 where the `**kwargs` attribute propagates the tags, visibility, and testonly attributes from `ts_project`.
 
-If you need to pass additional attributes to the transpiler rule, you can use a
+If you need to pass additional attributes to the transpiler rule such as `out_dir`, you can use a
 [partial](https://github.com/bazelbuild/bazel-skylib/blob/main/lib/partial.bzl)
 to bind those arguments at the "make site", then pass that partial to this attribute where it will be called with the remaining arguments.
```
alexeagle, line 29:
```suggestion
The transpiler rule or macro is responsible for predicting and declaring outputs.
If you want to pre-declare the outputs (so that they can be addressed by a Bazel label, like `//path/to:index.js`)
then you should use a macro which calculates the predicted outputs and supplies them to a `ctx.attr.outputs` attribute
on the rule.
```

alexeagle (approved):
Nice, I think you made a good case for this and the changes look right to me.

examples/transpiler/babel.bzl
jbedard: Note that currently this PR keeps this babel example as-is... one babel invocation per .ts file, and currently it `fail`s for non-ts srcs such as `filegroup`s.

We should have a proper example of what is being fixed and maybe this is the place to do it, or could demo both here...

alexeagle: yes that's true - we wanted to avoid a cyclic dependency rules_ts <-> rules_swc which is why it's like this. I think you could add that test in the mock_transpiler instead.

jbedard: I think a proper test needs a rule and not just a macro? So the rule can expand the target (such as a `filegroup`) into individual files to declare the outputs. Or output a directory...? 🤔

alexeagle: Yeah either a tiny rule or a directory makes sense

jbedard: It looks like a directory isn't possible because the `*_typings` rule outputs the `.d.ts` into that directory, so the transpiler can't declare the full directory as an output. That is (and was) a restriction that I'm not sure about... 🤔 

Any ideas for the easiest way to wrap babel in a rule?

alexeagle: Can't you use --outDir to put babel output in a subfolder?

jbedard: We could, but then it's in a different directory then the `.d.ts` outputted by the typechecking?

Note that this is putting it in a subfolder, but both the .js and .d.ts are put in the same subfolder atm.

alexeagle: the types and js files don't have to be in the same folder, package.json fields allow an npm package to ship with them that way. We just need something minimal here to exercise the new logic. If you want to make some other rule to exercise it in a way that makes more sense to users, I think that's okay

jbedard: See the changes to ts/test/mock_transpiler.bzl, similar to the change in rules_swc

examples/transpiler/babel.bzl
```diff
@@ -35,7 +43,16 @@ def babel(name, srcs, js_outs, map_outs, **kwargs):
...
+            outs = [js_out, map_out],
             args = args,
             **kwargs
         )
+
+        outs.append(js_out)
+        outs.append(map_out)
+
+    # The target producing the js files which ts_project() will reference
+    js_library(
```
jbedard, line 55: This is required because the main `js_library` within the `ts_project` macro now includes the compiled js/maps via rule name (this rule), where previously it included them all as file references.

Previously this `name` didn't have to be respected at all actually since everything was referenced as output files instead of rule targets.

ts/test/mock_transpiler.bzl
```diff
@@ -1,26 +1,68 @@
...
 _DUMMY_SOURCEMAP = """{"version":3,"sources":["%s"],"mappings":"AAAO,KAAK,CAAC","file":"in.js","sourcesContent":["fake"]}"""
 
-def mock(name, srcs, js_outs, map_outs, **kwargs):
-    """Mock transpiler macro.
+def _mock_impl(ctx):
+    src_files = [src for src in ctx.files.srcs if src.short_path.endswith(".ts") and not src.short_path.endswith(".d.ts")]
+    out_files = []
+
+    for src in src_files:
+        out_path = src.short_path[len(ctx.attr.pkg_dir) + 1:] if src.short_path.startswith(ctx.attr.pkg_dir) else src.short_path
```
jbedard, line 10: I don't understand why this is need (and the `pkg_dir` to do it). Without this it is nesting the output files too much, with the directory name repeated such as `test/test/ts/test/file.js`.

I don't see where in the default tsc this is avoided..? 🤔

ts/test/BUILD.bazel
```diff
@@ -52,6 +51,15 @@ ts_project(
     tsconfig = _TSCONFIG,
 )
 
+# Ensure the output files are predeclared
+build_test(
+    name = "out_refs_test",
+    targets = [
```
thesayyn, line 57: this is nice!

thesayyn (commented):
looks good but we may need more tests. It would be great to have a test with `filegroup` `select`, and `target with multiple files` in it.

jbedard:
@thesayyn I've added a `filegroup` test (see ts/test/BUILD.bazel). Adding a `select` would require https://github.com/aspect-build/rules_ts/issues/220 and can be done there if we decide to do it.

examples/transpiler/babel.bzl
```diff
@@ -35,7 +43,16 @@ def babel(name, srcs, js_outs, map_outs, **kwargs):
...
+            outs = [js_out, map_out],
             args = args,
             **kwargs
         )
+
+        outs.append(js_out)
+        outs.append(map_out)
+
+    # The target whose default outputs are the js files which ts_project() will reference
+    native.filegroup(
```
alexeagle, line 55: note, I changed this from `js_library` to `filegroup` to be more clear you don't actually need any providers returned

ts/test/mock_transpiler.bzl
alexeagle: maybe we should figure that out now then? I added a commit that uses the ts_lib logic.

alexeagle (approved):

examples/transpiler/babel.bzl
jbedard: sounds good 👍

ts/test/mock_transpiler.bzl
jbedard: I was looking at those but unsure what ts_lib was doing different 🤷 

I guess this is an internal test so there's no issue with using internal utils (ts_lib)?

jbedard:
Good to merge then?

ts/test/mock_transpiler.bzl
alexeagle: Yeah and ts_lib could be more public, others will want to do it this way I think

PR #244, opened 2022-11-16
build: turn off ts_project workers in e2e/bzlmod since they are currently broken on Windows and this is tripping the bazel central registry pre-submit test

gregmagolan:
The failure on Windows with workers looks non-trivial. At first glance, it looks as tho node is _not_ following symlinks into the virtual store and can't resolve transitive deps.

### DIFF
--- a/e2e/bzlmod/BUILD
+++ b/e2e/bzlmod/BUILD
@@ -4,6 +4,7 @@ load("@bazel_skylib//rules:build_test.bzl", "build_test")
 ts_project(
     name = "ts",
     srcs = ["foo.ts"],
+    supports_workers = False,
 )
 
 build_test(


e2e/bzlmod/BUILD
```diff
@@ -4,6 +4,7 @@ load("@bazel_skylib//rules:build_test.bzl", "build_test")
 ts_project(
     name = "ts",
     srcs = ["foo.ts"],
+    supports_workers = False,
```
alexeagle, line 7: Add a comment / link to issue?

alexeagle: What about using bazelrc so that only windows has a disable via the spawn strategy?

gregmagolan: Chatted about this at the OSS sync. We'll add a feature flag that the rule can check which will disable workers for all ts_project targets in a build. --spawn_stragety=local is too blunt as it applies to everything in the build

gregmagolan: Actually, simplest is to detect the host platform and disable workers entire on Windows for now. Will do that here.

gregmagolan:
Replaced by https://github.com/aspect-build/rules_ts/pull/245

PR #259, opened 2022-11-30
fix: report changes in deleted, created, changed order

thesayyn:
Reporting changes in random order causes tsc to call fstree functions when it's not in the final state and leads to hard failures. the safest way to do it is to report deletion first, creations second, and finally the changes.

example of a hard failing case;

```
/bazel-bin/node_modules/@types/node  -->  .store/@types+node+1.x/node_modules/@types/node
```
when version of `@types/node` changes from `1.x` to `2.x` the new symlink for `/bazel-bin/node_modules/@types/node` will be `.store/@types+node+2.x/node_modules/@types/node`.

given the example, 

if the worker reports the changes first, then tsc will try to determine the new path for `/bazel-bin/node_modules/@types/node` by making a realpath call. However, functions such as `fileExists`, `directoryExists` won't return the correct results since the fstree node addition of the new package (`.store/@types+node+2.x/node_modules/@types/node`) hasn't taken place yet. 

Therefore, correct way to report changes is; 

1- Report the deletion of `.store/@types+node+1.x/node_modules/@types/node`
2- Report the addition of `.store/@types+node+2.x/node_modules/@types/node`
3- Report the change of  `/bazel-bin/node_modules/@types/node`

### DIFF
--- a/e2e/worker/package.json
+++ b/e2e/worker/package.json
@@ -7,6 +7,6 @@
         "@nestjs/common": "^9.0.8",
         "rxjs": "^7.5.6",
         "reflect-metadata": "^0.1.13",
-        "@types/node": "^18.6.3"
+        "@types/node": "18.11.9"
     }
 }
--- a/ts/private/ts_project_worker.js
+++ b/ts/private/ts_project_worker.js
@@ -87,6 +87,9 @@ function createFilesystemTree(root, inputs) {
     function followSymlinkUsingRealFs(p) {
         const absolutePath = path.join(root, p)
         const stat = fs.lstatSync(absolutePath)
+        // bazel does not expose any information on whether an input is a REGULAR FILE,DIR or SYMLINK
+        // therefore a real filesystem call has to be made for each input to determine the symlinks.
+        // NOTE: making a readlink call is more expensive than making a lstat call
         if (stat.isSymbolicLink()) {
             const linkpath = fs.readlinkSync(absolutePath);
             const absoluteLinkPath = path.isAbsolute(linkpath) ? linkpath : path.resolve(path.dirname(absolutePath, linkpath))
--- a/ts/private/ts_project_worker.js
+++ b/ts/private/ts_project_worker.js
@@ -642,7 +642,7 @@ function emit(args, inputs) {
             } else if (previousInputs[input] != digest) {
                 changes.add(input);
             } else if (previousInputs[input] == null && digest == null) {
-                // Assume symlinks always change. bazel < 5.3 will always report symlinks without a digest.
+                // Assume symlinks always change. bazel <= 5.3 will always report symlinks without a digest.
                 // therefore there is no way to determine if a symlink has changed. 
                 changes.add(input);
             }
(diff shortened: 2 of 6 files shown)


e2e/worker/package.json
```diff
@@ -7,6 +7,6 @@
         "@nestjs/common": "^9.0.8",
         "rxjs": "^7.5.6",
         "reflect-metadata": "^0.1.13",
-        "@types/node": "^18.6.3"
+        "@types/node": "18.11.9"
```
jbedard, line 10: Are the version changes required for the tests?

ts/private/ts_project_worker.js
```diff
@@ -84,6 +84,17 @@ function createFilesystemTree(root, inputs) {
         return node;
     }
 
+    function followSymlinkUsingRealFs(p) {
```
jbedard, line 87: Can we not use `realpath`?

e2e/worker/package.json
thesayyn: yes. 18.6.3 yield errors when targeting tsc 4.9

ts/private/ts_project_worker.js
thesayyn: lstatSync is faster than realpath.

thesayyn: do you mean instead of readLinkSync? if not;

lstatSync is fast so it makes sense to use it for distinguishing symlinks. the number of symlinks is a tiny portion of the inputs so it doesn't make sense to do a realpath for every input.

jbedard: Sounds good 👍 

Worth a comment so I don't ask the same question in the future? 😅

thesayyn: Agreed.

ts/private/ts_project_worker.js
```diff
@@ -618,36 +632,46 @@ function emit(args, inputs) {
...
+    const changes = new Set(), creations = new Set();
+
     if (previousInputs) {
         timingStart(`invalidate`);
+        for (const [input, digest] of Object.entries(inputs)) {
+            if (!(input in previousInputs)) {
+                creations.add(input);
+            } else if (previousInputs[input] != digest) {
+                changes.add(input);
+            } else if (previousInputs[input] == null && digest == null) {
```
thesayyn, line 644: 🤡

ts/private/ts_project_worker.js
```diff
@@ -618,36 +632,46 @@ function emit(args, inputs) {
...
+                // Assume symlinks always change. bazel < 5.3 will always report symlinks without a digest.
```
thesayyn, line 645: 🤡

jbedard: I assume this is hard to write tests for? 😓

gregmagolan: bazel < 5.3 or Bazel <= 5.3?

thesayyn: > I assume this is hard to write tests for? 😓

i did write an e2e for it.

thesayyn: > bazel < 5.3 or bazel <= 5.3?

latter. i'll fix it

thesayyn: https://github.com/aspect-build/rules_ts/pull/259/files#diff-314a713aa24f705d09b849d18346ab921244c08ee6c6e7cda6f75c8d6894f59cR178-R199 this case will fail on bazel 5.3 without the fix.

PR #262, opened 2022-12-01
fix(transpile): don't throw away untranspiled srcs

alexeagle:
Note, this still needs some analysis testing, and an example that makes it load-bearing.

Also it's worth thinking about the correct behavior in the non-transpiler use case, though that's a bit more subtle.


ts/defs.bzl
```diff
@@ -315,14 +315,24 @@ def ts_project(
...
         test_target_name = "%s_typecheck_test" % name
 
-        if type(transpiler) == "function" or type(transpiler) == "rule":
-            transpiler(
+        # To emulate resolveJsonModule in typescript, the transpiler should copy non-TS sources
+        # to the output tree. We do this just by checking if the transpiler reports that some srcs
+        # were unused (it didn't produce any outputs for them) and send those to our terminal
+        # js_library.
+        if type(transpiler) == "rule":
+            not_transpiled_srcs = transpiler(
```
alexeagle, line 323: this one might not make sense, since a rule doesn't ever return anything

gregmagolan: Yea, that will always return None. You can only assume that all sources are consumed in this case. Perhaps mention that explicitly in the transpiler docstring above.

alexeagle:
filegroup srcs are a critical design flaw here. thinking about a simpler approach.

alexeagle:
@gregmagolan and I don't remember the context of this one. It looks like our goal was to permit `sass` source files to ts_project, but I don't see an issue filed for it. I guess if it's important, it will come up again and we can search for this WIP and try to make progress on the thorny design issue

PR #263, opened 2022-12-02
chore: cleanup transpiler_test_suite test

gregmagolan:
No functional changes; just moving all of the targets that the test_suite depends on into the .bzl so the code is much easier to follow.


jbedard (approved):
👍

kormide (approved):

Issue #268, opened 2022-12-05
[Bug]: Transitive npm deps not propagated through npm_package by default


mvukov:
### What happened?

Looks like ts_project -> js_binary -> docker image misses routing transitive npm deps. Reproduced a bug in https://github.com/oqton/rules_ts/tree/feature/rules_docker_bug.

The issue looks very similar to https://github.com/aspect-build/rules_jest/issues/72 to me.

### Version

Development (host) and target OS/architectures: Ubuntu 20.04

Output of `bazel --version`: 5.2.0

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file:

* rules_ts: 1.0.2

Language(s) and/or frameworks involved:

### How to reproduce

```shell
$ bazel run //app:image
$ docker run -it --rm bazel/app:image
node:internal/modules/cjs/loader:936
  throw err;
  ^

Error: Cannot find module 'ts-md5'
Require stack:
- /app/app/app.sh.runfiles/rules_jest_ts_bug/node_modules/.aspect_rules_js/@oqton+lib1@0.0.0/node_modules/@oqton/lib1/index.js
- /app/app/app.sh.runfiles/rules_jest_ts_bug/app/main.js
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:933:15)
    at Function.Module._load (node:internal/modules/cjs/loader:778:27)
    at Module.require (node:internal/modules/cjs/loader:1005:19)
    at require (node:internal/modules/cjs/helpers:102:18)
    at Object.<anonymous> (/app/app/app.sh.runfiles/rules_jest_ts_bug/node_modules/.aspect_rules_js/@oqton+lib1@0.0.0/node_modules/@oqton/lib1/index.js:4:18)
    at Module._compile (node:internal/modules/cjs/loader:1105:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1159:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Module.require (node:internal/modules/cjs/loader:1005:19) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/app/app/app.sh.runfiles/rules_jest_ts_bug/node_modules/.aspect_rules_js/@oqton+lib1@0.0.0/node_modules/@oqton/lib1/index.js',
[...]

PR #280, opened 2022-12-20
Fix transitive propagation of npm deps

mvukov:
Fixes:
* https://github.com/aspect-build/rules_ts/issues/268
* https://github.com/aspect-build/rules_jest/issues/72

### DIFF
--- a/ts/private/ts_project.bzl
+++ b/ts/private/ts_project.bzl
@@ -247,7 +247,7 @@ This is an error because Bazel does not run actions unless their outputs are nee
     )
 
     npm_package_store_deps = js_lib_helpers.gather_npm_package_store_deps(
-        targets = ctx.attr.data,
+        targets = ctx.attr.data + ctx.attr.deps,
     )
 
     output_declarations_depset = depset(output_declarations)


mvukov:
@alexeagle @thesayyn PTAL.

gregmagolan:
Adding this my list of OSS bug squashing for later this week

alexeagle:
We just need to be careful we don't re-introduce eager type-checking or cascading rebuilds.

gregmagolan:
I'll keep it in mind. We did add analysis test cases to cover that now.

mvukov:
Anything I can help with?

gregmagolan:
Thanks for the ping. I didn't have time to get to this before the Christmas break. Will look soon. This needs a bit of time to understand the side-effects of this change. In the meantime, it's a small patch to carry locally if you need this change.

flolu:
@gregmagolan awesome, thanks! How would I apply this patch to `ts_project.bzl` and use it for my project in the meantime?

gregmagolan:
It's a very common practice to patch Bazel rules with `http_archive` in your WORKSPACE. Even rule sets sometimes patch other rule sets. For example, https://github.com/bazelbuild/rules_go/blob/master/go/private/repositories.bzl#L78

mvukov:
@gregmagolan Friendly ping :) PTAL when you find some time.

gregmagolan:
Turns out that @jbedard and I have been discussing the underlying issue that the PR address already. I go into the details of the problem in this comment https://github.com/aspect-build/rules_ts/issues/268#issuecomment-1401598145. The principled fix for this starts with a PR in rules_js to introduce a `dev_npm_deps` attribute. Closing this PR as the fix will take a different shape once the change in rules_js lands. Will leave https://github.com/aspect-build/rules_ts/issues/268 open.

PR #264, opened 2022-12-02
refactor: improve worker correctness and add observability

thesayyn:
NIT: these changes allow `--worker_sandboxing` to be supported easily since we no longer include execroot part of the path in cache keys. 

this major refactor allows; 

- js sources without having to specify an outDir
- d.ts sources without having to specify an outDir
- slightly less memory usage there's fewer file watcher nodes
- fewer fs calls for paths that are outside of execroot since execroot is now the cwd.

paves the way for these features
- worker_sandboxing out of the box

closes #293 #272 #196 #189

PR #294, opened 2023-01-22
feat: implement custom flags for verbosity and workers

thesayyn:
I am still trying to make some analysis tests work...


ts/BUILD.bazel
```diff
@@ -1,4 +1,53 @@
...
+    flag_values = {
+        ":supports_workers": "true"
+    }
+)
+
+configuration(
+    name = "configuration",
+    visibility = ["//visibility:public"],
+    verbose = select({
+        ":legacy_verbose": True,
```
alexeagle, line 41: nice!

.bazelrc
```diff
@@ -1,6 +1,13 @@
 # import common bazelrc settings used in this repository and in nested workspaces
 import %workspace%/.bazelrc.common
 
+# ts flag aliases
+build --flag_alias=supports_workers=@aspect_rules_ts//ts:supports_workers
```
alexeagle, line 5: I'd alias this as "ts_workers"

ts/defs.bzl
```diff
@@ -222,19 +222,18 @@ def ts_project(
...
         supports_workers: Whether the worker protocol is enabled.
             To disable worker mode for a particular target set `supports_workers` to `False`.
-            Worker mode can be controlled as well via `--strategy` and `mnemonic` and  using .bazelrc.
 
-            Put this in your .bazelrc to disable it globally: `build --strategy=TsProject=sandboxed`
+            Worker mode can be disabled workspace wide by using the `--@aspect_rules_ts//ts:supports_workers` flag. 
+            To disable worker mode globally, insert `build --@aspect_rules_ts//ts:supports_workers=false` into the .bazelrc.
+
+            Alternatively, worker mode can be controlled via `--strategy`.
+            To disable worker mode globally via `--strategy` insert `build --strategy=TsProject=sandboxed` into the .bazelrc.
```
alexeagle, line 230: isn't this the choice that doesn't work correctly though?

ts/private/configuration.bzl
```diff
@@ -0,0 +1,43 @@
+"""A terminal rule collecting verbosity and worker support information"""
+
+ConfigurationInfo = provider(
```
alexeagle, line 3: I know this mirrors "config_setting", but naming this thing "configuration" feels too close to "ts_config" to me. Will be confusing in communication. Can we name it something else? "Options" maybe?

ts/private/ts_lib.bzl
```diff
@@ -53,9 +53,10 @@ STD_ATTRS = {
         allow_files = True,
         mandatory = True,
     ),
-    "supports_workers": attr.bool(
+    # TODO: make it None-able bool once https://github.com/bazelbuild/bazel/issues/14434 is fixed.
+    "supports_workers": attr.int(
```
alexeagle, line 57: is this already a  breaking change? What happens for users who are currently passing True/False here?

alexeagle (commented):
Generally looks great, do you have any specific questions I could answer?

ts/defs.bzl
thesayyn: when #264 lands, using `--strategy=TsProject=sandboxed` will be identical to having `supports_workers = False`

ts/private/ts_lib.bzl
thesayyn: yeah. this is a breaking change for those using ts_project_rule instead of the wrapper.  If that's not too bad, I could make this non-breaking by wrapping the `ts_project_rule` in a macro.

ts/defs.bzl
thesayyn: even without #264, `--strategy=TsProject=sandboxed` works correctly with minor changes in log output.

thesayyn:
> Generally looks great, do you have any specific questions I could answer?

I just wanted to get your opinion on this. i am planning on landing this after #264

alexeagle:
You mentioned this one at standup, but it doesn't look ready for review. Could you make it a bit more clear for me please? Do the conflict resolving rebase and mark it ready for review rather than draft?

thesayyn:
> You mentioned this one at standup, but it doesn't look ready for review. Could you make it a bit more clear for me please? Do the conflict resolving rebase and mark it ready for review rather than draft?

it's not ready for final review yet. I just wanted to get your opinion on the new attribute to prevent breaking changes. if that's good enough then I'll commit to the PR more.

ts/private/configuration.bzl
thesayyn: renamed it to options

thesayyn:
@alexeagle ready for review

.bazelrc
```diff
@@ -1,6 +1,14 @@
 # import common bazelrc settings used in this repository and in nested workspaces
 import %workspace%/.bazelrc.common
 
+# ts flag aliases
+build --flag_alias=ts_workers=@aspect_rules_ts//ts:supports_workers
```
alexeagle, line 5: this doesn't seem like a good idea to me unless we are sure that users will define these flag_alias's as well. With this we might read our code and not realize it won't work the same for a user.

ts/defs.bzl
```diff
@@ -267,19 +236,18 @@ def ts_project(
...
```
alexeagle, line 244: wait, why would we document two different ways to do the same thing? Isn't one of them better?

ts/defs.bzl
```diff
@@ -267,19 +236,18 @@ def ts_project(
 
         supports_workers: Whether the worker protocol is enabled.
             To disable worker mode for a particular target set `supports_workers` to `False`.
-            Worker mode can be controlled as well via `--strategy` and `mnemonic` and  using .bazelrc.
 
-            Put this in your .bazelrc to disable it globally: `build --strategy=TsProject=sandboxed`
+            Worker mode can be disabled workspace wide by using the `--@aspect_rules_ts//ts:supports_workers` flag.
```
alexeagle, line 240: I think it needs a bit more clarification how the global setting interacts with the local one.

ts/BUILD.bazel
```diff
@@ -31,13 +75,58 @@ config_setting(
...
+    name = "verbose_flag",
+    flag_values = {
+        ":verbose": "true"
+    }
+)
+
+config_setting(
+    name = "supports_workers_flag",
+    flag_values = {
+        ":supports_workers": "true"
```
alexeagle, line 94: TODO: 2.0 change default to false? (assuming we are able to make swc default transpiler)

ts/private/ts_project.bzl
```diff
@@ -84,6 +93,9 @@ See https://github.com/aspect-build/rules_ts/issues/228 for more details.
         execution_requirements["worker-key-mnemonic"] = "TsProject"
         executable = ctx.executable.tsc_worker
 
+    # Add all arguments from options first to allow users override them via args.
+    arguments.add_all(options.args)
```
alexeagle, line 97: note, this means you won't be able to see the `args=["--skipLibCheck"]` when you `bazel query //the:target` now, like you could previously. But since we already "magically add" to the args here I think it's fine to change from macro to rule doing this.

alexeagle (commented):
neat, nice approach :)

ts/defs.bzl
thesayyn: I can't really say that each one has its downsides and upsides and are inherently different even though you get the same result when executing locally. however, you may still want to disable workers when running locally and let the dynamic execution strategy override it when requested. https://bazel.build/remote/dynamic

ts/defs.bzl
thesayyn: agreed. will fix it

ts/BUILD.bazel
thesayyn: i'll put up a comment above

.bazelrc
thesayyn: these are not meant for the users. however, I agree with you that this could be confusing for us especially. perhaps i should remove these?

thesayyn: removed them.

PR #299, opened 2023-02-01
fix: transitive dependency type checking with transpiler

jbedard:
I'm seeing this at sourcegraph using a transpiler just like this. It seems to be when there is an indirect npm dependency? Or that's my guess at least, I haven't fully investigated that.

The test is the exact same as `deps_lib` but using babel as the transpiler and fails without the fix.

### DIFF
--- a/ts/defs.bzl
+++ b/ts/defs.bzl
@@ -378,6 +378,7 @@ def ts_project(
             # that this js_library can be a valid dep for downstream ts_project or other rules_js derivative rules.
             srcs = [transpile_target_name, tsc_target_name],
             deps = deps,
+            data = data,
             **common_kwargs
         )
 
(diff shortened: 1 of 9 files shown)


ts/defs.bzl
```diff
@@ -378,6 +378,7 @@ def ts_project(
             # that this js_library can be a valid dep for downstream ts_project or other rules_js derivative rules.
             srcs = [transpile_target_name, tsc_target_name],
             deps = deps,
+            data = data,
```
jbedard, line 381: Should `assets` also go into the `data`?

gregmagolan: assets should go in `srcs` actually

gregmagolan: can you put that up in a separate PR?

gregmagolan (approved):

PR #302, opened 2023-02-02
release: publish release artifact

gregmagolan:
GitHub's stability guarantee for the archive is iffy, and we want metrics on downloads. See bazel-contrib/SIG-rules-authors#11 (comment)

### DIFF
--- a/.github/workflows/release_prep.sh
+++ b/.github/workflows/release_prep.sh
@@ -7,7 +7,10 @@ set -o errexit -o nounset -o pipefail
 TAG=${GITHUB_REF_NAME}
 # Strip leading 'v'
 PREFIX="rules_ts-${TAG:1}"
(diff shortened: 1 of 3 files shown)


.github/workflows/release_prep.sh
```diff
@@ -40,7 +43,7 @@ http_archive(
     name = "aspect_rules_ts",
     sha256 = "${SHA}",
     strip_prefix = "${PREFIX}",
-    url = "https://github.com/aspect-build/rules_ts/archive/refs/tags/${TAG}.tar.gz",
+    url = "https://github.com/aspect-build/rules_js/releases/download/${TAG}/${ARCHIVE}",
```
kormide, line 46: typo?

gregmagolan: intentional. we're switching to release artifacts now for all of our rule sets

gregmagolan: ohh. you mean the rules_js bit.. yes, thanks for catching that.

jbedard (approved):

Issue #284, opened 2023-01-03
[Bug]: ts_project does not work in an external workspace


pcj:
### What happened?

For a `ts_project` rule defined in an external workspace such as:

```
# /private/var/tmp/_bazel_pcj/d7e82b68b84ece06dc82c3639a1eabcf/external/protobufapis/google/protobuf/BUILD.bazel:522:17
ts_project_rule(
  name = "descriptor_ts_project",
  visibility = ["//visibility:public"],
  generator_name = "descriptor_ts_project",
  generator_function = "proto_ts_project",
  generator_location = "google/protobuf/BUILD.bazel:522:17",
  allow_js = False,
  composite = False,
  declaration = True,
  declaration_map = True,
  emit_declaration_only = False,
  incremental = False,
  preserve_jsx = False,
  source_map = True,
  args = [],
  data = ["@protobufapis//google/protobuf:descriptor_proto"],
  deps = [...],
  srcs = ["@protobufapis//google/protobuf:descriptor.ts"],
  supports_workers = True,
  transpile = True,
  tsc = "@npm_typescript//:tsc",
  tsc_worker = "@npm_typescript//:tsc_worker",
  tsconfig = "//:tsconfig_proto",
  js_outs = ["@protobufapis//google/protobuf:descriptor.js"],
  map_outs = ["@protobufapis//google/protobuf:descriptor.js.map"],
  typing_maps_outs = ["@protobufapis//google/protobuf:descriptor.d.ts.map"],
  typings_outs = ["@protobufapis//google/protobuf:descriptor.d.ts"],
)
# Rule descriptor_ts_project instantiated at (most recent call last):
#   /private/var/tmp/_bazel_pcj/d7e82b68b84ece06dc82c3639a1eabcf/external/protobufapis/google/protobuf/BUILD.bazel:522:17            in <toplevel>
#   /private/var/tmp/_bazel_pcj/d7e82b68b84ece06dc82c3639a1eabcf/external/build_stack_rules_proto/rules/ts/proto_ts_project.bzl:7:22 in proto_ts_project
#   /Users/pcj/go/src/github.com/Omnistac/unity/bazel_tools/node/ts_project.bzl:31:16                                                in ts_project
#   /private/var/tmp/_bazel_pcj/d7e82b68b84ece06dc82c3639a1eabcf/external/aspect_rules_ts/ts/defs.bzl:360:20                         in ts_project
# Rule ts_project_rule defined at (most recent call last):
[...]

PR #304, opened 2023-02-02
fix: use full path for tsconfig.json, fix outdir for ts_project

alexander-fenster:
Hi folks,

We were using TypeScript `nodejs_binary` from the custom rule in another Bazel project. After switching from `rules_nodejs` to `aspect-build/rules_ts`, I found that this use case is now broken.

Specifically, two issues:

1. Where the original `ts_project` rule used `ctx.file.tsconfig.path` [here](https://github.com/bazelbuild/rules_nodejs/blob/4b8bcf2f7c07d2f44a2fa7c6ce1f7431d277f7ed/nodejs/private/ts_project.bzl#L55), the current rule uses `ctx.file.tsconfig.short_path`:

https://github.com/aspect-build/rules_ts/blob/3fee9c2c8cc41d52b29259ee0c3bac16336c8dc8/ts/private/ts_project.bzl#L91-L98

This breaks if the `js_binary` rule that corresponds to that `ts_project` rule is used as an executable from another Bazel project (linebreaks added for readability):

```
ERROR: /path/to/my/homedir/.cache/bazel/_bazel_username/0875673185f3ca7fa2dace5f957ce567/external/typescript_project/BUILD.bazel:9:11:
Compiling TypeScript project @typescript_project//:compile_typescript
[tsc -p ../typescript_project/tsconfig.json] [for tool] failed: (Exit 1):
tsc.sh failed: error executing command (from target @typescript_project//:compile_typescript)
bazel-out/k8-opt-exec-2B5CBBC6/bin/external/npm_typescript/tsc.sh \
  --module esnext --moduleResolution node \
  --project ../typescript_project/tsconfig.json \
  --outDir . --rootDir external/typescript_project

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
error TS5058: The specified path does not exist: '../typescript_project/tsconfig.json'.
Target //:do_stuff failed to build
```

(note how it's unable to find `tsconfig.json` where `--project` points to)

2. The `outdir` seems to be broken for this use case, `tsc` does not generate files where Bazel expects them to be.

I made the full repro: https://github.com/alexander-fenster/aspect-ts-repro, it shows the problem and how the suggested fix (from this PR) makes it work.

Thank you!

### DIFF
--- a/ts/private/ts_project.bzl
+++ b/ts/private/ts_project.bzl
@@ -85,12 +85,17 @@ See https://github.com/aspect-build/rules_ts/issues/228 for more details.
     # Add user specified arguments *before* rule supplied arguments
     arguments.add_all(ctx.attr.args)
 
-    outdir = _lib.join(ctx.label.package, ctx.attr.out_dir) if ctx.attr.out_dir else ctx.label.package
+    if ctx.attr.out_dir:
+        outdir = _lib.join(ctx.label.package, ctx.attr.out_dir)
+    elif ctx.label.package:
+        outdir = ctx.label.package
+    else:
+        outdir = ctx.label.workspace_root
     if outdir == "":
         outdir = "."
     arguments.add_all([
         "--project",
-        ctx.file.tsconfig.short_path,
+        ctx.file.tsconfig.path,
         "--outDir",
         outdir,
         "--rootDir",


alexander-fenster:
It's interesting that the CI fails with

```
TypeError: Cannot read properties of undefined (reading 'fileNames')
    at setConfigFileParsingResult 
```

Is it something related to my changes, or something that just fails in `main`? I spent some time figuring out this problem during my migration, and the best I could do was disable validation of `tsconfig`.

alexander-fenster:
Closing this in favor of https://github.com/aspect-build/rules_ts/issues/284, I will check if the fix works for our use case. Thank you!

PR #310, opened 2023-02-09
fix: ts_project external workspace builds

### LINKED ISSUE #284: [Bug]: ts_project does not work in an external workspace

dgp1130:
Fixes #284.

`short_path` is not sufficient to handle external paths and we need to strip `ctx.bin_dir.path` instead. This updates the `ts_project()` rules as well as adds a regression test which triggers a `ts_project()` target to build from the context of an external workspace.

### DIFF
--- a/e2e/external_dep/WORKSPACE
+++ b/e2e/external_dep/WORKSPACE
@@ -1,0 +1,29 @@
+workspace(name = "lib_wksp")
+
+local_repository(
+    name = "aspect_rules_ts",
+    path = "../..",
+)
+
+load("@aspect_rules_ts//ts:repositories.bzl", "LATEST_VERSION", "rules_ts_dependencies")
+rules_ts_dependencies(ts_version = LATEST_VERSION)
+
+# Fetch and register node, if you haven't already
+load("@rules_nodejs//nodejs:repositories.bzl", "DEFAULT_NODE_VERSION", "nodejs_register_toolchains")
+
+nodejs_register_toolchains(
+    name = "node",
+    node_version = DEFAULT_NODE_VERSION,
+)
+
+load("@aspect_rules_js//npm:npm_import.bzl", "npm_translate_lock")
+
+npm_translate_lock(
+    name = "npm",
+    pnpm_lock = "//:pnpm-lock.yaml",
+    verify_node_modules_ignored = "//:.bazelignore",
+)
+
+load("@npm//:repositories.bzl", "npm_repositories")
+
+npm_repositories()
--- a/ts/private/ts_project.bzl
+++ b/ts/private/ts_project.bzl
@@ -1,6 +1,7 @@
 "ts_project rule"
 
-load("@aspect_bazel_lib//lib:copy_to_bin.bzl", "copy_files_to_bin_actions")
+load("@aspect_bazel_lib//lib:copy_to_bin.bzl", "copy_file_to_bin_action", "copy_files_to_bin_actions")
+load("@aspect_bazel_lib//lib:paths.bzl", "to_output_relative_path")
 load("@aspect_bazel_lib//lib:platform_utils.bzl", "platform_utils")
 load("@bazel_skylib//lib:dicts.bzl", "dicts")
 load("@aspect_rules_js//js:providers.bzl", "JsInfo", "js_info")
@@ -45,6 +46,7 @@ def _ts_project_impl(ctx):
         list of providers
     """
     srcs_inputs = copy_files_to_bin_actions(ctx, ctx.files.srcs)
+    tsconfig = copy_file_to_bin_action(ctx, ctx.file.tsconfig)
 
     srcs = [_lib.relative_to_package(src.path, ctx) for src in srcs_inputs]
 
@@ -85,21 +87,21 @@ See https://github.com/aspect-build/rules_ts/issues/228 for more details.
     # Add user specified arguments *before* rule supplied arguments
     arguments.add_all(ctx.attr.args)
 
-    outdir = _lib.join(ctx.label.package, ctx.attr.out_dir) if ctx.attr.out_dir else ctx.label.package
-    if outdir == "":
-        outdir = "."
+    outdir = _lib.join(
+        ctx.label.workspace_root,
+        _lib.join(ctx.label.package, ctx.attr.out_dir) if ctx.attr.out_dir else ctx.label.package,
+    )
+    tsconfig_path = to_output_relative_path(tsconfig)
     arguments.add_all([
         "--project",
-        ctx.file.tsconfig.short_path,
+        tsconfig_path,
         "--outDir",
         outdir,
         "--rootDir",
         _lib.calculate_root_dir(ctx),
     ])
     if len(typings_outs) > 0:
-        declaration_dir = _lib.join(ctx.label.package, typings_out_dir)
-        if declaration_dir == "":
-            declaration_dir = "."
+        declaration_dir = _lib.join(ctx.label.workspace_root, ctx.label.package, typings_out_dir)
         arguments.add_all([
             "--declarationDir",
             declaration_dir,
@@ -230,7 +232,7 @@ This is an error because Bazel does not run actions unless their outputs are nee
             execution_requirements = execution_requirements,
             progress_message = "Compiling TypeScript project %s [tsc -p %s]" % (
                 ctx.label,
-                ctx.file.tsconfig.short_path,
+                tsconfig_path,
             ),
             env = {
                 "BAZEL_BINDIR": ctx.bin_dir.path,
--- a/ts/private/ts_validate_options.bzl
+++ b/ts/private/ts_validate_options.bzl
@@ -2,7 +2,8 @@
 
 load(":ts_config.bzl", "TsConfigInfo")
 load(":ts_lib.bzl", "COMPILER_OPTION_ATTRS", "ValidOptionsInfo")
-load("@aspect_bazel_lib//lib:copy_to_bin.bzl", "copy_files_to_bin_actions")
+load("@aspect_bazel_lib//lib:copy_to_bin.bzl", "copy_file_to_bin_action", "copy_files_to_bin_actions")
+load("@aspect_bazel_lib//lib:paths.bzl", "to_output_relative_path")
 load("@aspect_rules_js//js:providers.bzl", "JsInfo")
 
 def _tsconfig_inputs(ctx):
@@ -23,6 +24,7 @@ def _validate_options_impl(ctx):
     # Bazel won't run our action unless its output is needed, so make a marker file
     # We make it a .d.ts file so we can plumb it to the deps of the ts_project compile.
     marker = ctx.actions.declare_file("%s.optionsvalid.d.ts" % ctx.label.name)
+    tsconfig = copy_file_to_bin_action(ctx, ctx.file.tsconfig)
 
     # Provider validation
     if not ctx.attr.allow_js:
@@ -51,7 +53,12 @@ To disable this check, set the validate attribute to False:
         incremental = ctx.attr.incremental,
         ts_build_info_file = ctx.attr.ts_build_info_file,
     )
-    arguments.add_all([ctx.file.tsconfig.short_path, marker.short_path, ctx.attr.target, json.encode(config)])
+    arguments.add_all([
+        to_output_relative_path(tsconfig),
+        to_output_relative_path(marker),
+        ctx.attr.target,
+        json.encode(config),
+    ])
 
     inputs = _tsconfig_inputs(ctx)
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -64,6 +64,8 @@ jobs:
                 folder:
                     - '.'
                     - 'e2e/bzlmod'
+                    - 'e2e/external_dep'
+                    - 'e2e/external_dep/app'
                     - 'e2e/worker'
                     - 'e2e/workspace'
                 exclude:
--- a/e2e/external_dep/.bazelignore
+++ b/e2e/external_dep/.bazelignore
@@ -1,0 +1,2 @@
+app/
+node_modules/
--- a/e2e/external_dep/.bazelrc
+++ b/e2e/external_dep/.bazelrc
@@ -1,0 +1,2 @@
+# import common bazelrc shared with e2e workspaces
+import %workspace%/../../.bazelrc.common
--- a/e2e/external_dep/.bazelversion
+++ b/e2e/external_dep/.bazelversion
@@ -1,0 +1,1 @@
+../../.bazelversion
--- a/e2e/external_dep/BUILD
+++ b/e2e/external_dep/BUILD
@@ -1,0 +1,16 @@
+"""Tests a `ts_project()` in one workspace being used in another."""
+
+load("@aspect_rules_ts//ts:defs.bzl", "ts_project")
+load("@bazel_skylib//rules:build_test.bzl", "build_test")
+
+ts_project(
+    name = "lib",
+    srcs = ["lib.ts"],
+    declaration = True,
+    visibility = ["//visibility:public"],
+)
+
+build_test(
+    name = "test",
+    targets = [":lib"],
+)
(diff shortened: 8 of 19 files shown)


e2e/external_dep/WORKSPACE
```diff
@@ -0,0 +1,44 @@
+workspace(name = "lib_wksp")
+
+# Override http_archive for local testing
+local_repository(
+    name = "aspect_rules_ts",
+    path = "../..",
+)
+
+#---SNIP--- Below here is re-used in the workspace snippet published on releases
```
jbedard, line 9: This is only true for the `e2e/workspace`. See the WORKSPACE files in the other examples and most of the comments here can be deleted.

ts/private/ts_project.bzl
```diff
@@ -86,18 +86,23 @@ See https://github.com/aspect-build/rules_ts/issues/228 for more details.
     arguments.add_all(ctx.attr.args)
 
     outdir = _lib.join(ctx.label.package, ctx.attr.out_dir) if ctx.attr.out_dir else ctx.label.package
+    if ctx.label.workspace_root:
```
jbedard, line 89: I think we can drop the `if` and just do this 100% of the time?

jbedard: That would also cover the `outdir == "":` case, so we can just have a single `outdir = _lib.join(..)`?

ts/private/ts_project.bzl
```diff
@@ -86,18 +86,23 @@ See https://github.com/aspect-build/rules_ts/issues/228 for more details.
...
+        outdir = _lib.join(ctx.label.workspace_root, outdir)
     if outdir == "":
         outdir = "."
+    tsconfig_path = ctx.file.tsconfig.path.removeprefix("%s/" % ctx.bin_dir.path)
```
jbedard, line 93: Can `lib.relative_to_package` be used for this?

e2e/external_dep/WORKSPACE
dgp1130: Thanks for pointing that out, updated these two `WORKSPACE` files to copy from the `worker` `e2e` test.

ts/private/ts_project.bzl
dgp1130: I don't think so because that strips the `external/${workspaceName}/` prefix which seems to be required and I think that's desired in this case? I'm not totally sure how external targets are actually laid out in the temporary directory, but using `.removeprefix()` the path is `extenral/lib_wksp/tsconfig.json` while using `lib.relative_to_package()` just uses `tsconfig.json`. I think we want the former so we need to stick with `.removeprefix()`?

dgp1130:
Question: Where are the `e2e` workspaces tested in CI? I don't see anything which obvious runs `(cd e2e/*/ && bazel test //...)`. My new test case actually has a nested external workspaces and I realized I forgot to `.bazelignore` it in the outer workspace (which also doesn't include a test). I would have expected CI to report a failure but it didn't. Do I need to update something to register this example to run on CI?

jbedard:
See `.github/workflows/ci.yaml`, currently they're listed manually in there (if I'm reading that right...)

ts/private/ts_project.bzl
```diff
@@ -85,19 +85,23 @@ See https://github.com/aspect-build/rules_ts/issues/228 for more details.
...
-        ctx.file.tsconfig.short_path,
+        tsconfig_path,
         "--outDir",
         outdir,
         "--rootDir",
         _lib.calculate_root_dir(ctx),
     ])
     if len(typings_outs) > 0:
         declaration_dir = _lib.join(ctx.label.package, typings_out_dir)
+        if ctx.label.workspace_root:
```
jbedard, line 103: Can we remove the `if`s similar to `outdir`?

dgp1130: Done.

dgp1130:
Ah thanks, I did look in `ci.yaml` and saw the `bazel test //...` but missed the working directory matrix. Added the new e2e test there as well.

e2e/external_dep/WORKSPACE
```diff
@@ -0,0 +1,28 @@
...
+    name = "node",
+    node_version = DEFAULT_NODE_VERSION,
+)
+
+load("@aspect_rules_js//npm:npm_import.bzl", "npm_translate_lock")
+
+npm_translate_lock(
+    name = "npm",
+    pnpm_lock = "//:pnpm-lock.yaml",
+)
```
gregmagolan, line 24: should add
```
verify_node_modules_ignored = "//:.bazelignore",
```
here and the corresponding `node_modules` entry in `.bazelignore`

e2e/external_dep/workspace/WORKSPACE
```diff
@@ -0,0 +1,30 @@
...
+nodejs_register_toolchains(
+    name = "node",
+    node_version = DEFAULT_NODE_VERSION,
+)
+
+load("@aspect_rules_js//npm:npm_import.bzl", "npm_translate_lock")
+
+npm_translate_lock(
+    name = "npm",
+    pnpm_lock = "//:pnpm-lock.yaml",
```
gregmagolan, line 25: should add
```
verify_node_modules_ignored = "//:.bazelignore",
```
here and the corresponding `node_modules` entry in `.bazelignore`

ts/private/ts_project.bzl
```diff
@@ -85,21 +85,21 @@ See https://github.com/aspect-build/rules_ts/issues/228 for more details.
...
-        outdir = "."
+    outdir = _lib.join(
+        ctx.label.workspace_root,
+        _lib.join(ctx.label.package, ctx.attr.out_dir) if ctx.attr.out_dir else ctx.label.package,
+    )
+    tsconfig_path = ctx.file.tsconfig.path.removeprefix("%s/" % ctx.bin_dir.path)
     arguments.add_all([
         "--project",
-        ctx.file.tsconfig.short_path,
+        tsconfig_path,
```
gregmagolan, line 95: I think this can be simplified to,

```
to_output_relative_path(ctx.file.tsconfig)
```

and it will be correct even in the unexpected case where the file is not an output file.

you can load that from

```
load("@aspect_bazel_lib//lib:paths.bzl", "to_output_relative_path")
```

ts/private/ts_project.bzl
```diff
@@ -85,21 +85,21 @@ See https://github.com/aspect-build/rules_ts/issues/228 for more details.
...
         "--outDir",
         outdir,
         "--rootDir",
         _lib.calculate_root_dir(ctx),
     ])
     if len(typings_outs) > 0:
-        declaration_dir = _lib.join(ctx.label.package, typings_out_dir)
-        if declaration_dir == "":
-            declaration_dir = "."
+        declaration_dir = _lib.join(ctx.label.workspace_root, ctx.label.package, typings_out_dir)
```
gregmagolan, line 102: I think we still need
```
        if declaration_dir == "":
            declaration_dir = "."
```
since `ctx.label.workspace_root` will be "" for the `__main__` workspace

gregmagolan: We should also add a test case for this one. The case is a `ts_project` in the root BUILD file of a repository with `declaration = True` (may as well make it in an `e2e` so we're not polluting the root package of `rules_ts).

ts/private/ts_validate_options.bzl
```diff
@@ -51,7 +51,12 @@ To disable this check, set the validate attribute to False:
         incremental = ctx.attr.incremental,
         ts_build_info_file = ctx.attr.ts_build_info_file,
     )
-    arguments.add_all([ctx.file.tsconfig.short_path, marker.short_path, ctx.attr.target, json.encode(config)])
+    arguments.add_all([
+        ctx.file.tsconfig.path.removeprefix("%s/" % ctx.bin_dir.path),
```
gregmagolan, line 55: Can use `to_output_relative_path` for these two as well

gregmagolan (changes requested):
Left some suggestions

e2e/external_dep/WORKSPACE
dgp1130: Done.

e2e/external_dep/workspace/WORKSPACE
dgp1130: Done.

ts/private/ts_project.bzl
dgp1130: Done. This got slightly more complicated as it turns out `ctx.file.tsconfig.path` is misleading because it referenced the raw source file _before_ it was copied to bin, so `to_output_relative_path()` was using a bad input. Updated the rule to copy to bin and the very start with the rest of the sources so everything would use the copied file.

ts/private/ts_project.bzl
dgp1130: Actually this case passes correctly as is. `_lib.join()` defaults to "." if the result is empty, so I don't think we need to change anything.

https://github.com/aspect-build/rules_ts/blob/ce5df5155c96cb84826176462a310b81a62f8f10/ts/private/ts_lib.bzl#L142-L146

I updated the existing example to use `declaration = True` and added a `build_test()` directly to verify that it builds in both workspaces. I think that should at least serve as a regression test going forward.

ts/private/ts_validate_options.bzl
dgp1130: Done. Same issue here where `tsconfig.json` was being copied to bin after using it's path. Hoisted the copy to bin earlier in the rule. `marker` didn't have this problem since it is created internally in the rule, so it's always in the bin directory.

dgp1130:
Thanks for the suggestions @gregmagolan, updated with the fixes!

ts/private/ts_project.bzl
gregmagolan: Oh nice. That works 👍

gregmagolan (approved):
🦖

e2e/external_dep/WORKSPACE
```diff
@@ -0,0 +1,29 @@
+workspace(name = "lib_wksp")
```
gregmagolan, line 1: Oh shoot. git is going to get confused between the file `WORKSPACE` and the folder `workspace`.

One of them needs to be renamed to something else.

gregmagolan (changes requested):
Noticed one last thing

e2e/external_dep/WORKSPACE
dgp1130: Grr, case-insensitive file names strike again! Update the `workspace/` directory to `app/`, which is probably a more accurate name for its intended abstraction.

gregmagolan (approved):
🌮

PR #313, opened 2023-02-11
feat: propagate non-dev npm dependencies in ts_project deps to direct dependencies of downstream linked npm_package targets

gregmagolan:
Follow-up to https://github.com/aspect-build/rules_js/pull/860

Fixes https://github.com/aspect-build/rules_ts/issues/268

This also fixes the bug where "dev" dependencies in the data attribute are incorrectly propagated to the direct dependencies of downstream linked npm_package targets.


ts/defs.bzl
```diff
@@ -136,7 +136,21 @@ def ts_project(
         data: Files needed at runtime by binaries or tests that transitively depend on this target.
             See https://bazel.build/reference/be/common-definitions#typical-attributes
 
-        deps: List of labels of other rules that produce TypeScript typings (.d.ts files)
+        deps: List of labels of other rules that produce TypeScript typings (`.d.ts` files)
+
+            If this list contains linked npm packages, npm package store targets or other targets that provide
+            `JsInfo`, `NpmPackageStoreInfo` providers are gathered from `JsInfo`. This is done directly from
+            the `npm_package_store_deps` field of these. For linked npm package targets, the underlying
```
jbedard, line 143: I don't think people will know what `NpmPackageStoreInfo` or `npm_package_store_deps` is?

ts/defs.bzl
```diff
@@ -136,7 +136,21 @@ def ts_project(
...
+            `npm_package_store` target(s) that back the links is used. Gathered `NpmPackageStoreInfo`
+            providers are propagated to the direct dependencies of downstream linked `npm_package` targets.
```
jbedard, line 145: I think "direct dependencies" is confusing here, since we're talking about indirect deps aren't we? (even if it's implemented by propagating to the direct deps).

Can we say something such as "providers are available to downstream linked `npm_package` targets"?

ts/private/ts_lib.bzl
```diff
@@ -32,10 +32,11 @@ STD_ATTRS = {
     "declaration_dir": attr.string(
         doc = "https://www.typescriptlang.org/tsconfig#declarationDir",
     ),
-    # Note, this is overridden by the @bazel/typescript implementation in build_bazel_rules_nodejs
-    # to add an aspect on the deps attribute.
     "deps": attr.label_list(
-        doc = "Other targets which produce TypeScript typings",
+        doc = """List of labels of other rules that produce TypeScript typings (`.d.ts` files)
```
jbedard, line 36: Do we have any standard for the use of the "target" and "label" terms? A label being a name of a target... can a target be referenced by something other then its label such as referencing a pre-declared `.d.ts` output of a target?

jbedard:
We should really add tests here as well. Something such as `@types/*` and then ensuring something such as `require.resolve('@types/x')` fails?

ts/defs.bzl
gregmagolan: Hmm I think "direct dependencies of downstream linked packages" is correct. Basically saying that they go in the `dependencies` of the downstream linked package if you were to translate to `package.json` terms

ts/private/ts_lib.bzl
gregmagolan: This change was because the two docstrings for deps were out of sync between the macro and rule... but "labels of other rules" is a mouthful. Just `targets` is better.

gregmagolan: In this case they have to be targets that provide JsInfo so a reference to a `.d.ts` file won't work; that would have to go in `srcs`

ts/defs.bzl
gregmagolan: I think that's ok for now. The SoT for this docstring snippet is in rules_js. We can improve on it there first.

gregmagolan:
> We should really add tests here as well. Something such as `@types/*` and then ensuring something such as `require.resolve('@types/x')` fails?

Yes. New e2e of the same shape you added to rules_js perhaps?

ts/defs.bzl
jbedard: Yeah I think it's correct, just confusing with so many keywords in there. I had to read it multiple times :grimacing:

jbedard (approved):

gregmagolan:
Test case added: `examples/linked` and `examples/linked_consumer`

Issue #271, opened 2022-12-09
[Bug]: ts_config doesn't provide a JsInfo


gregmagolan:
See https://bazelbuild.slack.com/archives/CEZUUKQ6P/p1670621678898069?thread_ts=1670620951.407769&cid=CEZUUKQ6P for more context

gregmagolan:
Transitive npm deps from the `tsconfig` attribute in `ts_project` should _not_ get forwarded to downstream consumers via the JsInfo it provides or via runfiles.

Toxaris:
I put together a repro in https://github.com/Toxaris/ts_config_deps_repro.

Toxaris:
We figured out in Slack that disabling workers is a work-around. I updated the https://github.com/Toxaris/ts_config_deps_repro to also demonstrate the work around.

gregmagolan:
Gotcha. That is quite possible as there seem to be a few more gremlins in worker mode.
cc @thesayyn

gregmagolan:
@Toxaris The latest release of rules_ts has a big refactor for worker mode. When you have a chance could you please try turning working mode back on in your repro?

gregmagolan:
Fix will be included in v1.13.0 which I'm going to cut right after https://github.com/aspect-build/rules_ts/pull/316 lands

PR #315, opened 2023-02-11
feat: add JsInfo provider to ts_config rule

gregmagolan:
Fixes https://github.com/aspect-build/rules_ts/issues/271 although it looks like there is additional speed bump that `tsc` doesn't follow the symlinked to the virtual store location of its `extends` so transitive npm deps of an `extends` can't be resolved. I suspect this is easily reproducible outside of Bazel when `hoist = False` in `.npmrc`.

See comment left in `examples/linked_tsconfig_consumer/tsconfig.json` for more info: 

```
{
    "//": "the hacky path below is required because while @lib/tsconfig/tsconfig.json _will_ resolve, tsc won't follow the symlink into the virtual store so the transitive npm dep of @lib/tsconfig on @tsconfig/node18/tsconfig.json won't be resolvable",
    "extends": "../node_modules/.aspect_rules_js/@lib+tsconfig@0.0.0/node_modules/@lib/tsconfig/tsconfig.json"
}
```

There are other way this could be worked around including manually propagating the dependency to downstream ts_project targets but the hacky virtual store path work-around above seems the simplest; here we just abandon the standard node_modules resolution algorithm that would incorrectly resolve to `node_modules/@lib/tsconfig/tsconfig.json` from where transitive npm deps of `@lib/tsconfig` would not be resolvable unless they were elsewhere on the node_modules resolution path such as `node_modules/my_transitive_dep`. The downside of this work-around is that would not work outside of Bazel with the same tsconfig file since under pnpm the virtual store path would be different.

@Toxaris There is a follow-up task here of reproducing the issue with tsc outside of Bazel and filing an issue to the typescript repo but I've got too many other things on my plate atm so I'm going to punt on that. If you wanted to make the repro and file an issue upstream I bet the fix is a relatively straight forward one in typescript.

### DIFF
--- a/ts/private/ts_project.bzl
+++ b/ts/private/ts_project.bzl
@@ -218,7 +218,7 @@ This is an error because Bazel does not run actions unless their outputs are nee
     if len(outputs) > 0:
         inputs_depset = depset(
             copy_files_to_bin_actions(ctx, inputs),
-            transitive = transitive_inputs + [_gather_declarations_from_js_providers(ctx.attr.srcs + ctx.attr.deps)],
+            transitive = transitive_inputs + [_gather_declarations_from_js_providers(ctx.attr.srcs + [ctx.attr.tsconfig] + ctx.attr.deps)],
         )
 
         ctx.actions.run(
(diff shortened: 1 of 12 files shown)


kormide (approved):

ts/private/ts_project.bzl
```diff
@@ -218,7 +218,7 @@ This is an error because Bazel does not run actions unless their outputs are nee
     if len(outputs) > 0:
         inputs_depset = depset(
             copy_files_to_bin_actions(ctx, inputs),
-            transitive = transitive_inputs + [_gather_declarations_from_js_providers(ctx.attr.srcs + ctx.attr.deps)],
+            transitive = transitive_inputs + [_gather_declarations_from_js_providers(ctx.attr.srcs + [ctx.attr.tsconfig] + ctx.attr.deps)],
```
jbedard, line 221: Isn't it possible there will be no `tsconfig` so `[None]` will be passed to `_gather_declarations_from_js_providers`? Is that ok?

gregmagolan: I checked the same thing. Not possible as the attribute is `mandatory = True` in the rule. You either get the user tsconfig or an auto-generated one that is done in the `ts_project` macro

jbedard: 👍

PR #321, opened 2023-02-14
feat: global AST cache

thesayyn:


gregmagolan:
2-3% performance improvement in simple cases

alexeagle (commented):
I guess we can close this, since you're not likely coming back to more performance fixes here?

thesayyn:
giving up on this one as we don't plan to work on workers more.

PR #322, opened 2023-02-14
refactor: consider json a js source (if resolve_json_module)

gzm0:
This allows us to use the _out_paths machinery for json sources as well.

To achieve this, we add a resolve_json_module parameter to relevant methods in ts_lib (ts_lib is private so this is a backwards compatible change).

There are two small side-effects to this refactor:
- We use _lib.relative_to_package rather than a prefix removal.
- json inputs that are not sources and do not have an input / output collision are now declared.

### DIFF
--- a/ts/defs.bzl
+++ b/ts/defs.bzl
@@ -345,7 +345,7 @@ def ts_project(
     tsc_js_outs = []
     tsc_map_outs = []
     if not transpiler:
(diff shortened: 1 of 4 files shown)


gregmagolan:
I've forgotten all of the context on json inputs to ts_projects.... but if we're pushing to 2.0 @alexeagle we should look at this again and understand the change in this PR

alexeagle (approved):

PR #323, opened 2023-02-14
fix: source maps generated incorrectly

thesayyn:
waiting for confirmation from John

### DIFF
--- a/ts/private/ts_project_worker.js
+++ b/ts/private/ts_project_worker.js
@@ -776,8 +776,18 @@ function createProgram(args, inputs, output, exit) {
         if (p.endsWith(".map")) {
             // We need to postprocess map files to fix paths for the sources. This is required because we have a SYNTHETIC_OUTDIR suffix and 
             // tsc tries to relativitize sources back to rootDir. in order to fix it the leading `../` needed to be stripped out.
-            // TODO: figure out how to make tsc do this for us.
-            data = data.replace(`"../`, `"`)
+            // We tried a few options to make tsc do this for us.
+            // 
+            // 1- Using mapRoot to reroot map files. This didn't work because either path in `sourceMappingUrl` or path in `sources` was incorrect.
+            // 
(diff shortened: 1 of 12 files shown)


ts/private/ts_project_worker.js
```diff
@@ -771,12 +767,18 @@ function createProgram(args, inputs, output, exit) {
...
         ts.sys.createDirectory(p);
     }
 
     function writeFile(p, data, mark) {
-        p = p.replace(SYNTHETIC_OUTDIR, ".")
+        p = p.replace(SYNTHETIC_OUTDIR, "")
+        if (p.endsWith(".map")) {
+            // We need to postprocess map files to fix paths for the sources. This is required because we have a SYNTHETIC_OUTDIR suffix and 
+            // tsc tries to relativitize sources back to rootDir. in order to fix it the leading `../` needed to be stripped out.
+            // TODO: figure out how to make tsc do this for us.
```
gregmagolan, line 779: Can you expand on this comment a bit? I think the TODO should drop since we know its not possible with tsc in the current impl. Better to replace it with a more detailed explanation of the problem and what we tried that didn't work (namely mapRoot & the `../path/file` approach which broke in the root package.

ts/private/ts_project_worker.js
```diff
@@ -771,12 +767,18 @@ function createProgram(args, inputs, output, exit) {
...
+            data = data.replace(`"../`, `"`)
```
gregmagolan, line 780: Should probably contain this to the `sources` list in the `.map` file incase the `sourceRoot` happens to start with `../`?

thesayyn: right... I'll make it more specific and add some test cases for those as well.

ts/private/ts_project_worker.js
thesayyn: Sure.

gregmagolan (approved):

thesayyn:
@gregmagolan that was a fast re-review. Thanks!!!!!

PR #342, opened 2023-03-24
chore: update dependencies

gregmagolan:


docs/rules.md
```diff
@@ -62,7 +62,7 @@ Implementation rule behind the ts_project macro.
...
...oviders are gathered from <code>JsInfo</code>. This is done directly from the <code>npm_package_store_deps</code> field of these. For linked npm package targets, the underlying <code>npm_package_store</code> target(s) that back the links is used. Gathered <code>NpmPackageStoreInfo</code> providers are propagated to the direct dependencies of downstream linked <code>npm_package</code> targets.<br><br>NB: Linked npm package targets that are "dev" dependencies do not forward their underlying <code>npm_package_store</code> target(s) through <code>npm_package_store_deps</code> and will therefore not be propagated to the direct dependencies of downstream linked <code>npm_package</code> targets. npm packages that come in from <code>npm_translate_lock</code> are considered "dev" dependencies if they are have <code>dev: true</code> set in the pnpm lock file. This should be all packages that are only listed as "devDependencies" in all <code>package.json</code> files within the pnpm workspace. This behavior is intentional to mimic how <code>devDependencies</code> work in published npm packages.   | <a href="https://bazel.build/concepts/labels">List of labels</a> | optional | <code>[]</code> |
```
kormide, line 65: What triggrered this change?

gregmagolan: The docstring for data is shared from rules_js... although I plan remove that pattern in the near future since its much cleaner to just duplicate the strings. Exporting docstrings feels too coupled.

alexeagle: I think importing docstrings is a nice way to get current content in the place users would most like to read it, I don't see a problem with this side-effect of a dep upgrade

alexeagle (approved):

docs/rules.md
gregmagolan: The side-effect itself is not that bad,,... the problem I saw was when adding a dependency on a some newly exported docstring from rules_js that then ends up being a breaking change for users that upgrade the rule set but not rules_js in the same commit. It makes it a poor coupling in the long run.

PR #346, opened 2023-03-31
chore: Update tool versions to include TS 5.0.2 and 5.0.3

ericrafalovsky:


thesayyn:
tests will most likely fail due to recent changes to the typescript packaging.

ericrafalovsky:
> tests will most likely fail due to recent changes to the typescript packaging.

Thanks for pointing this out! I'm about to head on holiday so will close out this PR for now since more substantial changes will be required :)

PR #345, opened 2023-03-31
feat: introduce skip_lib_check attribute

thesayyn:
closes #326
```
Benchmark 1: bazel build examples/skip_lib_check/unspecified
  Time (abs ≡):         1.734 s               [User: 0.051 s, System: 0.060 s]
 
Benchmark 2: bazel build examples/skip_lib_check/on
  Time (abs ≡):        737.8 ms               [User: 59.2 ms, System: 50.6 ms]
 
Benchmark 3: bazel build examples/skip_lib_check/off
  Time (abs ≡):         1.180 s               [User: 0.049 s, System: 0.045 s]
 
Summary
  'bazel build examples/skip_lib_check/on' ran
    1.60 times faster than 'bazel build examples/skip_lib_check/off'
    2.35 times faster than 'bazel build examples/skip_lib_check/unspecified'
```

PR #347, opened 2023-04-01
Revert "feat: introduce skip_lib_check attribute"

alexeagle:
Reverts aspect-build/rules_ts#345

Let's roll-forward with the fix, is that okay Sahin? I've got some local WIP using a config_setting instead I'd like to design with you.


thesayyn (approved):
sounds good

PR #350, opened 2023-04-01
feat: skipLibCheck take 2

### LINKED ISSUE #348: FR: roll-forward a skipLibCheck performance gain for most users
### What happened?

#345 was reverted, it needs a bit more design. I'll work through it w/ Sahin.

- Default value for `skipLibCheck` is false per the TS team at Microsoft and it's confusing when our defaults don't match theirs.
- There are real cases where `skipLibCheck` is incorrect and masks type-errors. If you ship a library, that can mean your users hit the bug.
- We want users to be aware of this optimization and our recommendation that they elect to opt-in to it.
- There should be a place in their codebase that "skipLibCheck" appears, in tsconfig or a config setting or something, so they can figure out they chose it

### Version

HEAD

### How to reproduce

_No response_

### Any other information?

_No response_

alexeagle:
Here's the DX of what the `select` failure message looks like, will be live for 2.0 under this proposal:

```
$ bazel build examples/simple:ts --strategy=TsProject=sandboxed -s
ERROR: /home/alexeagle/Projects/rules_ts/examples/simple/BUILD.bazel:9:11: configurable attribute "args" in //examples/simple:ts doesn't match this configuration: 

######## Required Typecheck Performance Selection ########

TypeScript's type-checking exposes a flag `--skipLibCheck`:
https://www.typescriptlang.org/tsconfig#skipLibCheck

Using this flag saves substantial time during type-checking.
Rather than doing a full check of all d.ts files, TypeScript will only type-check the code you
specifically refer to in your app's source code.
We recommend this for most rules_ts users.

HOWEVER this performance improvement comes at the expense of type-system accuracy. 
For example, two packages could define two copies of the same type in an inconsistent way.
If you publish a library from your repository, your incorrect types may result in errors for your users.

You must choose exactly one of the following flags:

1. To choose the faster performance, put this in /.bazelrc:

    # passes an argument `--skipLibCheck` to *every* spawn of tsc
    build --@aspect_rules_ts//ts:skipLibCheck=always

2. To choose more correct typechecks, put this in /.bazelrc:

    # honor the setting of `skipLibCheck` in the tsconfig.json file
    build --@aspect_rules_ts//ts:skipLibCheck=honor_tsconfig

##########################################################

This instance of //examples/simple:ts has configuration identifier 1cd096f. To inspect its configuration, run: bazel config 1cd096f.

For more help, see https://bazel.build/docs/configurable-attributes#faq-select-choose-condition.

ERROR: Analysis of target '//examples/simple:ts' failed; build aborted: 
INFO: Elapsed time: 0.095s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (1 packages loaded, 0 targets configured)
```

Fixes #348

### DIFF
--- a/ts/BUILD.bazel
+++ b/ts/BUILD.bazel
@@ -1,42 +1,34 @@
 load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
-load("@bazel_skylib//rules:common_settings.bzl", "bool_flag")
+load("@bazel_skylib//rules:common_settings.bzl", "string_flag")
 
 exports_files(
     glob(["*.bzl"]),
     visibility = ["//docs:__pkg__"],
 )
 
-# Users can enable with --@aspect_rules_ts//ts:skipLibCheck
-bool_flag(
+# Users can enable with --@aspect_rules_ts//ts:skipLibCheck=always
+string_flag(
     # Note: this name is chosen to be searchable when users look for "skipLibCheck" in their repo,
     # as that's the way it appears in tsconfig.json or tsc command line.
     name = "skipLibCheck",
-    build_setting_default = False,
-)
-
-# Users can enable with --@aspect_rules_ts//ts:use_skipLibCheck_from_tsconfig
-bool_flag(
-    name = "use_skipLibCheck_from_tsconfig",
-    build_setting_default = False,
+    # TODO(2.0): change default to "unspecified"
+    build_setting_default = "honor_tsconfig",
+    values = [
+        "always",
+        "honor_tsconfig",
+        "unspecified",
+    ],
 )
 
 # Note, users could use a Transition to make a subgraph of their depgraph opt-in to skipLibCheck.
 config_setting(
     name = "skip_lib_check.always",
-    # Note: check both values to ensure they are mutually exclusive, not both set.
-    flag_values = {
-        ":skipLibCheck": "1",
-        ":use_skipLibCheck_from_tsconfig": "0",
-    },
+    flag_values = {":skipLibCheck": "always"},
 )
 
 config_setting(
     name = "skip_lib_check.honor_tsconfig",
-    # Again, check that they are mutually exclusive.
-    flag_values = {
-        ":skipLibCheck": "0",
-        ":use_skipLibCheck_from_tsconfig": "1",
-    },
+    flag_values = {":skipLibCheck": "honor_tsconfig"},
 )
 
 bzl_library(
--- a/ts/BUILD.bazel
+++ b/ts/BUILD.bazel
@@ -1,8 +1,31 @@
 load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
+load("@bazel_skylib//rules:common_settings.bzl", "int_flag")
 
 exports_files(
     glob(["*.bzl"]),
     visibility = ["//docs:__pkg__"],
+)
+
+# Enable with --@aspect_rules_ts//ts:skip_lib_check=1
+int_flag(
+    name = "skip_lib_check",
+    # Default is 0 to be a non-breaking change in rules_ts 1.x
+    # TODO(2.0): change default to -1 so we require users to consider which value of the flag is best.
+    build_setting_default = 0,
+)
+
+config_setting(
+    name = "skip_lib_check_honor_tsconfig",
+    flag_values = {
+        ":skip_lib_check": "0",
+    },
+)
+
+config_setting(
+    name = "skip_lib_check_always",
+    flag_values = {
+        ":skip_lib_check": "1",
+    },
 )
 
 bzl_library(
--- a/ts/BUILD.bazel
+++ b/ts/BUILD.bazel
@@ -1,30 +1,41 @@
 load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
-load("@bazel_skylib//rules:common_settings.bzl", "int_flag")
+load("@bazel_skylib//rules:common_settings.bzl", "bool_flag")
 
 exports_files(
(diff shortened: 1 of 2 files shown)


ts/BUILD.bazel
```diff
@@ -1,10 +1,33 @@
 load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
+load("@bazel_skylib//rules:common_settings.bzl", "int_flag")
 
 exports_files(
     glob(["*.bzl"]),
     visibility = ["//docs:__pkg__"],
 )
 
+# Enable with --@aspect_rules_ts//ts:skip_lib_check=1
+int_flag(
```
thesayyn, line 10: I'd use a string flag here. `true` `false` `unspecified` or something close to this. self-describing values are better.

ts/BUILD.bazel
```diff
@@ -1,10 +1,33 @@
...
+    name = "skip_lib_check",
+    # Default is 0 to be a non-breaking change in rules_ts 1.x
+    # TODO(2.0): change default to -1 so we require users to consider which value of the flag is best.
+    build_setting_default = 0,
```
thesayyn, line 14: `-1` should be the default. 1/0 True/False  -1/honor ts_config

ts/BUILD.bazel
thesayyn: and limit it via `values` attribute to prevent user errors. https://github.com/bazelbuild/bazel-skylib/blob/141432789c92e9db2402ef0be58e2a2d2c4dd1fd/rules/common_settings.bzl#L92

alexeagle: yeah, OTOH I did that for stamping once in rules_docker and it was a mistake  because -1,0,1 had already won in many repos, and it seemed better to be consistent with trinary flags.

alexeagle: Maybe two Boolean flags is even better. Not sure if we need error handling in case you set both as they are meant to be mutually exclusive

ts/BUILD.bazel
alexeagle: PTAL with a pair of bool_flags instead, even less confusing than any int scheme

ts/defs.bzl
```diff
@@ -400,7 +431,16 @@ def ts_project(
     ts_project_rule(
         name = tsc_target_name,
         srcs = srcs,
-        args = args,
+        args = args + select(
+            {
+                "//ts:skip_lib_check.always": ["--skipLibCheck"],
```
thesayyn, line 436: what happens when users set both?

alexeagle: They get the same error as if they select neither.
We could easily improve that though, by having a config_setting that's just for error reporting - I'll add now

alexeagle: err no, I'm wrong, if one of the config_settings matches in the select then I don't know how to print an error.

They'll just have to figure out from how the error says "exactly one must be set"

thesayyn: hmm, maybe it's better to have a string flag/enum? that may allow users to use the flag repeatedly as well. (in places such as .bazelrc.user)

I am thinking `always/honor_tsconfig/never`

alexeagle: I don't have a way to implement `never` - we either inject an extra command-line arg or we don't.
Does that change it for you?

thesayyn: can't `--skipLibCheck` take true/false as a flag value? and when you do it overrides whatever is in tsconfig?

alexeagle: right... I filed that issue HAHA https://github.com/microsoft/TypeScript/issues/10967

However, I still don't think `never` is useful, we never wished for a way to turn *on* libCheck globally overriding what's in tsconfig. And we still don't wish for that. Would it be useful somehow?

thesayyn: it might be useful for some. we can leave it out in this PR and leave the possibility of adding it open with enum string flags.

alexeagle: hm in my testing with `string_flag(values=["x", "y"])` I'm not getting any error if I give some other value... something seems broken there

thesayyn: that's weird. maybe an old version of skylib? https://github.com/bazelbuild/bazel-skylib/blob/141432789c92e9db2402ef0be58e2a2d2c4dd1fd/rules/common_settings.bzl#L81-L98

alexeagle: something on my swc branch. Tested here and it works correctly if illegal value is provided.

string_flag LGTM - PTAL

ts/BUILD.bazel
alexeagle: now changed to `string_flag`

ts/defs.bzl
```diff
@@ -400,7 +431,13 @@ def ts_project(
...
```
thesayyn, line 436:
```suggestion
                "@rules_ts//ts:skip_lib_check.always": ["--skipLibCheck"],
```

ts/defs.bzl
```diff
@@ -433,7 +433,7 @@ def ts_project(
         srcs = srcs,
         args = args + select(
             {
-                "//ts:skip_lib_check.always": ["--skipLibCheck"],
+                "@rules_ts//ts:skip_lib_check.always": ["--skipLibCheck"],
                 "//ts:skip_lib_check.honor_tsconfig": [],
```
thesayyn, line 437: this line needs it too.

thesayyn (approved):
awesome

ts/defs.bzl
alexeagle: yeah that explains the CI failure I was ignoring :)

PR #354, opened 2023-04-19
refactor: rename LATEST_VERSION to LATEST_TYPESCRIPT_VERSION

jbedard:
Fix https://github.com/aspect-build/rules_js/issues/817

WDYT of changing this in 2.0+? Could add it as an alias for now...

### DIFF
--- a/e2e/external_dep/WORKSPACE
+++ b/e2e/external_dep/WORKSPACE
@@ -5,8 +5,9 @@ local_repository(
     path = "../..",
 )
 
-load("@aspect_rules_ts//ts:repositories.bzl", "LATEST_VERSION", "rules_ts_dependencies")
-rules_ts_dependencies(ts_version = LATEST_VERSION)
+load("@aspect_rules_ts//ts:repositories.bzl", "LATEST_TYPESCRIPT_VERSION", "rules_ts_dependencies")
+
+rules_ts_dependencies(ts_version = LATEST_TYPESCRIPT_VERSION)
 
 # Fetch and register node, if you haven't already
 load("@rules_nodejs//nodejs:repositories.bzl", "DEFAULT_NODE_VERSION", "nodejs_register_toolchains")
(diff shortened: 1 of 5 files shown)


gregmagolan:
We could export both and then its not breaking

jbedard:
Only if we plan to remove it in 2.0. Sound good?

gregmagolan:
We have `DEFAULT_NODE_VERSION` in rules_nodejs and I saw this in one of our bazel-examples

```
load("@aspect_rules_esbuild//esbuild:repositories.bzl", "esbuild_register_toolchains", LATEST_ESBUILD_VERSION = "LATEST_VERSION")

esbuild_register_toolchains(
    name = "esbuild",
    esbuild_version = LATEST_ESBUILD_VERSION,
)
```

so perhaps be explicit and follow the same convention with `LATEST_TYPESCRIPT_VERSION` as the exported symbol with `LATEST_VERSION` around for backward compat and to be removed in 2.0?

gregmagolan:
Should probably make the same change in rules_swc, rules_esbuild & rules_cypress

jbedard:
Ok I've switched to the `LATEST_*_VERSION` style and marked the `LATEST_VERSION` to be removed in 2.0. If that looks right I'll do the same in other rulesets as well.

e2e/external_dep/WORKSPACE
```diff
@@ -5,8 +5,9 @@ local_repository(
     path = "../..",
 )
 
-load("@aspect_rules_ts//ts:repositories.bzl", "LATEST_VERSION", "rules_ts_dependencies")
-rules_ts_dependencies(ts_version = LATEST_VERSION)
+load("@aspect_rules_ts//ts:repositories.bzl", "LATEST_TS_VERSION", "rules_ts_dependencies")
+
+rules_ts_dependencies(ts_version = LATEST_TS_VERSION)
```
gregmagolan, line 10: hmmm. how about LATEST_TYPESCRIPT_VERSION to match the npm package name of the tool?

jbedard: Yeah I like that idea, can do the same with rollup etc.

Updated.

gregmagolan (approved):
🌮

PR #362, opened 2023-04-27
Add ts 5.0 and disable workers for it

alexeagle:
We aren't ready to spend the time making workers function correctly with TS 5 so just disable it silently.

---

### Type of change

New feature or functionality (change which adds functionality)

Anyone using our `LATEST_VERSION` symbol will be on TS 5.0 now

### Test plan


ts/private/ts_lib.bzl
```diff
@@ -59,7 +59,11 @@ See more details on the `assets` parameter of the `ts_project` macro.
         doc = "Whether the tsc compiler understands Bazel's persistent worker protocol",
         default = False,
     ),
-    # TODO(2.0): remove this and make supports_workers a tri-state int. 
+    "is_typescript_5_or_greater": attr.bool(
```
jbedard, line 62: Should we prefix this with `internal_` or `_` to indicate it's internal?

jbedard (approved):

Issue #361, opened 2023-04-27
Support TS 5.0 with workers


alexeagle:
### What happened?

It has a bug with the persistent workers compiler wrapper:

```
#   ERROR: /tmp/bats-run-Sc0WkS/file/18/BUILD.bazel:5:11: Compiling TypeScript project @//:foo [tsc -p tsconfig.json] failed: (Exit 1): tsc_worker.sh failed: error executing command (from target //:foo) bazel-out/k8-opt-exec-2B5CBBC6/bin/external/npm_typescript/tsc_worker.sh @bazel-out/k8-fastbuild/bin/source.js-0.params
#
#   node_modules/.aspect_rules_js/@feature+cool@0.0.0/node_modules/@feature/cool/index.d.ts(1,40): error TS2307: Cannot find module '@feature/notcool' or its corresponding type declarations.
```

### Version

HEAD, TS >= 5.0

### How to reproduce

```shell
Turn on workers
```

alexeagle:
We will likely NEVER do this, because we'd rather stop carrying a custom wrapper around `tsc` which is subject to being broken by new TypeScript versions. Instead, we aim to address speed by using a different program like SWC for transpilation, and the feature described in https://github.com/microsoft/TypeScript/issues/47947 for type-checks.

devversion:
This is an interesting blocker because on Windows, sandbox doesn't exist and execution in the real `bazel-bin` can be quite problematic if a tsconfig is used from e.g. the workspace root.

tsc would then pull and see files from other packages/directories and complain about them not being part of the `rootDir`. Workers solved this issue via their virtual file system—so in practice `rules_ts` is no longer Windows compatible I'd say

devversion:
Even with swc or other alternatives suggested here, you ultimately want type checking and for that, you need tsc; which breaks on Windows

PR #363, opened 2023-04-28
chore: update e2e to use TS 5

alexeagle:
See #361


alexeagle:
We're not trying to fix worker mode with TS 5 anymore, as rules_ts 2.0 disables workers by default.

PR #365, opened 2023-05-04
fix: use correct file extensions for commonjs

MaartenStaa:
Some files have a `.js` extension, and are written as CommonJS. This causes issues when the `package.json` includes `"type": "module"` and is present in the sandbox during execution, because Node will treat `.js` files as ESM in that case. This is currently failing our Windows builds (but not Linux or macOS, strangely enough).


ts/defs.bzl
```diff
@@ -252,10 +252,10 @@ def ts_project(
     """
 
     if srcs == None:
-        include = ["**/*.ts", "**/*.tsx"]
+        include = ["**/*.ts", "**/*.mts", "**/*.cts", "**/*.tsx"]
```
jbedard, line 255: I don't think this is required for the fix? This seems more like a feature that would need testing...

jbedard:
How about putting a package.json with `"type": "commonjs"` in the rules_ts root instead? Then it won't be inheriting anything...

Can you add a test that fails? Maybe just a simple e2e test with a package.json?

alexeagle:
Hey @MaartenStaa it looks like this went stale. Also our `main` branch now has the 2.x branch history on it, so this would need a rebase. Do you mind opening a new PR that's more minimal and just makes the changes required to fix your issue? Also, if you could file an issue for the problem this is solving, that would help others discover it, and maybe someone else can make the PR.

PR #371, opened 2023-05-16
feat: ask users about skipLibCheck when upgrading to 2.0

alexeagle:
Previously this was an undiscoverable feature.

BREAKING CHANGE:
Users will be required to choose a setting. The error message tells them what to do.

---

### Type of change

### Test plan

### DIFF
--- a/ts/BUILD.bazel
+++ b/ts/BUILD.bazel
@@ -1,5 +1,5 @@
 load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
-load("@bazel_skylib//rules:common_settings.bzl", "string_flag", "bool_flag")
+load("@bazel_skylib//rules:common_settings.bzl", "bool_flag", "string_flag")
 load("//ts/private:options.bzl", "options")
 
 exports_files(
@@ -43,8 +43,7 @@ string_flag(
     # Note: this name is chosen to be searchable when users look for "skipLibCheck" in their repo,
     # as that's the way it appears in tsconfig.json or tsc command line.
     name = "skipLibCheck",
-    # TODO(2.0): change default to "unspecified"
-    build_setting_default = "honor_tsconfig",
+    build_setting_default = "unspecified",
     values = [
         "always",
         "honor_tsconfig",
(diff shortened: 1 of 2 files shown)


ts/BUILD.bazel
```diff
@@ -84,49 +84,47 @@ config_setting(
...
     flag_values = {
-        ":supports_workers": "true"
-    }
+        ":supports_workers": "true",
+    },
 )
 
 options(
     name = "options",
-    visibility = ["//visibility:public"],
```
alexeagle, line 100: note, this is just buidifier formatting, apparently it was previously not buildifier-clean

ts/BUILD.bazel
```diff
@@ -84,49 +84,47 @@ config_setting(
...
         {
             "@aspect_rules_ts//ts:skip_lib_check.always": True,
             "@aspect_rules_ts//ts:skip_lib_check.honor_tsconfig": False,
         },
         no_match_error = _skip_lib_check_selection_required,
-    )
+    ),
+    supports_workers = select({
+        ":supports_workers_flag": True,
+        # TODO(2.0): default should be False if we set swc as default
```
thesayyn, line 109: nit: remove TODO

thesayyn (approved):

ts/BUILD.bazel
jbedard, line 100: Can we avoid all the formatting changes? For a one line change this file has quite the diff, especially in a non-main branch where this needs to be merged :/

jbedard (approved):

ts/BUILD.bazel
alexeagle: The way to avoid it would be to force the previous committer on this file to have formatted it. We don't use Aspect Workflows and haven't bothered with a buildifier test on our "simple" OSS repos.

ts/BUILD.bazel
alexeagle: I think this is unrelated

ts/BUILD.bazel
jbedard: But we haven't done that :(

It just makes the diff ugly for such a tiny actual change, and when in a feature branch increases the likelihood of merge conflicts. I'd prefer one big format commit in `main` but 🤷

alexeagle: Sure we can just format main. Maybe it's time to add buildifier assertion in rules-template and everything we cloned from it

PR #372, opened 2023-05-16
chore: drop support for Bazel 5

alexeagle:
We'll introduce features like validation actions which are only in 6.

BREAKING CHANGE:
Users will likely be forced to upgrade to Bazel 6 in order to use rules_ts 2.0

---

### Type of change

### Test plan


jbedard (approved):
😃

thesayyn (approved):

PR #375, opened 2023-05-29
examples: rules_proto_grpc with @bufbuild

satajo:
Add an example of how to use @bufbuild based proto and grpc compilation with rules_js.

Last week I found myself trying to put together a @bufbuild based proto build setup, but kept running into rules that were either simply missing or used outdated libraries. Thanks to the discussion in https://github.com/aspect-build/rules_js/issues/397 I eventually go it to work, and figured that somebody might find a proper example useful.

The example setup is rather large, pulling in rules_proto_grpc, rules_js, rules_jest, etc, and so I figured it best to implement it as a separate workspace.

---

### Type of change

- Documentation (updates to documentation or READMEs)

### Test plan

- Manual testing; please provide instructions so we can reproduce: Run the test target inside the example workspace.

### DIFF
--- a/examples/rules_proto_grpc_bufbuild/.bazelrc
+++ b/examples/rules_proto_grpc_bufbuild/.bazelrc
@@ -1,0 +1,2 @@
+# Use bzlmod
+common --enable_bzlmod
--- a/examples/rules_proto_grpc_bufbuild/BUILD.bazel
+++ b/examples/rules_proto_grpc_bufbuild/BUILD.bazel
@@ -1,0 +1,98 @@
+load("@aspect_rules_jest//jest:defs.bzl", "jest_test")
+load("@aspect_rules_ts//ts:defs.bzl", "ts_project")
+load("@npm//:defs.bzl", "npm_link_all_packages")
+load("@rules_proto//proto:defs.bzl", "proto_library")
+load("//tools:defs.bzl", "ts_grpc_library")
+
+npm_link_all_packages(name = "node_modules")
+
+proto_library(
+    name = "proto",
+    srcs = ["example.proto"],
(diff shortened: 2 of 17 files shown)


jbedard:
With such a large example touching multiple areas and using multiple `rules_*` etc maybe it would be more appropriate to put this in https://github.com/aspect-build/bazel-examples?

thesayyn:
beware that we have https://github.com/aspect-build/bazel-examples/pull/224 already. could be a small addition to that.

satajo:
> With such a large example touching multiple areas and using multiple `rules_*` etc maybe it would be more appropriate to put this in https://github.com/aspect-build/bazel-examples?

Very good! I was also unsure of the fit of this PR into the `rules_ts` for the same reason, but I was not aware of the dedicated examples repo. It should be a small change to base this change onto that repo instead, and close this PR in favor of a new one.

Before doing that however, 

> beware that we have https://github.com/aspect-build/bazel-examples/pull/224 already. could be a small addition to that.

@thesayyn can you explain a bit more about this? Specifically, do think that instead of submitting this as a separate example, it would be better to replace the use of `rules-proto-grpc` and the npm-imported `@bufbuild/protoc-gen-*` binary usage with the tooling defined in https://github.com/aspect-build/bazel-examples/pull/224, effectively resulting in a single larger example?

If so, I could try to see about it once https://github.com/aspect-build/bazel-examples/pull/224 is merged. Assuming it's not too much of a job of course, this `rules-proto-grpc` based setup was already finicky enough to get working :stuck_out_tongue:

alexeagle:
FYI https://github.com/aspect-build/rules_ts/issues/415

PR #378, opened 2023-06-08
fix: add missing ts_version_from to bzlmod API

gregmagolan:
Bug in the sense that it should be in the bzlmod API since it is already in the WORKSPACE API.

---

### Type of change

### Test plan

- New test cases added

### DIFF
--- a/ts/extensions.bzl
+++ b/ts/extensions.bzl
@@ -8,11 +8,18 @@ load("//ts:repositories.bzl", "LATEST_TYPESCRIPT_VERSION")
 def _extension_impl(module_ctx):
     for mod in module_ctx.modules:
         for attr in mod.tags.deps:
-            npm_dependencies(ts_version = attr.ts_version, ts_integrity = attr.ts_integrity)
+            ts_version = attr.ts_version
+            if not ts_version and not attr.ts_version_from:
+                ts_version = LATEST_TYPESCRIPT_VERSION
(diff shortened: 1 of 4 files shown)


ts/extensions.bzl
```diff
@@ -8,11 +8,18 @@ load("//ts:repositories.bzl", "LATEST_TYPESCRIPT_VERSION")
...
-            npm_dependencies(ts_version = attr.ts_version, ts_integrity = attr.ts_integrity)
+            ts_version = attr.ts_version
+            if not ts_version and not attr.ts_version_from:
+                ts_version = LATEST_TYPESCRIPT_VERSION
+            npm_dependencies(ts_version = ts_version, ts_version_from = attr.ts_version_from, ts_integrity = attr.ts_integrity)
 
 ext = module_extension(
     implementation = _extension_impl,
     tag_classes = {
-        "deps": tag_class(attrs = {"ts_version": attr.string(default = LATEST_TYPESCRIPT_VERSION), "ts_integrity": attr.string()}),
```
jbedard, line 16: Why not use `default` anymore?

jbedard (approved):

ts/extensions.bzl
gregmagolan: `npm_dependencies` will complain if both ts_version and ts_version_from are set

PR #381, opened 2023-06-12
fixes #380 - add urls for dependencies

### LINKED ISSUE #380: [FR]: Allow ts to be downloaded through other URLs

When invoking `rules_ts_dependencies`, Typescript is downloaded from registry.npmjs.org (https://github.com/aspect-build/rules_ts/blob/main/ts/private/npm_repositories.bzl#L87) and there is no way to override this.

### Describe the feature

It would be nice to be able to supply another URL where typescript can be downloaded from (like bazelisk https://github.com/bazelbuild/bazelisk#where-does-bazelisk-get-bazel-from for example).

It would be possible to for example add an optional urls parameter to `npm_dependencies`.

In our company we proxy all external repositories through proxies to be able to revoke packages, and to get speed improvements by having a cache near our site.

hjellek:
Attempt at making urls to download Typescript from customizable by the user

### Type of change

### Test plan

### DIFF
--- a/docs/repositories.md
+++ b/docs/repositories.md
@@ -11,7 +11,7 @@ See https://docs.bazel.build/versions/main/skylark/deploying.html#dependencies
 ## rules_ts_dependencies
 
 <pre>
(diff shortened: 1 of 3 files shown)


hjellek:
Closing, found out about the bazel downloader rewrite feature

PR #383, opened 2023-06-28
Add tool version for TypeScript 5.1.5

ericrafalovsky:
## Type of change
Chore (any other change that doesn't affect source or test files, such as configuration)


mattem (approved):
Thanks!

PR #394, opened 2023-08-07
feat: allow json5 for tsconfig

### LINKED ISSUE #388: [FR]: Nit, possible for tsconfigs to support the json5 extension?

A set up like this yields the error below

```
ts_config(
    name = "project_root_base_tsconfig",
    src = "tsconfig.allRootPaths.json5", # This is the problem line here
)

ts_project(
    name = "all_ts",
    srcs = [
        ":stuff",
    ],
    extends = ":project_root_base_tsconfig",
)
```

> ERROR: /workspace/typescript/BUILD.bazel:175:11: in extends attribute of validate_options rule //typescript:_validate_all_ts_options: '//typescript:project_root_base_tsconfig' does not produce any validate_options extends files (expected .json). Since this rule was created by the macro 'ts_project', the error might have been caused by the macro implementation

Confirmed that this is an aspect-only restrict because `yarn tsc -b typescript/tsconfig.json` works well

### Describe the feature

Possible to accept .json5 extensions too?

alexeagle:
TypeScript apparently supports it, though it's not mentioned on the tsconfig docs.

Fixes #388

---

### Type of change

### Test plan

- New test cases added


alexeagle:
oh, our tsconfig validator program will need to be updated to understand json5 as well.

thesayyn (approved):

alexeagle:
I'm not going to find time for this :(

PR #401, opened 2023-08-09
chore: flip supports_workers to default False

### LINKED ISSUE #400: [2.0] disable persistent workers by default
### What happened?

They are still somewhat buggy. I just closed several issues about them because we'd like to stop supporting our custom worker-aware `tsc` binary.
Instead, 2.0 will encourage users to use swc as the transpiler, and wait for #374 to get fast type-checks.

### Version

2.0

### How to reproduce

_No response_

### Any other information?

_No response_

alexeagle:
Fixes #400

---

### Type of change

Persistent workers are now off by default.
See the docs for `ts_project#supports_workers` if you want to enable them again.

### Test plan

Well, actually I removed them because they were hard to read, but could restore them and figure out what the new logic should be.


thesayyn (approved):
wohooo!

thesayyn:
looks like you need to turn on workers explicitly for worker e2e.

e2e/test/tracing_and_performance.bats
```diff
@@ -15,7 +15,7 @@ teardown() {
     tsconfig
     ts_project --src "source.ts"
     echo "export const f = 1;" > source.ts
-    run bazel build :foo
+    run bazel build :foo --@aspect_rules_ts//ts:supports_workers
```
thesayyn, line 18: nit: you don't need these here as it's explicitly turned on in .bazelrc

alexeagle: I did need this to make the PR green. It's only in e2e/worker that the bazelrc contains the line, but AFAICT that directory doesn't contribute into these e2e setups under bats.

PR #402, opened 2023-08-09
chore(2.0): remove VERBOSE_LOGS env var

alexeagle:
### Type of change

`--define=VERBOSE_LOGS` no longer has an effect, use the `--@aspect_rules_ts//ts:verbose=true` flag instead

### Test plan


thesayyn (approved):
YES!

PR #408, opened 2023-08-14
fix(transpiler): allow explicit 'opt-in' to tsc

alexeagle:
This avoids the need to set a flag on the build.

BREAKING CHANGE:
- the type of `transpiler` on the ts_project rule (not the macro) is now a tri-state `[-1, 0, 1]`

---

### Type of change

### Test plan

- New test cases added
Changed one e2e to check this case.


e2e/bzlmod/.bazelrc
```diff
@@ -1,4 +1,4 @@
-# import common bazelrc shared with e2e workspaces
-import %workspace%/../../.bazelrc.common
+# NB: we don't import common bazelrc shared with e2e workspaces
```
jbedard, line 1: Why? What caused this in this PR?

jbedard (approved):

e2e/bzlmod/.bazelrc
alexeagle: I intentionally changed this as the regression test. It was erroring because the transpiler flag wasn't set. Now it's green because there's an explicit "tsc"

.bazelrc.common
```diff
@@ -70,8 +70,8 @@ build:macos --nolegacy_external_runfiles
 build:freebsd --nolegacy_external_runfiles
 build:openbsd --nolegacy_external_runfiles
 
-build --@aspect_rules_ts//ts:skipLibCheck=honor_tsconfig
-build --@aspect_rules_ts//ts:default_to_tsc_transpiler
+common --@aspect_rules_ts//ts:skipLibCheck=honor_tsconfig
```
thesayyn, line 73: nice!

thesayyn (approved):

PR #410, opened 2023-08-16
chore: add Aspect Workflows CI (on AWS + GitHub Actions)

gregmagolan:
Run CI on an Aspect Workflows deployment on AWS + GitHub Actions. Just the root WORKSPACE & buildifier for now.

I've included the terraform in this repo to serve as an live working example for the full configuration of Aspect Workflows for AWS + GitHub Actions.

---

### Type of change

### Test plan

Observe Aspect Workflows GitHub Actions CI working.


.github/workflows/aspect-workflows-template.yaml
```diff
@@ -0,0 +1,142 @@
+# This reusable GitHub Actions workflows must either be vendored into the target repository as is
+# the case here -or- a fork made of https://github.com/aspect-build/workflows-action repository and
+# the reusable workflow referenced from the fork within the same organization as the target
+# repository.
+#
+# At this time, GitHub Actions does not allow referencing reusable Workflows from public
+# repositories in other organizations. See
```
alexeagle, line 7: we are in the same organization, so you shouldn't need to vendor this file

alexeagle: I know your goal is to make this an example, but it's an example for sales calls, not for committed trial customers who are happy to have our guidance. So I think we ought to make this simple where possible.

.aspect/workflows/terraform/terraform.bzl
```diff
@@ -0,0 +1,20 @@
+"""Wrapper around a terraform sh_binary target"""
+
+def terraform(name, **kwargs):
```
alexeagle, line 3: IMO there's too much code here just for the purpose of pinning the terraform version. We like to do that when we're all-out crazy about reproducibility, but that's not what other terraform users do. I think we should keep our OSS footprints small, as there's enough code here that it's going to become a maintenance burden to manage many copies.

Possible alternative is to create/find&contribute a rules_terraform to fix the duplication problem, but my suggestion is we just run `terraform` like "normal engineers" in our OSS repos.

.aspect/workflows/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Aspect Workflows config yaml
+
+This is the `config.yaml` file in this directory.
+
+## GitHub Actions workflows
+
+This includes 3 files:
+
+-   `.github/workflows/aspect-workflows-template.yaml` : Aspect Workflows reusable workflow
```
alexeagle, line 23: let's drop this one? see below

.github/workflows/aspect-workflows-template.yaml
gregmagolan: I think these will serve as full fidelity examples for the docsite and for trial users who want to be more self-serve... but it is also a lot of code to vendor and maintain when we don't need to. Perhaps we use write_source_file to keep it up to date?

gregmagolan: My guess is that most trial users would prefer to vendor with a write_source_files vs fork the repo since forking requires another set of hoops to jump through with whoever is managing the GitHub account. The fork is also another place to have to sync whereas an http_archive to a particular version with a write_source_file is all in the same repo.

.aspect/workflows/terraform/terraform.bzl
gregmagolan: The lock file may jump around depending on who's running the terraform apply. Personally, I'm very used to running `bazel run terraform` and don't really want to worry about what version I've got install locally or if I even have one installed. Maybe a rules_terraform is the way to go here so we can re-use across our OSS repos for all the deployments and use it internally for silo as well.

.github/workflows/aspect-workflows-template.yaml
alexeagle: As the one datapoint here, I discussed with AAI eng leadership and they decided forking the repo is better, but I don't think we had a good write_source_files proposal for them at the time.

I don't have a strong opinion here, and it's perhaps "cheating" a bit to leave this file out just because the repo is under the aspect-build org so we can be "special"

.aspect/workflows/terraform/terraform.bzl
gregmagolan: I'll leave it out for now so this can land with less code but will think about an aspect-build/rules_terraform.

.github/workflows/aspect-workflows-template.yaml
gregmagolan: Ok, I'll go with write_source_files and mention at the use site that it can be forked instead.

gregmagolan: Guess I can point to the AAI repo as an example of that since its public.

.aspect/workflows/terraform/terraform.bzl
alexeagle: you can just `alias terraform=/absolute/path/to/silo/bazel-out/some/path/terraform` right? Is that enough solution for now to get the terraform binary you want for yourself?

gregmagolan: suppose it is for now so I don't have to go make a new rule set :)

.github/workflows/aspect_workflows_deps.bzl
```diff
@@ -0,0 +1,16 @@
+"""Aspect Workflows bazel dependencies"""
+
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+ASPECT_WORKFLOWS_VERSION = "5.7.0-rc4"
```
alexeagle, line 5: how will users discover that there's a newer release?
We can't add a test asserting that this version matches their terraform, because we don't plan to release workflows-action at every release, and also they often have their terraform in a separate repo

alexeagle (approved):

.github/workflows/aspect_workflows_deps.bzl
gregmagolan: I don't have a good answer for that. Typically customers will not have their Aspect Workflows GitHub Actions configuration in the same repository as their terraform. Updating the vendored workflows if there are fixes and/or new features will have to be part of the upgrade process. We'll need to be careful to keep new versions of Workflows backward compatible with older versions of the re-usable workflow.

PR #418, opened 2023-08-22
feat: read typescript version from resolved.json

### LINKED ISSUE #308: [FR]: ts_version_from using pnpm lock file

We can only pull the version from the package.json file and the version there must be a pinned version

### Describe the feature

Pulling from the pnpm lock file will allow the package.json specification to be unpinned such as `4.9.x`. We can also pull the integrity out of the lock file so users can easily upgrade to versions of typescript even before they are added to versions.bzl in rules_ts.

### Fund our work

- [ ] Sponsor our open source work by donating a [feature bounty](https://opencollective.com/aspect-build/)

alexeagle:
Fixes #308 

---

### Type of change

### Test plan

Built with update to this repo

### DIFF
--- a/docs/repositories.md
+++ b/docs/repositories.md
@@ -16,7 +16,7 @@ rules_ts_dependencies(<a href="#rules_ts_dependencies-ts_version_from">ts_versio
 
 Dependencies needed by users of rules_ts.
 
-To skip fetching the typescript package, define repository called `npm_typescript` before calling this.
+To skip fetching the typescript package, call `rules_ts_bazel_dependencies` instead.
 
 
 **PARAMETERS**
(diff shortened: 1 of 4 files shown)


docs/repositories.md
```diff
@@ -16,15 +16,15 @@ rules_ts_dependencies(<a href="#rules_ts_dependencies-ts_version_from">ts_versio
 
 Dependencies needed by users of rules_ts.
 
-To skip fetching the typescript package, define repository called `npm_typescript` before calling this.
+To skip fetching the typescript package, call `rules_ts_bazel_dependencies` instead.
```
jbedard, line 19: I don't think this should be getting updated? This is also the internal one?

ts/repositories.bzl
```diff
@@ -9,34 +9,18 @@ load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")
...
-            Exactly one of `ts_version` or `ts_version_from` must be set.
-        ts_version: version of the TypeScript compiler.
-            Exactly one of `ts_version` or `ts_version_from` must be set.
-        ts_integrity: integrity hash for the npm package.
-            By default, uses values mirrored into rules_ts.
-            For example, to get the integrity of version 4.6.3 you could run
-            `curl --silent https://registry.npmjs.org/typescript/4.6.3 | jq -r '.dist.integrity'`
-    """
-
+def rules_ts_bazel_dependencies():
```
jbedard, line 23: Can we put a `_` to make it clear this one is internal?

alexeagle: No, because then we'd get an error calling it from our WORKSPACE

docs/repositories.md
alexeagle: The new function introduced here really is public API. It's better than the suggestion I'm removing here, just like the case in this repo, you might want to install rules_ts bazel dependencies, then declare `@npm`, then load from it, so it's not nice to require that you create an `npm_typescript` repo before installing the bazel deps.

jbedard (approved):

alexeagle:
<details>
  <summary>Click me</summary>
  
  ### Heading
  1. Foo
  2. Bar
     * Baz
     * Qux

  ### Some Javascript
  ```js
  function logSomething(something) {
    console.log('Something', something);
  }
  ```
</details>

PR #439, opened 2023-09-03
feat: add check_for_update

alexeagle:
### Type of change

### Test plan

Did a build and watched for traffic on the cloudflare endpoint

### DIFF
--- a/ts/private/npm_repositories.bzl
+++ b/ts/private/npm_repositories.bzl
@@ -74,28 +74,39 @@ def _http_archive_version_impl(rctx):
         executable = False,
     )
 
-    if rctx.attr.check_for_updates:
-        vars = ["{}={}".format(v, rctx.os.environ[v]) for v in _TELEMETRY_VARS if v in rctx.os.environ]
-        if rctx.attr.bzlmod:
-            vars.append("bzlmod=true")
-        url = "https://update.aspect.build/aspect_rules_ts/{}?{}".format(
-            RULES_TS_VERSION,
-            "&".join(vars),
-        )
-        output_path = str(rctx.path(".output/update_check_result"))
-        command = ["curl", url, "--write-out", "%{http_code}", "--output", output_path]
-        result = rctx.execute(command)
-        if result.return_code != 0:
-            # Ignore failures when trying to check for new version
-            return
-        status_code = int(result.stdout.strip())
-        if status_code != 302:
-            # 304: Not Modified means we have the latest version already
-            return
+    if not rctx.attr.check_for_updates:
+        return
 
-        # buildifier: disable=print
-        print("NOTICE: a newer version of rules_ts is available")
-        # TODO: print content of output_path which now has the link to the newer version
+    if RULES_TS_VERSION.startswith("$Format"):
+        # The placeholder string wasn't replaced.
+        # That means we aren't running from a release artifact, so we don't know the version
+        # and can't tell if an update is available.
+        return
+
+    vars = ["{}={}".format(v, rctx.os.environ[v]) for v in _TELEMETRY_VARS if v in rctx.os.environ]
+    if rctx.attr.bzlmod:
+        vars.append("bzlmod=true")
+    url = "https://update.aspect.build/aspect_rules_ts/{}?{}".format(
+        RULES_TS_VERSION,
+        "&".join(vars),
+    )
+    output_path = str(rctx.path(".output/update_check_result"))
+    command = ["curl", url, "--write-out", "%{http_code}", "--output", output_path]
+    result = rctx.execute(command)
+    if result.return_code != 0:
+        # Ignore failures when trying to check for new version
+        return
+    status_code = int(result.stdout.strip())
+    if status_code != 302:
+        # 304: Not Modified means we have the latest version already
+        return
+
+    # buildifier: disable=print
+    # TODO: print content of output_path which now has the link to the newer version
+    print("""\
+NOTICE: a newer version of rules_ts is available.
+See https://github.com/aspect-build/rules_ts/releases
+""")
 
 http_archive_version = repository_rule(
     doc = "Re-implementation of http_archive that can read the version from package.json",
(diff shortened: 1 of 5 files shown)


ts/private/npm_repositories.bzl
```diff
@@ -63,10 +74,35 @@ def _http_archive_version_impl(rctx):
         executable = False,
     )
 
+    if rctx.attr.check_for_updates:
+        vars = ["{}={}".format(v, rctx.os.environ[v]) for v in _TELEMETRY_VARS if v in rctx.os.environ]
+        if rctx.attr.bzlmod:
+            vars.append("bzlmod=true")
+        url = "https://update.aspect.build/aspect_rules_ts/{}?{}".format(
+            "v0.0.0",  # FIXME
```
alexeagle, line 82: This is quite hard actually, since our release is so simple we don't have a way to stamp it with our own version number.

alexeagle: resolved in #441

ts/private/npm_repositories.bzl
```diff
@@ -63,10 +74,35 @@ def _http_archive_version_impl(rctx):
...
+        if result.return_code != 0:
+            # Ignore failures when trying to check for new version
+            return
+        status_code = int(result.stdout.strip())
+        if status_code != 302:
+            # 304: Not Modified means we have the latest version already
+            return
+
+        # buildifier: disable=print
+        print("NOTICE: a newer version of rules_ts is available")
```
gregmagolan, line 97: Can just add a link to https://github.com/aspect-build/rules_ts/releases to find the latest release.

ts/private/npm_repositories.bzl
```diff
@@ -63,10 +74,35 @@ def _http_archive_version_impl(rctx):
         executable = False,
     )
 
+    if rctx.attr.check_for_updates:
+        vars = ["{}={}".format(v, rctx.os.environ[v]) for v in _TELEMETRY_VARS if v in rctx.os.environ]
+        if rctx.attr.bzlmod:
+            vars.append("bzlmod=true")
+        url = "https://update.aspect.build/aspect_rules_ts/{}?{}".format(
+            RULES_TS_VERSION,
```
gregmagolan, line 82: If someone is consuming from source they'll send `0.0.0` as the RULES_TS_VERSION? How is that handled by `update.aspect.build`?

ts/private/npm_repositories.bzl
alexeagle: yeah that works for now

ts/private/npm_repositories.bzl
alexeagle: No they would send this `$Format...` string actually https://github.com/aspect-build/rules_ts/blob/main/ts/private/versions.bzl#L9
I should filter that out here instead.

alexeagle: The downside is that anyone using a commit SHA rather than a release will just not get counted in our telemetry. I think that's a pretty small fraction of traffic so we can live with that?

ts/private/npm_repositories.bzl
```diff
@@ -63,10 +74,35 @@ def _http_archive_version_impl(rctx):
...
+        )
+        output_path = str(rctx.path(".output/update_check_result"))
+        command = ["curl", url, "--write-out", "%{http_code}", "--output", output_path]
+        result = rctx.execute(command)
+        if result.return_code != 0:
+            # Ignore failures when trying to check for new version
+            return
+        status_code = int(result.stdout.strip())
+        if status_code != 302:
+            # 304: Not Modified means we have the latest version already
```
jbedard: 302 or 304? We're checking one but commenting on the other.

ts/private/npm_repositories.bzl
jbedard: Can we not send that SHA in that case? I'm not sure if that's accessible anywhere...?

jbedard: I guess we couldn't (easily) check for updates with only a SHA but could still be valuable?

jbedard:
I thought we'd do this on rules_js first, is this on ts first for the 2.0 release?

ts/private/npm_repositories.bzl
gregmagolan: We won't know the SHA. It isn't baked into the sources anywhere.

Counting unpinned uses is valuable. Don't think you should filter it out at the service. Unpinned would mean they are bleeding edge from HEAD for some reason or have forked potentially forked it.

jbedard: I was wondering if the SHA from the original http_archive might somehow be available, but I assume not.

We could put the version or commit hash into the source but that might not be worth it?

jbedard: I do think we should send every usage possible though. Let the client send any info available and let the server decide if/what to ignore

ts/private/npm_repositories.bzl
```diff
@@ -63,10 +74,48 @@ def _http_archive_version_impl(rctx):
...
+
+    vars = ["{}={}".format(v, rctx.os.environ[v]) for v in _TELEMETRY_VARS if v in rctx.os.environ]
+    if rctx.attr.bzlmod:
+        vars.append("bzlmod=true")
+    url = "https://update.aspect.build/aspect_rules_ts/{}?{}".format(
+        RULES_TS_VERSION,
+        "&".join(vars),
+    )
+    output_path = str(rctx.path(".output/update_check_result"))
+    command = ["curl", url, "--write-out", "%{http_code}", "--output", output_path]
```
gregmagolan, line 95: add a short timeout as well so we're not blocking for very long if the service is down

ts/private/npm_repositories.bzl
alexeagle: Okay, I rewrote it as `v0.0.0` and still call the service. On the server-side, we'll have to special-case this and not report that there's an update available.
(I considered making this something like v99.99.99 so it would artificially appear as "later than all releases" but I'd rather not invent something new. We already understand 0.0.0 is the "from HEAD" version in our rulesets.)

jbedard (approved):

PR #441, opened 2023-09-04
feat: stamp our version into release

alexeagle:
We can use this at runtime to support features like checking for updates.

### DIFF
--- a/.github/workflows/release_prep.sh
+++ b/.github/workflows/release_prep.sh
@@ -9,6 +9,14 @@ TAG=${GITHUB_REF_NAME}
 PREFIX="rules_ts-${TAG:1}"
 ARCHIVE="rules_ts-$TAG.tar.gz"
 
+# Don't include examples in the distribution artifact, to reduce size
+# Note, we do need to include e2e/bzlmod since BCR runs our test based on distribution artifact.
+# Also stamp the release version into the repo
+cat >.git/info/attributes <<EOF
+examples                 export-ignore
+ts/private/versions.bzl  export-subst
+EOF
+
 git archive --format=tar --prefix=${PREFIX}/ ${TAG} | gzip > $ARCHIVE
 SHA=$(shasum -a 256 $ARCHIVE | awk '{print $1}')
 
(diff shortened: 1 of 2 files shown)


.github/workflows/release_prep.sh
```diff
@@ -9,6 +9,13 @@ TAG=${GITHUB_REF_NAME}
 PREFIX="rules_ts-${TAG:1}"
 ARCHIVE="rules_ts-$TAG.tar.gz"
 
+# Don't include examples in the distribution artifact, to reduce size
+# Note, we do need to include e2e/bzlmod since BCR runs our test based on distribution artifact.
+echo >>.git/info/attributes "examples export-ignore"
+
+# Stamp our release version into source files
+git grep -l '"0.0.0"' | xargs sed -i "s/0.0.0/${TAG:1}/"
```
gregmagolan, line 17: hmmm. we may pickup `0.0.0` all over the place in lock files.. should we make it 0.0.0-PLACEHOLDER?

alexeagle: in my testing, the double quotes prevented that and only matched on our files. But I'm investigating the `export-subst` feature of https://git-scm.com/docs/gitattributes#_creating_an_archive to make this more targeted

alexeagle: The git version on my machine is old, so testing "live in prod" to be sure https://github.com/aspect-build/rules_esbuild/commit/a8c4379990372c48bc05aea608caf929e59b485d

alexeagle: Current GHA runners have

```
Getting Git version info
  Working directory is '/home/runner/work/rules_esbuild/rules_esbuild'
  /usr/bin/git version
  git version 2.41.0
```

alexeagle: Okay, verified this new approach:
https://github.com/aspect-build/rules_esbuild/releases/tag/v0.15.3

I had to upgrade `git` on my machine before I could make it work locally as well.
PTAL

kormide (approved):

alexeagle:
Note: I'm not going to try to upstream this approach to the bazel-contrib/rules-template since it doesn't cover the root MODULE.bazel file and so there's still a patch required for the BCR, I think ppl will object and point out that the stamp mechanism should replace that file too.

PR #446, opened 2023-09-11
chore: add coverage to Aspect Workflows configuration

gregmagolan:


gregmagolan:
Added coverage to rules_jest intead

PR #478, opened 2023-11-09
fix: allow dict() tsconfig defined at analysis time

jbedard:
This was an interesting one, I'm surprised we haven't seen it before. See https://bazel.build/rules/language#mutability

### Type of change

### Test plan

- New test cases added


thesayyn (approved):
LGTM. interesting

Issue #436, opened 2023-08-31
[Bug]: ts_project#assets not copied to bin


alexeagle:
### What happened?

the `assets` parameter may point to artifacts which later tools will want to read from `bazel-out`

Similar to https://github.com/aspect-build/rules_ts/issues/250#issuecomment-1428108866
Context: https://bazelbuild.slack.com/archives/CEZUUKQ6P/p1692468805524629

### Version

HEAD

xzdev:
Is there any update on the issue? Anyone can help? We are trying to use rules_webpack and without this issue been fixed, we cannot use the rule to bundle resources file from the packages.

PR #493, opened 2023-11-27
fix: respect ts_project root_dir and out_dir for assets

jbedard:
Fix https://github.com/aspect-build/rules_ts/issues/436


ts/private/ts_project.bzl
```diff
@@ -141,6 +142,17 @@ See https://github.com/aspect-build/rules_ts/issues/361 for more details.
     tsconfig_inputs = copy_files_to_bin_actions(ctx, _validate_lib.tsconfig_inputs(ctx).to_list())
     inputs.extend(tsconfig_inputs)
 
+    assets_outs = []
+    for a in ctx.files.assets:
+        a_path = _lib.relative_to_package(a.short_path, ctx)
+        a_out = _lib.to_out_path(a_path, ctx.attr.out_dir, ctx.attr.root_dir)
+        if a_out != a:
```
gregmagolan, line 149: Would this work without the conditional? You could just use

```
            asset = ctx.actions.declare_file(a_out)
            copy_file_action(ctx, a, asset)
```
in either case and it would mimic the copy_to_bin behavior if the output path was the same

gregmagolan (approved):

ts/private/ts_project.bzl
jbedard: Yeah that works and simplifies things. Logically I think it ends up doing the exact same thing but reduces the code here 👍

YvanGuidoin:
This causes an issue if the asset file is generated by another rule.
Looks like ts_project is not using _calculate_assets_outs but directly to_out_path which skips the test that input filepath == output filepath

jbedard:
@YvanGuidoin can you provide an example test case? A PR adding a failing test would be best, or log an issue with a snippet reproducing it if it's easy enough.

YvanGuidoin:
@jbedard tried to create a testcase that gives me this issue in a PR, should we follow up there ?

jbedard:
Yeah, lets discuss it there 👍

PR #509, opened 2023-12-21
examples: reproduce swc paths resolving to sandbox

jackvincentnz:
### Type of change

- Issue reproduction

### Motivation

- Minimal reproduction of sandbox paths being present in transpiled javascript after swc transpilation of `.ts` imports using tsconfig paths.

### Context

I am bundling javascript modules that are transpiled using swc. The bundler cannot resolve the paths transpiled via swc when tsconfig paths are configured since the transpiled paths contain sandbox paths. 

Prior to switching to swc, tsc ignores these imports (treating them as pure javascript) and so there is no issue. With swc, these paths are resolved to their relative paths during transpilation. This works for relative paths that are alongside the tsconfig, but not when those relative paths are nested from the tsconfig.

### Reproduction steps

1. `bazel build //examples/grandparent`.
2. review `bazel-bin/examples/grandparent/index.js` (resolved path is relative to .ts file).

```
import { parentName } from "./parent/index";
export const grandparentName = parentName + "'s grandparent";
```

3. review `bazel-bin/examples/grandparent/parent/index.js` (resolved path is relative to sandbox).

```
import { name } from "../../../../../../../../private/var/tmp/_bazel_jack.vincent/5f33d7eefddc1bdc8bb8bb38dea57fd4/sandbox/darwin-sandbox/797/execroot/grandparent/parent/child";
export const parentName = name + "'s parent";
```

The only difference between the parent and the grandparent is the distance from the `tsconfig` in the directory tree. The grandparent is in the same directory, and the parent is in a nested directory.

### DIFF
--- a/examples/grandparent/BUILD.bazel
+++ b/examples/grandparent/BUILD.bazel
@@ -1,0 +1,33 @@
+load("@aspect_rules_swc//swc:defs.bzl", "swc")
+load("@aspect_rules_ts//ts:defs.bzl", "ts_config", "ts_project")
+load("@bazel_skylib//lib:partial.bzl", "partial")
+load("@npm//examples:tsconfig-to-swcconfig/package_json.bzl", tsconfig_to_swcconfig = "bin")
+
+ts_config(
+    name = "tsconfig",
+    src = "tsconfig.json",
+    visibility = [":__subpackages__"],
+)
+
+tsconfig_to_swcconfig.t2s(
+    name = "write_swcrc",
+    srcs = ["tsconfig.json"],
+    args = [
+        "--filename",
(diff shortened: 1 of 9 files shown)


jackvincentnz:
I have added another repro: https://github.com/jackvincentnz/bazel-swc-paths-bug-repo.

I think it better shows swc and tsc working as expected outside Bazel in comparison to not when within Bazel.

jackvincentnz:
Closing in favour of https://github.com/aspect-build/rules_swc/pull/223, and waiting on the upstream fix: https://github.com/swc-project/swc/issues/4057.

Seems to be a duplicate of https://github.com/aspect-build/rules_swc/issues/192

PR #510, opened 2023-12-22
chore: upgrade to Aspect Workflows 5.9.0-rc.9

gregmagolan:
Add configure task

### DIFF
--- a/.aspect/workflows/bazelrc
+++ b/.aspect/workflows/bazelrc
@@ -1,6 +1,6 @@
 # build without the bytes
-common --remote_download_minimal
-common --nobuild_runfile_links
+test --remote_download_minimal
+test --nobuild_runfile_links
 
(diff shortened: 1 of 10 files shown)


.aspect/workflows/bazelrc
```diff
@@ -1,6 +1,6 @@
 # build without the bytes
-common --remote_download_minimal
-common --nobuild_runfile_links
+test --remote_download_minimal
```
jbedard, line 2: What was the reason for this change?

gregmagolan: Looks like a bug in Bazel for the handling of `common`. It turns out that `bazel query` doesn't like `common --remote_download_minimal`. We now do a `bazel query` in the delivery manifest step.

kormide (approved):

PR #520, opened 2024-01-23
fix: wire the validation action

### LINKED ISSUE #517: [Bug]: ts_project macro creates an unused validate_options target if validate is True
### What happened?

foo/BUILD.bazel:
```bazel

ts_project(
    name = "foo",
    srcs = [
        "index.ts",
    ],
    validate = True,
)
```
The ts_project macro above creates two targets: //foo:foo (ts_project_rule) and //foo:_validate_foo_options (validate_options). Currently //foo:_validate_foo_options is not used by //foo:foo. I think //foo:foo should depend on //foo:_validate_foo_options.

### Version

Output of `bazel --version`: bazel 7.0.1

`WORKSPACE` or `MODULE.bazel` file: aspect_rules_ts 2.1.0

Language(s) and/or frameworks involved: TypeScript

### How to reproduce

Minimal repro: https://github.com/PeterD1524/cuddly-octo-giggle

Running
```shell
bazel query '...'
```
will show
```
//foo:_validate_foo_options
//foo:foo
```
Running
```shell
bazel build '//foo:foo'
```
will succeed.

Running
```shell
bazel build '//foo:_validate_foo_options'
```
will fail with the following error:
```
ERROR: ts_project rule //foo:foo was configured with attributes that don't match the tsconfig
 - attribute declaration=false does not match compilerOptions.declaration=true
```

I think `bazel build '//foo:foo'` should fail.

### Any other information?

related commit: https://github.com/aspect-build/rules_ts/commit/18a506393ef1fe988702863bbd30beb03ab835d2
validate_options creates a target with a validation action which checks the tsconfig.
suggestions how to fix:
ts_project_rule can accept a new attribute label which should be a validate_options target and the ts_project macro should send the validate_options target it created to ts_project_rule.

alexeagle:
Sadly in #138 we only did this halfway: it stopped depending on the output of the validation action, but didn't actually execute one either. It has been 'failed open' this whole time, so likely users have ts_project attributes that have drifted from the tsconfig settings.

BREAKING CHANGE:
the validator is enabled again, so it will complain about ts_project attributes that don't match tsconfig. You should just run the buildozer commands it prints to repair this.

Fixes #517 

---

### Type of change

Will need to follow the instructions printed by the validator to green up the build.

### Test plan

1. Make an error in an attribute and see that it's reported again now.
2. Test with `--norun_validations`
3. Set `validate=False`

This really needs an integration test case but sadly don't have time right now.


ts/private/ts_validate_options.bzl
```diff
@@ -51,20 +59,18 @@ To disable this check, set the validate attribute to False:
         resolve_json_module = ctx.attr.resolve_json_module,
         source_map = ctx.attr.source_map,
         incremental = ctx.attr.incremental,
-        ts_build_info_file = ctx.attr.ts_build_info_file,
+        # ts_build_info_file = ctx.attr.ts_build_info_file,
```
jbedard, line 62: ?

alexeagle: err yeah... there's no such attribute on ts_project so I think it was a mistake here

alexeagle: oh I see, there's a string-typed `ts_build_info_file` and then `ts_project#buildinfo_out` was a slightly different label.

Fixed this the simple (and conservative) way to restore how it worked before by adding another attribute.

jbedard (approved):

PR #531, opened 2024-01-27
chore: re-enable RBE testing

alexeagle:
Revert #529 with fixes

Suspicious workarounds to avoid protobuf problems: non-hermeticity with protoc->glibc and something in C++ toolchain

### Type of change

### Test plan


alexeagle:
fyi @sluongng

thesayyn (approved):

PR #539, opened 2024-02-05
chore: remove Go

alexeagle:
We don't have any Go sources in rules_ts, so there's no reason we should install a Go toolchain or worry about upgrading it. 'aspect configure' provides the auto-generation for bzl_library targets so we don't need Gazelle either.


alexeagle:
Confirmed that aspect workflows already has a `configure` task, so we shouldn't lose CI checking that bzl_library targets are up-to-date.

alexeagle:
gazelle is currently load-bearing for testing Aspect Workflows on AWS+GHA so closing this for now. Can revisit when/if we stop recommending gazelle.

PR #543, opened 2024-02-08
refactor: remove unnecessary default param values

jbedard:
BREAKING CHANGE to the ts_lib API.

These defaults are never used internally, and are different defaults then the rules actually use which is confusing.

---

### Type of change

### Test plan


alexeagle (commented):
Technically these are exposed through the ts_lib and this change will probably break someone. What's the motivation?

jbedard:
I didn't think those were public? The reason is just that they are never used and add confusion. I think the real defaults are `None` yet the (unused) default param values are `"."`?

gregmagolan (approved):
🌮

PR #544, opened 2024-02-08
perf: use starlark string.removeprefix

jbedard:
I think this is even in 5.1 and ok for rules_ts? https://github.com/bazelbuild/bazel/commit/50bb742fc35c04ab422a7ce723160b27d6900e6c

---

### Type of change

- Performance (a code change that improves performance)

### Test plan

### DIFF
--- a/ts/private/ts_lib.bzl
+++ b/ts/private/ts_lib.bzl
@@ -164,15 +164,12 @@ def _join(*elements):
         return "/".join(segments)
     return "."
 
-def _strip_external(path):
-    return path[len("external/"):] if path.startswith("external/") else path
-
 def _relative_to_package(path, ctx):
-    for prefix in [ctx.bin_dir.path, ctx.label.workspace_name, ctx.label.package]:
-        prefix += "/"
-        path = _strip_external(path)
-        if path.startswith(prefix):
-            path = path[len(prefix):]
+    # TODO: "external/" should only be needed to be removed once
+    path = path.removeprefix("external/").removeprefix(ctx.bin_dir.path + "/")
+    path = path.removeprefix("external/").removeprefix(ctx.label.workspace_name + "/")
+    if ctx.label.package:
+        path = path.removeprefix("external/").removeprefix(ctx.label.package + "/")
     return path
 
 def _is_typings_src(src):


alexeagle (approved):
This seems fine, are you starting from some profile of the analysis time that showed you this code was inefficient?

jbedard:
I know creating the large temp arrays has a side-effect. The temp array in this `for` loop is tiny, but the method is invoked frequently.

This change on it's own made no significant difference (when using `ts_project` + `swc`) but https://github.com/aspect-build/rules_swc/pull/234, https://github.com/aspect-build/rules_swc/commit/273eace5351b01a5321e82c5696b3775b36f85f5 and https://github.com/aspect-build/rules_swc/commit/6bf6610494cbb7589b5e34f3d834abd8c19e55d0 had very significant improvements so I'd prefer making this a pattern even if individual cases are not always significant.

jbedard:
Also repeating `path.removeprefix("external/")` 3x should make this functionally the same, but points out that we probably don't need it 3x?

ts/private/ts_lib.bzl
```diff
@@ -164,15 +164,11 @@ def _join(*elements):
...
-def _strip_external(path):
-    return path[len("external/"):] if path.startswith("external/") else path
-
 def _relative_to_package(path, ctx):
-    for prefix in [ctx.bin_dir.path, ctx.label.workspace_name, ctx.label.package]:
-        prefix += "/"
-        path = _strip_external(path)
-        if path.startswith(prefix):
-            path = path[len(prefix):]
+    path = path.removeprefix("external/").removeprefix(ctx.bin_dir.path + "/")
```
alexeagle, line 168: I think we should only remove "external/" once now like you suggest. It seems like it was just wrong before

jbedard: I think it would only be needed once *after* removing `ctx.bin_dir.path`?

jbedard: Although we have no tests breaking if I remove it so I think I'd prefer leaving it for now and just putting a TODO?

jbedard: Unless you're more confident then I am? 😅 

For now I just put a TODO...

Issue #551, opened 2024-02-22
[Bug]: ts_project does not support root_dir = ".."


jgao54:
### What happened?

When root_dir is set to "..", instead of a successful build, got vague errors like:
```
output 'scripts/script_2.js' was not created
Transpiling TypeScript project @//scripts:bad [tsc -p scripts/tsconfig_bad.json] failed: not all outputs were created or valid
Target //scripts:bad failed to build
```

### Version

Development (host) and target OS/architectures:

Output of `bazel --version`: 
- 7.0.2

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file:
- rules_ts-v2.1.1
- bazel-lib-v1.40.0

Language(s) and/or frameworks involved:
- typescript

### How to reproduce

Minimum Repro: https://github.com/jgao54/rule-ts-bug-minimum-repro/tree/main

PR #552, opened 2024-02-23
feat: add ts_project validation of root/out/typing_out dir options

jbedard:
To catch issues that may otherwise result in hard to understand messages, especially `root_dir` where the rule output declarations are declared incorrectly causing a "not all outputs were created" error.

Close https://github.com/aspect-build/rules_ts/issues/551

---

### Type of change

### Test plan


jbedard:
WDYT? Worth adding?

alexeagle (approved):
yeah for sure, thanks!

PR #568, opened 2024-03-15
fix: typecheck_test depends on the .d.ts files

### LINKED ISSUE #567: [Bug]: build_test is always a cache hit
### What happened?

Looking at examples/transpiler in this repo,  the inputs to the babel_typecheck_test are

```
% bazel aquery //examples/transpiler:babel_typecheck_test | grep Inputs
  Inputs: [bazel-out/k8-fastbuild/bin/examples/transpiler/babel_typecheck_test.sh, bazel-out/k8-fastbuild/bin/examples/transpiler/babel_typecheck_test.sh.runfiles/MANIFEST, bazel-out/k8-fastbuild/bin/examples/transpiler/babel_typecheck_test_0__deps.out]
```
that `babel_typecheck_test_0__deps.out` file is an empty marker file. This means that this test doesn't ever fail, and it's always a cache hit once it has run. It only forces some build actions to be added to the graph - if the target doesn't type-check, it's a build failure, not a test failure.

### Version

HEAD as of today

### How to reproduce

_No response_

### Any other information?

_No response_

alexeagle:
Fixes #567
workaround for https://github.com/bazelbuild/bazel-skylib/issues/480

---

### Type of change

### Test plan

### DIFF
--- a/ts/private/build_test.bzl
+++ b/ts/private/build_test.bzl
@@ -1,0 +1,35 @@
+"""Fork of @bazel_skylib//rules:build_test.bzl
+
+Workaround for https://github.com/bazelbuild/bazel-skylib/issues/480
+"""
+
+_DOC = """\
+build_test is a trivial test that always passes.
+
+It is used to force Bazel to build some of its dependencies, in the case where
+`--build_tests_only` would have otherwise caused those to be skipped.
+"""
+
+def _build_test_impl(ctx):
+    executable = ctx.actions.declare_file(ctx.label.name + ".sh")
+    ctx.actions.write(
+        output = executable,
+        is_executable = True,
+        content = "#!/usr/bin/env bash\nexit 0",
+    )
+
+    return DefaultInfo(
+        files = depset([executable]),
+        executable = executable,
+        # The critical fix, so that the test inputs include the default outputs of the targets to check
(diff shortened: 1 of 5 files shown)


ts/private/build_test.bzl
```diff
@@ -0,0 +1,35 @@
...
+    ctx.actions.write(
+        output = executable,
+        is_executable = True,
+        content = "#!/usr/bin/env bash\nexit 0",
+    )
+
+    return DefaultInfo(
+        files = depset([executable]),
+        executable = executable,
+        # The critical fix, so that the test inputs include the default outputs of the targets to check
```
jbedard, line 24: > so that the test **outputs** include
?

alexeagle: no, it's an input to the test. We want the test to be a cache miss when that input changes.

jbedard: I think there's just too many inputs + outputs and this doesn't clarify which. This is the "test action" inputs? Not the `build_test` inputs?

jbedard:
Is there some bazel-skylib code this is based on or copied from that could be linked to in the rule implementation?

jbedard (approved):

alexeagle:
It's not copied from skylib, the only relevant link would be to the repo where @dzbarsky already has this patch running.

PR #569, opened 2024-03-15
feat: provide a way for individual ts_project targets to reserve extr…

alexeagle:
…a RAM

Some targets are globs of a large amount of code. By default Bazel allocates only 250MB See https://github.com/bazelbuild/bazel/commit/d7f0724b6b91b6c57039a1634ff00ccebd872714#diff-139b184febb86269f1ceee705bde7274e2efcccaa3f435a8c6cb2f1f0854ce46R96

---

### Type of change

### Test plan

- New test cases added


ts/private/ts_lib.bzl
```diff
@@ -29,6 +81,14 @@ See more details on the `assets` parameter of the `ts_project` macro.
     "out_dir": attr.string(
         doc = "https://www.typescriptlang.org/tsconfig#outDir",
     ),
+    "resource_set": attr.string(
```
jbedard, line 84: I feel like if we're using the bazel `resource_set` term we should be using the bazel `resource_set` API and not create our own enum API. How will this current API expand in the future to support changing the CPU attribute?

Maybe just rename this to `resource_set_memory`? Or use the bazel `resource_set` API?

ts/private/ts_lib.bzl
```diff
@@ -3,6 +3,58 @@
...
+    return { "memory": 8192 }
+
+def _resource_set_mem_16g(_, __):
+    return { "memory": 16384 }
+
+def _resource_set_mem_32g(_, __):
+    return { "memory": 32768 }
+
+# buildifier: disable=function-docstring
+def resource_set(name):
```
DavidZbarsky-at, line 39: if you make this `_resource_set` you don't need the buildifier disable

alexeagle: but it is loaded by another file so the symbol can't be private

alexeagle:
Landed https://github.com/aspect-build/bazel-lib/pull/792

This ought to depend on bazel-lib 2.0 now, but we have the usual problem that it's a breaking change. Maybe we can have a separate entry point for ts_project_with_resource_set. Or just wait for rules_ts 3.0? Or give up on our adherence to semver and just require bazel-lib 2.0 in a 2.x release of rules_ts.

gregmagolan:
Related internal thread: https://aspect-build.slack.com/archives/C05S94VJWNM/p1710536751433789?thread_ts=1710508945.458429&cid=C05S94VJWNM

alexeagle:
now that `main` targets rules_ts 3.x and is guaranteed to have bazel-lib 2.x, I'm unblocked to rely on the new resource_sets helper :)

ts/private/ts_lib.bzl
alexeagle: updated to use the feature that landed in bazel-lib 2.x

jbedard (approved):

PR #571, opened 2024-03-19
 fix: handle various ts_project srcs labels syntaxes

jbedard:
### Type of change

### Test plan

- New test cases added


ts/private/ts_lib.bzl
```diff
@@ -190,6 +190,9 @@ def _is_ts_src(src, allow_js, resolve_json_module):
 
     return _is_js_src(src, allow_js, resolve_json_module)
 
+def _is_asset_src(src):
+    return not src.startswith(":") and not src.startswith("//")
```
jbedard, line 194: Note that the other `_is_*_src` methods only check `src.endswith(".ext")`, but "assets" don't have any known or expected extension so instead of checking if the `src` "looks like a source file" we're doing the opposite and checking if it does NOT "look like a target". I think the js/dts predeclared outputs actually have a bug here...

Aghassi (approved):

ts/private/ts_lib.bzl
alexeagle: this doesn't seem very principled, I'm free to omit a leading `:` when referring to another target. Could we use some API like `native.package_relative_label()` to try taking the thing you gave us and convert to a label? Even then I don't think we can distinguish since source files are valid labels as well.

jbedard: Summary after some discussions:

* `package_relative_label` just prefixes everything with `@//the/package:` and doesn't give any more info. It is `attr.files.x.is_source` in rule implementations that would be ideal
* pre-declaring `.d.ts`, `.js` etc outputs is useful, despite the issues this PR shows (essentially "guessing" if a `ts_project(srcs)` entry is a file vs target)
* pre-declaring asset outputs is probably not useful and worth removing in 3.x (IMO)

alexeagle (approved):

Issue #88, opened 2022-07-11
Typecheck-only rule


alexeagle:
So far we told users they have to use `tsc_test` if they just want a rule that errors when `.d.ts` inputs have typing errors. That's because Bazel won't run actions whose outputs aren't needed, and there are no outputs of a `ts_project(srcs=["foo.d.ts"])` where we just want the typechecker to run.

But Bazel 6 has validation actions, as observed here https://github.com/aspect-build/rules_ts/issues/85#issuecomment-1180520106
so then we could have a typecheck-only rule, maybe it would be `ts_project` with some different behavior.

PR #574, opened 2024-03-26
Support For Typecheck-Only Validation Action

jacobgardner:
---

### Type of change

My attempt at fixing #88 as a validation action.

  - Requires bazel 6 or higher for validation action support.
  - Removes some warnings about typecheck only not being supported.

ts_project no longer needs to emit declaration files in order to work with a non-tsc transpiler.  If no declaration emit and no transpilation takes place, `tsc --noEmit` is still run as a validation action. 

### Test plan

- New test cases added

### DIFF
--- a/e2e/test/common.bats
+++ b/e2e/test/common.bats
@@ -121,20 +121,6 @@ function tsconfig() {
 EOF
 }
 
-function load_mock_transpiler() {
-    local path="."
-
-    while (( $# > 0 )); do
-        case "$1" in
-        --path) shift; path="$1"; shift ;;
-        *) break ;;
-        esac
-    done
-    
-    cat > "$path/BUILD.bazel" <<EOF
-load("@aspect_rules_ts//ts/test:mock_transpiler.bzl", "mock")     
-EOF 
-}
 
 function ts_project() {
     local path="."
@@ -174,6 +160,7 @@ function ts_project() {
 load("@aspect_rules_ts//ts:defs.bzl", "ts_project")
 ${npm_link_all_packages}load("@npm//:defs.bzl", "npm_link_all_packages")
 ${npm_link_all_packages}npm_link_all_packages(name = "node_modules")
+load("@aspect_rules_ts//ts/test:mock_transpiler.bzl", "mock")     
 
 ts_project(
     name = "${name}",
--- a/e2e/test/common.bats
+++ b/e2e/test/common.bats
@@ -77,6 +77,7 @@ EOF
   fi
 }
 
+
 function tsconfig() {
     local path="."
     local no_implicit_any="false"
@@ -120,6 +121,20 @@ function tsconfig() {
 EOF
 }
 
+function load_mock_transpiler() {
+    local path="."
+
+    while (( $# > 0 )); do
+        case "$1" in
+        --path) shift; path="$1"; shift ;;
+        *) break ;;
+        esac
+    done
+    
+    cat > "$path/BUILD.bazel" <<EOF
+load("@aspect_rules_ts//ts/test:mock_transpiler.bzl", "mock")     
+EOF 
+}
 
 function ts_project() {
     local path="."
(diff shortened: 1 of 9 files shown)


ts/test/transpiler_tests.bzl
```diff
@@ -132,35 +117,6 @@ def transpiler_test_suite():
         ],
     )
 
-    # This target proves that transpilation doesn't require typechecking:
```
jacobgardner, line 135: This test is essentially moved to `e2e/test/validation.bats` because it needs to be run with `--norun_validations`  for it to preserve the existing functionality.

jfirebaugh:
@jacobgardner Thanks for working on this! I tested it out in our repo at Figma, and it works well. I did have to make an addition on top of your changes, to support an explicit `no_emit` attribute for `ts_project`, which we needed for our use case. (It's a bit niche, but we have two `ts_project` targets with overlapping .ts inputs, each typechecking with a different tsconfig. These targets are to be used only for typechecking and linting. Our bundler is a separate target.) My additions are here: https://github.com/jfirebaugh/rules_ts/commit/f5df77ca655b7b0fbe9d83cfe8317ac4319f0a21

jacobgardner:
Sweet. I will merge your changes in soon.

jacobgardner:
I merged the no_emit code. I want to add some tests for that later this week as well though.

jfirebaugh:
I've found another hiccup with this approach: when using it in tandem with [rules_lint](https://github.com/aspect-build/rules_lint) / [aspect lint](https://docs.aspect.build/cli/commands/aspect_lint), the typechecking actions will trigger even when you're just trying to run eslint. This is because the validations output group "is special in that its outputs are always requested, regardless of the value of the `--output_groups` flag, and regardless of how the target is depended upon" (https://bazel.build/extending/rules#validations_output_group).

I think perhaps `aspect lint` should pass `--run_validations=false` to the underlying bazel command it runs? Or provide a hook so that we can add that in `.bazelrc`. For example `aspect lint` could run `bazel build --config=lint ...` and then I could put `config:lint --run_validations=false` in `.bazelrc`.

jfirebaugh:
A more serious issue with this approach is that it negates attempts to use [`ts_project#transpiler`](https://docs.aspect.build/rulesets/aspect_rules_ts/docs/transpiler#ts_projecttranspiler) to avoid typechecking in situations when only the transpiled outputs are necessary, as a performance optimization. The behavior of validation actions makes typechecking always eager again.

Is a validation action actually providing value in this case? It still requires a "maker" output. Could we instead just have a normal action that outputs the same marker?

e2e/test/common.bats
```diff
@@ -133,6 +134,7 @@ function ts_project() {
     local source_map=""
     local declaration=""
     local composite=""
+    local tranpsiler=""
```
jfirebaugh, line 137:
```suggestion
    local transpiler="" 
```

e2e/test/common.bats
```diff
@@ -146,6 +148,7 @@ function ts_project() {
         --source-map) shift; source_map="source_map = True," ;;
         --declaration) shift; declaration="declaration = True," ;;
         --composite) shift; composite="composite = True," ;;
+        --mockTranspiler) shift; composite="transpiler = mock," ;;
```
jfirebaugh, line 151:
```suggestion
        --mockTranspiler) shift; transpiler="transpiler = mock," ;;
```

And then `$transpiler` should be inserted in the appropriate place in the heredoc string below.

ts/private/ts_project.bzl
```diff
@@ -248,25 +224,32 @@ This is an error because Bazel does not run actions unless their outputs are nee
...
-            },
-        )
+
+    # Validation actions still need to produce some output, so we output a .validation file
+    # that ends up in the _validation output group.
+    if supports_workers:
+        run_cmd = "{} $@".format(executable.path)
+        arguments.add_all(["--bazelValidationFile", validation_output.short_path])
+    else:
+        run_cmd = """{} $@ && echo "" > {} """.format(executable.path, validation_output.path)
```
alexeagle, line 234: instead of `echo ""` could we just use https://github.com/aspect-build/rules_js/blob/main/js/private/js_binary.sh.tpl#L22

PR #590, opened 2024-04-16
feat(ts_proto_library): add connect-query plugin

mrmeku:
Prior to this PR ts_proto_library could only
generate nodejs service. Now it can generate both
connect-query (grpc-web / react) based services.

---

### Type of change

### Test plan
- New test cases added


e2e/bzlmod/BUILD
```diff
@@ -15,8 +15,8 @@ proto_library(
 
 ts_proto_library(
     name = "foo_ts_proto",
-    has_services = False,
     copy_files = False,
+    gen_connect_es = False,
```
alexeagle, line 19: normally we couldn't land a breaking change on `main` - but the top of `proto.md` claims that this rule isn't yet a stable public API so we can still get away with it here :)

alexeagle (approved):
Could you include the release notes for the breaking change in the PR description?

alexeagle:
Never mind, I updated PR description.

PR #593, opened 2024-04-20
feat: upgrade to rules_js 2.0

gregmagolan:
---

### Type of change

### Test plan


examples/project_references/lib_b/BUILD.bazel
```diff
@@ -27,13 +26,13 @@ ts_project(
...
-    name = "test",
-    args = ["*.spec.js"],
-    chdir = package_name(),
-    data = [
-        "index.spec.js",
-        ":lib_b",
-    ],
-    node_modules = "//examples:node_modules",
-)
+# jasmine_test(
```
jbedard, line 29: What was the issue here?

ts/private/ts_project.bzl
```diff
@@ -6,31 +6,25 @@ load("@aspect_bazel_lib//lib:paths.bzl", "to_output_relative_path")
...
-        target[JsInfo].transitive_npm_linked_package_files
+        target[JsInfo].npm_sources
         for target in targets
-        if JsInfo in target and hasattr(target[JsInfo], "transitive_npm_linked_package_files")
-    ])
-    files_depsets.extend([
-        target[NpmPackageStoreInfo].transitive_files
-        for target in targets
-        if NpmPackageStoreInfo in target and hasattr(target[NpmPackageStoreInfo], "transitive_files")
+        if JsInfo in target and hasattr(target[JsInfo], "npm_sources")
```
jbedard, line 27: Why do we need the `hasattr`? Doesn't the factory method always set it?

examples/project_references/lib_b/BUILD.bazel
gregmagolan: We need to bump the rules_jasmine dep to be compatible with rules_js 2.0. It will be easier to just remote the akward dep between rules_ts and rules_jasmine and use a straight js_test instead.

ts/private/ts_project.bzl
gregmagolan: It does but we can't say for certain that other custom rule sets that derive from rules_js will set it so we check as a defence-in-depth.

examples/project_references/lib_b/BUILD.bazel
gregmagolan: ✅

.bazeliskrc
```diff
@@ -1,2 +1,2 @@
 BAZELISK_BASE_URL=https://github.com/aspect-build/aspect-cli/releases/download
-USE_BAZEL_VERSION=aspect/5.8.19
+USE_BAZEL_VERSION=aspect/5.9.25
```
jbedard, line 2: Should we do some of these in 2.x?

.gitignore
```diff
@@ -2,3 +2,12 @@ bazel-*
 .bazelrc.user
 **/.terraform/*
 node_modules/
+
+# Don't commit lockfile for now as it is unstable. Do allow for it to be
```
jbedard, line 6: This isn't in 2.x? 🤔

.bazeliskrc
gregmagolan: We could but I'd rather optimize for time here since there are 9 of these downstream repos and if we start pre-factoring that is dozens more PRs to create and manage. I miss monorepo.

.gitignore
gregmagolan: 2.x is still on Bazel 6. MDOULE.bazel.lock generation didn't come until Bazel 7.

.bazeliskrc
gregmagolan: Ugh. The bazel 7 upgrade snowballed and now it is half of this PR. I'll pre-factor that to main.

gregmagolan: https://github.com/aspect-build/rules_ts/pull/598

alexeagle (approved):

PR #603, opened 2024-04-30
chore: drop dependency on rules_jasmine

gregmagolan:
Final pre-factor from https://github.com/aspect-build/rules_ts/pull/593 🤞 

The dep on rules_jasmine is not needed here. In general these are easier to maintain if they only dep on rules_js. In the case of the rules_js 2.0 bump it is much easier to do here if the rules_jasmine dep is gone.


kormide (approved):

MODULE.bazel
```diff
@@ -30,31 +30,19 @@ bazel_dep(name = "rules_go", version = "0.46.0", dev_dependency = True, repo_nam
 bazel_dep(name = "rules_nodejs", version = "5.8.2", dev_dependency = True)
 bazel_dep(name = "stardoc", version = "0.6.2", dev_dependency = True, repo_name = "io_bazel_stardoc")
 
-npm = use_extension(
```
jbedard, line 33: What are all these changes from?

jbedard:
Shouldn't we just need `dev_dependency = True` when declaring the dep?

Although this dep seems pointless so I guess the cleanup is nice either way 👍

MODULE.bazel
gregmagolan: Just formatting cleanup. My main goal here with these pre-factors is to reduce the diff on https://github.com/aspect-build/rules_ts/pull/593 as that is the big hairy one to review.

gregmagolan:
> Shouldn't we just need `dev_dependency = True` when declaring the dep?
> 
> Although this dep seems pointless so I guess the cleanup is nice either way 👍

As noted in the summary, having the additional dep just makes it harder to maintain long term.

MODULE.bazel
gregmagolan: The big hairy one should now be much easier to review: https://github.com/aspect-build/rules_ts/pull/593

PR #617, opened 2024-06-02
fix(ts_proto_library): require explicit srcs to copy

alexeagle:
The hack of using a glob over the srcs causes a collision when there is more than one ts_proto_library in a bazel package.
The resulting error doesn't give a clue what fix is required.

---

- Breaking change (forces users to change their own code or config): yes (note, ts_proto_library isn't a stable API yet)

When ts_proto_library#copy_files is True, users must now repeat the .proto files whose outputs should be copied.

### Test plan

- New test cases added


mrmeku (approved):
It’s a shame we are in macro land and can’t reference the ProtoInfo provider. Maybe in the future we can convert the rule to use the implementation more directly

https://docs.bazel.build/versions/5.3.1/skylark/lib/ProtoInfo.html#direct_sources

PR #620, opened 2024-06-04
always set 'use_default_shell_env' in ctx.actions.run

malt3:
Aspect uses shell wrappers that invoke tools like `dirname` and `uname`. When the default shell env is not inherited, this prevents tools from being found on systems without a working fallback $PATH (like NixOS).

This might be incompatible with setting the `env` attribute, unless the bazel flag `--incompatible_merge_fixed_and_default_shell_env` is set on Bazel versions < 7.

See also: https://github.com/NixOS/nixpkgs/issues/289505

---

### Test plan

### DIFF
--- a/.github/workflows/BUILD.bazel
+++ b/.github/workflows/BUILD.bazel
@@ -3,7 +3,8 @@ load("@aspect_bazel_lib//lib:bazelrc_presets.bzl", "write_aspect_bazelrc_presets
 write_aspect_bazelrc_presets(
     name = "update_aspect_bazelrc_presets",
     presets = [
-        "bazel6",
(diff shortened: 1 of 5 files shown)


dimatosaurus:
thanks for this fix! any chance to merge that?

keegancsmith:
FYI this PR unblocks me using rules_ts on my NixOS development machine. Would be great to remove the `single_version_override` in our work repo :)

malt3:
Friendly ping to @alexeagle

alexeagle (approved):

keegancsmith:
CI seems to be stuck?

PR #621, opened 2024-06-04
ts_proto_library error on proto with import_prefix property

egormodin:


liningpan:
Is the solution here striping away `proto_info.proto_source_root` for all direct_sources like https://github.com/protocolbuffers/protobuf/blob/f396f7cb9fc6d8fcd897f3d88c05721642db6b46/bazel/common/proto_common.bzl#L75-L84 ?

liningpan:
Right now there is also no way to configure `es_out`, which might also interact with this feature in unexpected ways.

PR #622, opened 2024-06-05
Error on build `ts_proto_library` from external proto

egormodin:
Error:
`Error in directory_path: illegal rule name: _@com_google_protobuf//:timestamp_pb.d.ts_dirpath: invalid target name '_@com_google_protobuf//:timestamp_pb.d.ts_dirpath': target names may not contain '//' path separators`

Second issue: `proto` file is not usually accessible, but the `d.ts` files is needed to be copied, so there is no valid way to build it into source tree.

I suppose the d.ts file should be created from external proto and copied into source tree. Since there is no access to the `proto` file then the rule should be able to use only `proto_library`. By adding dependency into `ts_proto_library` I am trying to ask compiler to use same ts file as I am able to use in the code.


egormodin:
@thesayyn

egormodin:
In the current tapestry project we are using `go_googleapis` protos from https://github.com/googleapis/googleapis/tree/master/google, but I was not able to modify WORKSPACE file to bring it in to the example project

thesayyn:
No worries, i believe timestamp proto is different than googleapis as for WKT types no generation needed. I'll fix it. 

BTW you can do in `MODULE.bazel` for rules_ts.

```
http_archive = use_repo_rule("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
http_archive(
    name = "googleapis", 
    urls = ["https://github.com/googleapis/googleapis/archive/ef8d4740570bd099dd1a484e200a3d9433f793e2.tar.gz"]
)
```

egormodin:
right, no generation needed, but the current example results with the same error that I have by using `go_googleapis`.

egormodin:
btw to add `@go_googleapis` space we have a rule in the project WORKSPACE file
```
load("@com_google_api//:repository_rules.bzl", "switched_rules_by_language")

switched_rules_by_language(
    name = "com_google_googleapis_imports",
    go = True,
    grpc = True,
    java = True,
)
```

PR #625, opened 2024-06-07
chore: add --check_direct_dependencies to .bazelrc

gregmagolan:
Setting this flag in our the rule sets we maintain. Don’t want to push a rules author to update their deps if not needed.

https://bazel.build/reference/command-line-reference#flag--check_direct_dependencies

https://bazelbuild.slack.com/archives/C014RARENH0/p1691158021917459?thread_ts=1691156601.420349&cid=C014RARENH0

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -25,6 +25,11 @@ build --repo_env=JAVA_HOME=../bazel_tools/jdk
 # verbose
 common:verbose --@aspect_rules_ts//ts:verbose --worker_verbose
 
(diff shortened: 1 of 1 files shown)


jbedard (approved):

alexeagle:
Note, I just added this to bazel-lib presets, since we think all our rules and customers should have it?

gregmagolan:
> Note, I just added this to bazel-lib presets, since we think all our rules and customers should have it?

We just want this off for rule sets right? Default is warning and is correct for terminal workspaces.

PR #632, opened 2024-06-24
refactor: remove unnecessary JsInfo hasattr calls

jbedard:
### Test plan

### DIFF
--- a/ts/private/ts_project.bzl
+++ b/ts/private/ts_project.bzl
@@ -19,12 +19,12 @@ def _gather_types_from_js_infos(targets):
     files_depsets.extend([
         target[JsInfo].transitive_types
         for target in targets
-        if JsInfo in target and hasattr(target[JsInfo], "transitive_types")
+        if JsInfo in target
     ])
     files_depsets.extend([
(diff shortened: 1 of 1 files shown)


ts/private/ts_project.bzl
```diff
@@ -19,12 +19,12 @@ def _gather_types_from_js_infos(targets):
     files_depsets.extend([
         target[JsInfo].transitive_types
         for target in targets
-        if JsInfo in target and hasattr(target[JsInfo], "transitive_types")
```
thesayyn, line 22: Hmm, this looks like it was put here for backwards compatibility. Are we sure that it no longer an issue?

jbedard: rules_ts v3 (which `main` is for) only supports rules_js 2+ (and bazel-lib 2+) so I think this is ok

thesayyn (approved):
LGTM

ts/private/ts_project.bzl
thesayyn: okay, that makes sense

PR #636, opened 2024-06-25
support isolated declarations

MichaelMitchell-at:

- Searched for relevant documentation and updated as needed: no

### Test plan

- Manual testing; please provide instructions so we can reproduce: Used in the Airtable monorepo. Projects using isolated declarations pass `isolated_declarations=True` to `ts_project`. `bazel build --build_tag_filters=typing --output_groups=types,typecheck <target>`


alexeagle:
Rebased on the TS upgrade, so this example should be functional now. I'd still like to verify all the actions have expected inputs and outputs, and maybe add an analysis test showing that.

ericrafalovsky:
Perhaps out of scope of this PR, but it would be interesting to consider an API similar to https://docs.aspect.build/rulesets/aspect_rules_ts/docs/rules#transpiler to allow specifying a non-tsc decleration emit tool (such as [swc](https://github.com/swc-project/swc/pull/9093))

alexeagle:
TypeScript doesn't ship a release yet that calls the new `transpileDeclaration` API, and I don't know of another tool that does. Maybe we'll have to cut scope here to just support the part of the feature which verifies your code is ready for `isolatedDeclarations` semantics but doesn't actually give the speed improvement yet.

alexeagle:
Merged a PR on rules_swc to mirror the newest release that has the declaration emit feature, I'll give it a try...

jbedard:
Completed in https://github.com/aspect-build/rules_ts/pull/679

PR #640, opened 2024-06-26
chore: pin another e2e that was floating to get main green

alexeagle:


alexeagle:
Never mind, this is just making us work with a broken version that won't work for anyone else either ;(

PR #646, opened 2024-06-29
feat: upgrade examples to TS 5.5

alexeagle:
Pre-factoring for #636 

---

### Test plan

### DIFF
--- a/examples/package.json
+++ b/examples/package.json
@@ -1,7 +1,7 @@
 {
     "private": true,
     "dependencies": {
-        "@angular/compiler-cli": "^18",
+        "@angular/compiler-cli": "18.1.0-next.0",
         "@angular/compiler": "^18",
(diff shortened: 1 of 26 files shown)


examples/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "private": true,
     "dependencies": {
-        "@angular/compiler-cli": "^18",
+        "@angular/compiler-cli": "18.1.0-next.0",
```
jbedard, line 4: Why change this and only this one?

jbedard (approved):

examples/package.json
alexeagle: Ah, it's required since Angular 18.0 has a TS upper bound of <5.5.0
I could bump the other two packages but only the compiler-cli needs the fix.
Temporary until Angular 18.1

PR #647, opened 2024-07-01
Add example for isolatedDeclarations

alexeagle:
Supports the new isolatedDeclarations feature, as shipped in TypeScript 5.5.

This doesn't yet change the Bazel dependency graph to make things faster, because the tooling for declarationEmit is still evolving. This PR only introduces the "stricter" mode for explicit typing on exported symbols that allows codebases to prepare for faster typechecking.

---

`ts_project` can now enforce that sources are compliant with `isolatedDeclarations` semantics. In the future, this allows for faster Typechecking of a graph of `ts_project` rules.

### Test plan

- New test cases added


ts/defs.bzl
```diff
@@ -346,7 +355,7 @@ def ts_project(
         native.filegroup(
             name = typecheck_target_name,
             srcs = [tsc_target_name],
-            # This causes the types to be produced, which in turn triggers the tsc action to typecheck
+            # This causes the declarations to be produced, which in turn triggers the tsc action to typecheck
```
jbedard, line 358: This is reverting https://github.com/aspect-build/rules_ts/commit/9c3445e060f5db958b7285224e99d25354cc89b1#diff-fd5b55e792f846b8312680267efd413bf73bfe655630485e1514db77eb7d187cL363

ts/defs.bzl
```diff
@@ -152,6 +153,10 @@ def ts_project(
             See https://www.typescriptlang.org/docs/handbook/compiler-options.html#compiler-options
             Typically useful arguments for debugging are `--listFiles` and `--listEmittedFiles`.
 
+        isolated_declarations: Whether to enforce that declaration output (.d.ts file) can be produced for a
```
jbedard, line 156: Now we're mixing `types` and `declarations`, but this is aligning with the tsconfig so we should use `declarations`?

I'm still uncertain if we should have done that :/

ts/defs.bzl
```diff
@@ -152,6 +153,10 @@ def ts_project(
...
+            single source file at a time. Requires some additional explicit types on exported symbols.
```
jbedard, line 157: Is this "Equivalent to the TypeScript --isolatedDeclarations option." like the docs for other params say?

ts/defs.bzl
alexeagle: oh, I'll just change it back, no reason that was included in this PR

jbedard (approved):

alexeagle:
Hm, thinking about this with fresh eyes today. The only reason we currently mirror any attributes from tsconfig or `args` as attributes to the rule is when we need to know the value in order to predict what outputs will be written and where. In this case we don't need to know that - you're free to enable `isolatedDeclarations` and not tell `ts_project` about it - and we do exactly the same thing.
I'll close this and make more progress with the `declarations_transpiler` attribute instead.

alexeagle:
May as well use this one to add the example, since it lets us play with the new flag in manual testing to reason about what happens, and it will be useful to build on top of. You already reviewed it so let's not do it again :)

PR #648, opened 2024-07-02
feat: emit declarations with a tool other than tsc

alexeagle:
Mirrors how the `transpiler` attribute works.

Here's a profile, showing that `frontend` and `backend` targets are type-checked in parallel with `core` even though they both depend on it. The small green boxes in the upper-left are the 10ms `swc` action that produces the `core/index.d.ts` file needed by the type-checker for `frontend` and `backend`

Screenshot from 2024-07-02 15-45-24

---

### Test plan

- New test cases added


examples/isolated_declarations/backend/BUILD.bazel
```diff
@@ -5,7 +5,8 @@ ts_project(
     name = "backend",
     declaration = True,
     tsconfig = "//examples/isolated_declarations:tsconfig",
-    deps = ["//examples/isolated_declarations/core"],
+    # FIXME: should just be core:core
```
alexeagle, line 8: @jbedard this is the next step in the PR, to make the SWC-transpiled declarations appear as the types on the terminal rule in the macro produced for `core:core`. I have to juggle the logic for when to produce a `js_library` target "tree".

Also before landing this I think the declaration producing rule should be in rules_swc rather than a bzl file here.

matthewjh:
@alexeagle this is really exciting

jbedard:
Done in https://github.com/aspect-build/rules_ts/pull/697

PR #650, opened 2024-07-12
feat(ts_proto_library): expose protoc_gen_options on ts_proto_library

mattem:
Exposes a dict for setting options on protoc_gen on the `ts_proto_library`

---

- Suggested release notes appear below:

Allow custom options to be set on `ts_proto_library` for protoc-gen.

### Test plan

Run locally patched in Silo.


jbedard:
Can we make this an explicit argument in the public macro and document it there?

ts/private/ts_proto_library.bzl
```diff
@@ -21,6 +18,11 @@ def _protoc_action(ctx, proto_info, outputs, options = {
     """
     inputs = depset(proto_info.direct_sources, transitive = [proto_info.transitive_descriptor_sets])
 
+    options = dict(ctx.attr.protoc_gen_options, **{
+        "keep_empty_files": True,
+        "target": "js+dts",
```
mrmeku, line 23: The way the rule is written, the target NEEDS to be set this way. Please update the rule such that if the user passes any target other than this we tell them it won't work

jbedard: Currently this will overwrite anything the user passes. Should we error instead?

mrmeku (approved):

PR #651, opened 2024-07-15
test: ts_project transpiler linked with failing tsc

jbedard:
Passes after https://github.com/aspect-build/rules_js/pull/1850

---

### Test plan

- New test cases added


examples/lib_nocompile_linked/index.ts
```diff
@@ -0,0 +1,5 @@
+console.log(3 + global)
+
+const B: string = 42
```
gregmagolan, line 3: nit: add a comment that this is an intentional type-check failure

gregmagolan (approved):

Issue #667, opened 2024-08-06
[Bug]: `ts_proto_library` protoc cannot execute generated plugin js_binary *.bat file wrappers with forward-slashes


willjschmitt:
### What happened?

Building a `ts_proto_library` (with or without connect/query enabled) on Windows, we get an error suggesting `'bazel-out' is not recognized as an internal or external command, operable program or batch file.`: 
```
bazel build :foo_ts_pb --verbose_failures
INFO: Analyzed target //:foo_ts_pb (0 packages loaded, 0 targets configured).
ERROR: C:/users/willschmitt/documents/achillea/rules_ts_repro/BUILD.bazel:15:17: Generating .js/.d.ts from //:foo_ts_pb failed: (Exit 1): protoc.exe failed: error executing ProtocGenEs command (from target //:foo_ts_pb)
  cd /d C:/users/willschmitt/_bazel_willschmitt/hqqsioi3/execroot/_main
  SET BAZEL_BINDIR=bazel-out/x64_windows-fastbuild/bin
  bazel-out\x64_windows-opt-exec-ST-d57f47055a04\bin\external\protobuf~\protoc.exe --plugin=protoc-gen-es=bazel-out/x64_windows-opt-exec-ST-d57f47055a04/bin/_foo_ts_pb.gen_es.bat --es_opt=keep_empty_files=true --es_opt=target=js+dts --es_out=bazel-out/x64_windows-fastbuild/bin --plugin=protoc-gen-connect-es=bazel-out/x64_windows-opt-exec-ST-d57f47055a04/bin/_foo_ts_pb.gen_connect_es.bat --connect-es_opt=keep_empty_files=true --connect-es_opt=target=js+dts --connect-es_out=bazel-out/x64_windows-fastbuild/bin --descriptor_set_in bazel-out/x64_windows-fastbuild/bin/foo_proto-descriptor-set.proto.bin foo.proto
# Configuration: a5a9a62db26d025cbc2fa4b4fbccebfc6b077137210c76db063d38b8ff442d14
# Execution platform: @@platforms//host:host
'DOSKEY' is not recognized as an internal or external command,
operable program or batch file.
'bazel-out' is not recognized as an internal or external command,
operable program or batch file.
--es_out: protoc-gen-es: Plugin failed with status code 1.
Target //:foo_ts_pb failed to build
INFO: Elapsed time: 1.155s, Critical Path: 0.09s
INFO: 2 processes: 2 internal.
[...]

PR #668, opened 2024-08-06
fix: convert forward slashes to backslashes for protoc plugins on Windows.

willjschmitt:
Fixes https://github.com/aspect-build/rules_ts/issues/667

This takes the approach [rules_proto_grpc does](https://github.com/rules-proto-grpc/rules_proto_grpc/blob/9e61a4c3dfc78de163febe7530bd04dabfbcca16/modules/core/internal/protoc.bzl#L79-L83) to detect if we are building on Windows via the `host_path_separator`. protoc will fail to call into the plugins if they are expressed to protoc as forward-slash paths. Alternatively, we could make this more limited like my patch in the linked issue to check `.bat` file extensions, which is fragile in its own way, but works for the given use case.

---

### Test plan
- Manual testing; please provide instructions so we can reproduce: check out https://github.com/AchilleaResearch/rules_ts_667 on Windows and `bazel build //:foo_ts_pb`. This PR's change is a patch in that repo, so I could also change that to directly use my commit instead, if preferred

There aren't any tests as far as I can tell currently for ts_proto_library, so my manual test is the best I can currently offer, and it's not clear if the Windows test runners are fully active, but I'd be happy to write a test for initial coverage on `ts_proto_library` if desired


jbedard:
Would you be able to add some initial `ts_proto_library` tests? It would be great to merge some tests in and then try to reproduce+fix the windows after 🙏

willjschmitt:
Rebased, and the unit test that would repro and show this PR improving things is [f00e046](https://github.com/aspect-build/rules_ts/pull/671/commits/f00e0460e94bac7900b84a82cba8b202f271b05d) in #671

willjschmitt:
Now that #671 is merged in, I rebased and the same unit test should apply, showing a noop on Linux and now should work on Windows

willjschmitt:
Any chance this can be re-reviewed based on my last comment in reply to adding test coverage?

PR #671, opened 2024-08-11
fix: join `--descriptor_set_in` with host path separator

### LINKED ISSUE #670: [Bug]: `--descriptor_set_in` input files should be joined by `;` on Windows rather than `:` in `ts_proto_library`
### What happened?

Building a `ts_proto_library` with a proto library that has transitive dependencies fails to build with a "<entire-descriptor_set_in-arg>: No such file or directory" error

Inspecting the arguments provided to protoc, it looks like the paths are separated with `:` characters, which makes sense, since it's the typical construction on Linux, but it turns out the Windows protoc build wants to see them `;`-separated. This is reinforced looking at some examples of how `descriptor_set_in` is joined in the protobuf repo: https://github.com/protocolbuffers/protobuf/blob/b9d1cfff8ca6814723889bade011f3fa4675d46d/protobuf.bzl#L294-L298 (other examples: https://github.com/search?q=repo%3Aprotocolbuffers%2Fprotobuf+--descriptor_set_in&type=code), where they string join on `ctx.configuration.host_path_separator`. 

I don't see where the use of a system path separator for path delimiting is documented (the docs just say "delimited"), but the parsing by this character is defined here: https://github.com/protocolbuffers/protobuf/blob/b9d1cfff8ca6814723889bade011f3fa4675d46d/src/google/protobuf/compiler/command_line_interface.cc#L2134-L2157, with the path separator conditionally defined to `:` or `;` at https://github.com/protocolbuffers/protobuf/blob/b9d1cfff8ca6814723889bade011f3fa4675d46d/src/google/protobuf/compiler/command_line_interface.cc#L950-L954

The only other use of this delimiter [is in `--proto_path`](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/compiler/command_line_interface.cc#L2071-L2073), but I don't see any uses of `--proto_path` in this repo

### Version

Development (host) and target OS/architectures: Windows 10

Output of `bazel --version`: 7.1.1

`WORKSPACE` or `MODULE.bazel` file:
 * rules_ts - 2.4.2 (patched with #668)

Typescript, protobuf

### How to reproduce

_No response_

### Any other information?

_No response_

willjschmitt:
Fixes #670.

As described in #670, protoc splits the arguments to `--proto_path` and `--descriptor_set_in` using an OS-specific path-separator. On posix, this is `:`, but on Windows this is `;`. The protobuf library takes the approach for its bazel rules to join on `ctx.configuration.host_path_separator`, so I've taken the same approach here as well.

---

### Test plan
https://github.com/AchilleaResearch/rules_ts_667/tree/repro_670 demonstrates the fix as a patch. `bazel build //:foo_ts_pb` on Windows with and without the patch https://github.com/AchilleaResearch/rules_ts_667/blob/repro_670/MODULE.bazel#L47

### DIFF
--- a/ts/private/ts_proto_library.bzl
+++ b/ts/private/ts_proto_library.bzl
@@ -47,7 +47,7 @@ def _protoc_action(ctx, proto_info, outputs):
         args.add_joined(["--connect-query_out", ctx.bin_dir.path], join_with = "=")
 
     args.add("--descriptor_set_in")
-    args.add_joined(proto_info.transitive_descriptor_sets, join_with = ":")
+    args.add_joined(proto_info.transitive_descriptor_sets, join_with = ctx.configuration.host_path_separator)
 
     args.add_all(proto_info.direct_sources)
 
--- a/ts/test/BUILD.bazel
+++ b/ts/test/BUILD.bazel
@@ -4,12 +4,15 @@ load(":flags_test.bzl", "ts_project_flags_test_suite")
 load(":mock_transpiler.bzl", "mock")
 load(":transpiler_tests.bzl", "transpiler_test_suite")
 load(":ts_project_test.bzl", "ts_project_test_suite")
+load(":ts_proto_library_test.bzl", "ts_proto_library_test_suite")
 
 transpiler_test_suite()
 
 ts_project_test_suite(name = "ts_project_test")
 
 ts_project_flags_test_suite(name = "ts_project_flags_test")
+
+ts_proto_library_test_suite(name = "ts_proto_library_test")
 
 # Ensure that when determining output location, the `root_dir` attribute is only removed once.
 ts_project(
--- a/ts/test/ts_proto_library/BUILD.bazel
+++ b/ts/test/ts_proto_library/BUILD.bazel
@@ -3,8 +3,6 @@
 # since protoc would otherwise not find the proto files in the descriptor
 # database.
 
-load("//ts:proto.bzl", "ts_proto_library")
-
 proto_library(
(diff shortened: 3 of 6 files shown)


jbedard:
Can you rebase this to kickoff CI again? I'm not sure why it's red atm.

This is simple enough (and a noop on *nix) I think we can merge it, but if you have a chance to add some tests along with those for https://github.com/aspect-build/rules_ts/pull/668 that would be great.

willjschmitt:
@jbedard sorry for the delay, I was out on baby bonding leave and just am getting back in. I rebased, but I don't think it's auto kicking off a build. I manually ran `bazel test //...` on linux and ran `bazel test //ts/test:ts_proto_library_with_dep_test` on windows, cherry picking #668, although unfortunately, due to https://github.com/aspect-build/rules_js/issues/1739, a patch to handle bsdtar on windows is still needed. The Windows build is failing but getting past this and the issues of #667 at least

I did borrow from the node_modules in `//examples/proto_grpc`, since the repo root is not the pnpm workspace root, which might not be acceptable, so let me know if 1) the existing examples essentially are already the test coverage we are looking for, or 2) the pnpm workspace can/should be promoted to repo root, which I figured wouldn't be desirable

jbedard (approved):

jbedard:
Looks like we just have a buildifier error, otherwise LGTM and thanks for the tests 👍

willjschmitt:
Sorry about that! Removed the extraneous imports, so we should be good to go now

PR #673, opened 2024-08-14
fix(ts_proto_library): Depend on @bufbuild/protobuf

mrmeku:
Adds the peer dependency of proto-gen-es to the
deps of the js_binary used to run it

---

### Test plan

Tested on my local repo which was experiencing the module resolution error


alexeagle (commented):
This seems wrong, it's a regular dep not a peerDep
https://www.npmjs.com/package/@bufbuild/protoc-gen-es?activeTab=dependencies
and we don't import from `@bufbuild/protobuf` so it seems you're being forced to add a transitive?
That means there's a bug somewhere else

PR #676, opened 2024-08-29
perf: reduce array array operations

jbedard:
Reduce unnecessary and duplicate array loops and concatenating.

---

### Test plan


gregmagolan (approved):
Logic looks right to me

PR #677, opened 2024-08-29
fix: support custom tsc version numbers when integrity passed

jbedard:
If a `ts_integrity` is passed then the `TOOL_VERSIONS[version]` is not used and this validation is incorrect. The `TOOL_VERSIONS[version]` lookup already has its own `fail` if the `version` is invalid so the removed check is unnecessary.

---

### Test plan

- New test cases added


ts/private/npm_repositories.bzl
```diff
@@ -28,9 +28,6 @@ def _http_archive_version_impl(rctx):
             ts = p["dependencies"]["typescript"]
         else:
             fail("key 'typescript' not found in either dependencies or devDependencies of %s" % json_path)
-        if any([not seg.isdigit() for seg in ts.split(".")]):
```
alexeagle, line 31: What error message does a user get instead, if they use `ts_version_from="//package.json"` and the package.json has `"typescript": "^4.9.0"`? I suspect it will be a lot less helpful than this one

jbedard: https://github.com/aspect-build/rules_ts/blob/1a42ef4fc44a90c03d42dd5bab7e03f68a5bd739/ts/private/npm_repositories.bzl#L40-L42

I think that is the correct location for the `fail` but we can update the message...

alexeagle: yeah update to "If so, you must X or Y"

alexeagle (approved):

PR #679, opened 2024-08-29
Isolated declarations

jbedard:
---

### Test plan

- New test cases added


jbedard:
Depends on https://github.com/aspect-build/rules_ts/pull/679

ts/defs.bzl
```diff
@@ -153,6 +154,10 @@ def ts_project(
             See https://www.typescriptlang.org/docs/handbook/compiler-options.html#compiler-options
             Typically useful arguments for debugging are `--listFiles` and `--listEmittedFiles`.
 
+        async_typecheck: Whether to type-check asynchronously as a separate bazel action.
```
jbedard, line 157: TODO: names:

* emphasize type-checking is a separate action: (parallel|async|defer|non_blocking|independent)(_typecheck)?
* emphasize transpiling is faster and doesn't wait for deps: (untyped|nocheck|parallel)(_transpile)?
* do it automatically instead of opt-in when isolated_declarations && is_ts56

jbedard: @dzbarsky [says](https://github.com/aspect-build/rules_ts/pull/698#discussion_r1769024468)
> For us we did a single flag enforce_isolated_declarations that enabled this mode + isoDecls in the tsconfig. Do you think there's any case where someone would set isoDecls in tsconfigs but not want to enable this?

One reason would be when not on typescript >=5.6. Otherwise I'm not sure... would it ever be slower running tsc twice like this?

Overall I'm actually in favour of the 3rd option I listed which I think aligns with yours where we'd do this automatically when `isolatedDeclarations` is set, and on typescript >=5.6.

DavidZbarsky-at: My vote would be to do it automatically, and if anyone complains, provide an opt-out mechanism for them. I think that's default people will want

jbedard: Then we just need the typescript >=5.6 check which I think we can do the exact same as `is_typescript_5_or_greater` today?

ts/private/ts_project.bzl
```diff
@@ -184,81 +184,110 @@ See https://github.com/aspect-build/rules_ts/issues/361 for more details.
...
-            if ctx.attr.declaration:
-                verb = "Transpiling & type-checking"
-            else:
-                verb = "Transpiling"
-        else:
-            verb = "Type-checking"
+        elif not use_tsc_for_dts:
+            # Not emitting declarations
+            # TODO: why doesn't this work with workers?
+            if not supports_workers:
```
gregmagolan, line 268: I'm in favor of dropping worker support in rules_ts 4.0. Not something we want to maintain and with async typecheck the need to keep large declaration graphs in memory for large typecheck actions is gone.

ts/private/ts_project.bzl
```diff
@@ -184,81 +184,110 @@ See https://github.com/aspect-build/rules_ts/issues/361 for more details.
     output_types = typings_outs + typing_maps_outs + typings_srcs
     use_tsc_noemit = ctx.attr.no_emit or (ctx.attr.transpile == 0 and not ctx.attr.declaration)
```
gregmagolan, line 185: this line can be removed since you're now setting this below

jbedard: Opps, that should have been dropped after my last refactor that was merged.

jbedard: Lets drop that before this PR: https://github.com/aspect-build/rules_ts/pull/701

ts/private/ts_project.bzl
DavidZbarsky-at: FWIW isolatedDeclarations is no panacea. You can still end up with typecheck actions that take 30-90s especially if they have a big transitive tree.

I know you're not keen to maintain the worker but Jason and I were talking about the possibility of resurrecting it in a simpler form (check only, no emit). We (@MichaelMitchell-at) might do some work on this if we have cycles.

Another possibility is even if the worker is not doing --watch mode, we have found 20%+ performance improvement just from caching previously-parsed ASTs in memory instead of parsing from scratch for each project.

I guess my point is I'd be a fan of keeping the current level of (non)-support for workers a bit longer while we do a bit of investigation into how they could potentially help (or not) with isolatedDeclarations.

ts/defs.bzl
DavidZbarsky-at: yes, the condition is slightly more complex but I agree with your assessment

ts/private/ts_lib.bzl
```diff
@@ -87,6 +87,18 @@ https://docs.aspect.build/rulesets/aspect_rules_js/docs/js_library#deps for more
         mandatory = True,
         allow_single_file = [".json"],
     ),
+    "isolated_typecheck": attr.bool(
+        doc = """\
+        Whether type-checking should be a separate action.
+
+        This allows the transpilation action to run without waiting for typings from dependencies.
+
+        Requires the typescript 5.6 [noCheck](https://www.typescriptlang.org/tsconfig#noCheck).
```
gregmagolan, line 96: "Requires a minimum version of typescript 5.6 for the [noCheck](https://www.typescriptlang.org/tsconfig#noCheck) flag which is automatically set on the transpilation action when the typecheck action is isolated."

ts/private/ts_lib.bzl
```diff
@@ -87,6 +87,18 @@ https://docs.aspect.build/rulesets/aspect_rules_js/docs/js_library#deps for more
...
+
+        Requires tsconfig [isolatedDeclarations](https://www.typescriptlang.org/tsconfig#isolatedDeclarations)
```
gregmagolan, line 98: "Requires [isolatedDeclarations](https://www.typescriptlang.org/tsconfig#isolatedDeclarations) to be set in tsconfig to so that declarations are emitted without dependencies."

gregmagolan: Also mention that this flag can cause compile errors because of <insert reasons> and that it is best to front-load that change before flipping this to True

ts/private/ts_project_options_validator.js
```diff
@@ -131,6 +131,17 @@ function main(_a) {
             )
         }
     }
+    function check_nocheck() {
+        if (attrs.isolated_typecheck) {
+            var optionVal = getTsOption('isolatedDeclarations')
+            if (!optionVal) {
+                failures.push(
+                    'attribute isolated_typecheck=True requires compilerOptions.isolatedDeclarations=true'
```
gregmagolan, line 139: also mention here that setting `compilerOptions.isolatedDeclarations=true` to True can cause compile errors because of <insert reasons> and that it is recommended to front-load that change separately from flipping `isolated_typecheck` to True

ts/private/ts_project.bzl
```diff
@@ -182,83 +182,111 @@ See https://github.com/aspect-build/rules_ts/issues/361 for more details.
...
+
+    # What tsc will be emitting
+    use_tsc_for_js = len(js_outs) > 0
+    use_tsc_for_dts = len(typings_outs) > 0
+
+    # Use a separate non-emitting action for type-checking when:
+    #  - a separate action was explicitly requested
+    #  or
+    #  - not invoking tsc for output files at all
+    use_tsc_noemit = ctx.attr.isolated_typecheck or not (use_tsc_for_js or use_tsc_for_dts)
```
gregmagolan, line 194: nit: may as well call this `use_isolated_typecheck`. `--noEmit` is implied on that action.

gregmagolan (approved):
🌮

ts/defs.bzl
gregmagolan: @jbedard and I chatted about flipping the default of `isolated_typecheck` to True if ts>5.6, however, it depends on the user having set `isolatedDeclarations` to true in their tsconfig, which we can't know at analysis time, so there is no way to set a reasonable default. Punting on the automatic opt-in for now.

jbedard: Note we decided on `isolated_typecheck` as the name for this param instead of the `async_typecheck` I originally had.

ts/private/ts_project.bzl
```diff
@@ -182,83 +182,111 @@ See https://github.com/aspect-build/rules_ts/issues/361 for more details.
...
-        outputs.append(typecheck_output)
-        stdout_file = typecheck_output.path
-
+        typecheck_arguments = ctx.actions.args()
         if supports_workers:
-            arguments.add("--bazelValidationFile", typecheck_output.short_path)
+            typecheck_arguments.add("--bazelValidationFile", typecheck_output.short_path)
 
-        arguments.add("--noEmit")
+        typecheck_arguments.add("--noEmit")
```
MichaelMitchell-at, line 229: When `support_workers` is `True` on line 108 we call `arguments.use_param_file`, in which case I think we'd want `"--noEmit"` to be added to that params file, rather than a command-line argument.

jbedard: Ohhh, you mean instead of `typecheck_arguments` it must be `arguments`?

MichaelMitchell-at: Yeah, though it seems tricky since `arguments` is also used down below where we wouldn't want it to have `--noEmit`. As far as I can tell there's not an easy way to create a copy of an `Args` object or remove an element. Something like this might be able to help with that: https://github.com/bazelbuild/bazel/issues/6230, but it looks stalled.

MichaelMitchell-at: Pretty dirty, but something like this seems to work
```diff
diff --git a/ts/private/ts_project.bzl b/ts/private/ts_project.bzl
index df88bc6..71c9ff3 100644
--- a/ts/private/ts_project.bzl
+++ b/ts/private/ts_project.bzl
@@ -74,7 +74,7 @@ def _ts_project_impl(ctx):
 
     typecheck_outs = []
 
-    arguments = ctx.actions.args()
+    argument_mods = []
     execution_requirements = {}
     executable = ctx.executable.tsc
 
@@ -105,34 +105,35 @@ See https://github.com/aspect-build/rules_ts/issues/361 for more details.
 
     if supports_workers:
         # Set to use a multiline param-file for worker mode
-        arguments.use_param_file("@%s", use_always = True)
-        arguments.set_param_file_format("multiline")
+        argument_mods.append(lambda arguments: arguments.use_param_file("@%s", use_always = True))
+        argument_mods.append(lambda arguments: arguments.set_param_file_format("multiline"))
         execution_requirements["supports-workers"] = "1"
         execution_requirements["worker-key-mnemonic"] = "TsProject"
         executable = ctx.executable.tsc_worker
 
     # Add all arguments from options first to allow users override them via args.
-    arguments.add_all(options.args)
+    argument_mods.append(lambda arguments: arguments.add_all(options.args))
 
     # Add user specified arguments *before* rule supplied arguments
-    arguments.add_all(ctx.attr.args)
+    argument_mods.append(lambda arguments: arguments.add_all(ctx.attr.args))
 
     outdir = _lib.join(
         ctx.label.workspace_root,
         _lib.join(ctx.label.package, ctx.attr.out_dir) if ctx.attr.out_dir else ctx.label.package,
     )
     tsconfig_path = to_output_relative_path(tsconfig)
-    arguments.add_all([
+    argument_mods.append(lambda arguments: arguments.add_all([
         "--project",
         tsconfig_path,
         "--outDir",
         outdir,
         "--rootDir",
         _lib.calculate_root_dir(ctx),
-    ])
+    ]))
+
     if len(typings_outs) > 0:
         declaration_dir = _lib.join(ctx.label.workspace_root, ctx.label.package, typings_out_dir)
-        arguments.add("--declarationDir", declaration_dir)
+        argument_mods.append(lambda arguments:  arguments.add("--declarationDir", declaration_dir))
 
     inputs = srcs_inputs + tsconfig_inputs
 
@@ -156,7 +157,7 @@ See https://github.com/aspect-build/rules_ts/issues/361 for more details.
 
     outputs = js_outs + map_outs + typings_outs + typing_maps_outs
     if ctx.outputs.buildinfo_out:
-        arguments.add("--tsBuildInfoFile", to_output_relative_path(ctx.outputs.buildinfo_out))
+        argument_mods.append(lambda arguments: arguments.add("--tsBuildInfoFile", to_output_relative_path(ctx.outputs.buildinfo_out)))
         outputs.append(ctx.outputs.buildinfo_out)
 
     output_sources = js_outs + map_outs + assets_outs
@@ -222,16 +223,19 @@ See https://github.com/aspect-build/rules_ts/issues/361 for more details.
         typecheck_output = ctx.actions.declare_file(ctx.attr.name + ".typecheck")
         typecheck_outs.append(typecheck_output)
 
-        typecheck_arguments = ctx.actions.args()
-        if supports_workers:
-            typecheck_arguments.add("--bazelValidationFile", typecheck_output.short_path)
+        arguments = ctx.actions.args()
+        for mod in argument_mods:
+            mod(arguments)
 
-        typecheck_arguments.add("--noEmit")
+        arguments.add("--noEmit")
+
+        if supports_workers:
+            arguments.add("--bazelValidationFile", typecheck_output.short_path)
 
         ctx.actions.run(
             executable = executable,
             inputs = transitive_inputs_depset,
-            arguments = [arguments, typecheck_arguments],
+            arguments = [arguments],
             outputs = [typecheck_output],
             mnemonic = "TsProjectCheck",
             execution_requirements = execution_requirements,
@@ -249,6 +253,10 @@ See https://github.com/aspect-build/rules_ts/issues/361 for more details.
         typecheck_outs.extend(output_types)
 
     if use_tsc_for_js or use_tsc_for_dts:
+        arguments = ctx.actions.args()
+        for mod in argument_mods:
+            mod(arguments)
+
         tsc_emit_arguments = ctx.actions.args()
 
         # Type-checking is done async as a separate action and can be skipped.
```

MichaelMitchell-at: ^ 
https://github.com/aspect-build/rules_ts/blob/4a4bf962b758cc5c754834c1b069c234178131df/ts/private/ts_project_worker.js#L997-L999

Notably, `--bazelValidationFile` is also expected to be in the params file

PR #681, opened 2024-09-04
feat: support --noEmit for type-checking as a validation action

### LINKED ISSUE #88: Typecheck-only rule

jbedard:
Add support for type-checking using validation actions when the ts `noEmit` option is used.

In the future this validation action can be used for other cases where tsc is only being used for type-checking.

fixes #88 

---

### Test plan

- New test cases added


ts/defs.bzl
```diff
@@ -309,14 +313,14 @@ def ts_project(
...
-    tsc_typing_maps_outs = _lib.calculate_typing_maps_outs(srcs, typings_out_dir, root_dir, declaration_map, allow_js)
+    tsc_typings_outs = _lib.calculate_typings_outs(srcs, typings_out_dir, root_dir, declaration, composite, allow_js, no_emit)
+    tsc_typing_maps_outs = _lib.calculate_typing_maps_outs(srcs, typings_out_dir, root_dir, declaration_map, allow_js, no_emit)
 
     tsc_js_outs = []
     tsc_map_outs = []
-    if not transpiler or transpiler == "tsc":
-        tsc_js_outs = _lib.calculate_js_outs(srcs, out_dir, root_dir, allow_js, resolve_json_module, preserve_jsx, emit_declaration_only)
-        tsc_map_outs = _lib.calculate_map_outs(srcs, out_dir, root_dir, source_map, preserve_jsx, emit_declaration_only)
+    if no_emit or not transpiler or transpiler == "tsc":
```
alexeagle, line 321: This logic change feels suspicious to me. What's the use case for going into this branch with noEmit? Is it correct to allow no_emit with a custom transpiler, or should we just `fail` in that case?

ts/private/ts_project.bzl
```diff
@@ -181,7 +181,7 @@ See https://github.com/aspect-build/rules_ts/issues/361 for more details.
 
     typings_srcs = [s for s in srcs_inputs if _lib.is_typings_src(s.path)]
 
-    if len(js_outs) + len(typings_outs) < 1:
+    if len(js_outs) + len(typings_outs) < 1 and not ctx.attr.no_emit:
```
alexeagle, line 184: the code inside this branch mentions issues/88 so it needs an update also

ts/private/ts_project.bzl
```diff
@@ -225,23 +225,53 @@ This is an error because Bazel does not run actions unless their outputs are nee
...
+    inputs_depset = depset(
+        copy_files_to_bin_actions(ctx, inputs),
+        transitive = transitive_inputs,
+    )
+
+    if ctx.attr.no_emit:
+        verb = "Type-checking"
+
+        # Validation actions still need to produce some output, so we output a .validation file
+        # that ends up in the _validation output group.
```
alexeagle, line 248: add to the comment that the content of that file will be any stdout written by `tsc`

ts/defs.bzl
jbedard: This was just copied from the other PR and I agree it's odd.

I think we can change this and then also revert the changes to `calculate_*_outs`.

ts/private/ts_project.bzl
```diff
@@ -225,23 +225,53 @@ This is an error because Bazel does not run actions unless their outputs are nee
...
+        validation_output = ctx.actions.declare_file(ctx.attr.name + ".validation")
+        validation_outs.append(validation_output)
+
+        if supports_workers:
+            arguments.add_all(["--bazelValidationFile", validation_output.short_path])
 
+        ctx.actions.run_shell(
+            command = "{} $@".format(executable.path),
```
alexeagle, line 256: this seems like a trivial one-liner command where this should be a `ctx.actions.run` rather than a `ctx.actions.run_shell`?

And, having made that change, I suspect you'll be able to avoid duplicating with the one below - you already set `verb` so all you need is to extract `env` to a variable so you can add one environment variable for this case

ts/private/ts_project_options_validator.js
```diff
@@ -131,25 +131,6 @@ function main(_a) {
...
-            'ERROR: ts_project rule ' +
-                target +
-                " cannot be built because the 'noEmit' option is specified in the tsconfig."
-        )
-        console.error(
-            'This is not compatible with ts_project, which always produces outputs.'
-        )
-        console.error(
-            '- If you mean to only typecheck the code, use the tsc_test rule instead.' +
-                '  (see https://github.com/aspect-build/rules_ts/tree/main/examples/typecheck_only)'
```
alexeagle, line 145: I don't see an update to that example in this PR?

ts/private/ts_project_worker.js
```diff
@@ -880,6 +880,9 @@ async function emit(request) {
     debug(`# Beginning new work`);
     debug(`arguments: ${request.arguments.join(' ')}`)
 
+    const validationPath = request.arguments[request.arguments.indexOf('--bazelValidationFile') + 1]
```
alexeagle, line 883: blah, too bad we still have to maintain this code :/

ts/test/mock_transpiler.bzl
```diff
@@ -53,7 +53,7 @@ def mock(name, srcs, source_map = False, **kwargs):
         srcs = srcs,
         # Calculate pre-declared outputs so they can be referenced as targets.
         # This is an optional transpiler feature aligning with the default tsc transpiler.
-        js_outs = lib.calculate_js_outs(srcs, ".", ".", False, False, False, False),
-        map_outs = lib.calculate_map_outs(srcs, ".", ".", True, False, False) if source_map else [],
+        js_outs = lib.calculate_js_outs(srcs, ".", ".", False, False, False, False, False),
```
alexeagle, line 56: these are getting gross, I think we should consider a post-factoring where we pass a struct to these helpers instead. (we could even pass the entire `ctx.attr` but feels violating law of demeter to me)

alexeagle (commented):
Nice, it's the right next step. I like the simplicity that `noEmit` == use validation action.

Note that I added a couple commits on the branch, so avoid force-push please

ts/defs.bzl
jbedard: `no_emit` + `transpiler` just means typecheck-only - a single rule with a validation action and no outputs (other then the validation action stdout file). I think that's valid?

ts/private/ts_project_worker.js
jbedard: Yeah :(

ts/private/ts_project.bzl
jbedard: We can also just delete this entire block now. I'm just not sure if we want to make such a change at this time? Currently this PR has is entirely opt-in with the `no_emit` flag...

ts/private/ts_project_options_validator.js
jbedard: Updated.

Note that I had to change the example to be a *passing* type-check and not failing, now that we can't use `expected_exit_code = 2`.

ts/private/ts_project.bzl
jbedard: Done 👍 

Note the 2 actions will be needed in #679 but I guess we don't need it yet so lets not add it yet.

ts/defs.bzl
jbedard: UPDATE: I think we should keep this as-is for now in this PR. Going into this `if` branch to only use tsc is easiest for now as it completely avoids changing the `else` branch for now - that `else` will change significantly when `isolated_declarations` is added so I'd like to wait until then.

alexeagle: okay great, thanks for investigating it

ts/private/ts_project.bzl
alexeagle: so then let's just be conservative and update the text, since issue 88 is being closed here

ts/private/ts_project_options_validator.js
alexeagle: yeah that's understood thx

ts/private/ts_project.bzl
```diff
@@ -269,6 +286,7 @@ This is an error because Bazel does not run actions unless their outputs are nee
             ),
             env = {
                 "BAZEL_BINDIR": ctx.bin_dir.path,
+                "JS_BINARY__STDOUT_OUTPUT_FILE": stdout_file,
```
alexeagle, line 289: nit: I'd feel safer if we didn't pass an environment variable with this key at all, rather than rely on an empty string value being interpreted the same as absence

ts/private/ts_project.bzl
jbedard: done

alexeagle (approved):

PR #684, opened 2024-09-07
perf: remove _replace_ext in lib.bzl

dzbarsky:
We don't need to look up the index of the `.` multiple times. Also flip the `allow_js` condition while I'm here, it's probably faster this way.

### Test plan

### DIFF
--- a/ts/private/ts_lib.bzl
+++ b/ts/private/ts_lib.bzl
@@ -180,11 +180,11 @@ def _is_typings_src(src):
     return src.endswith(".d.ts") or src.endswith(".d.mts") or src.endswith(".d.cts")
 
 def _is_js_src(src, allow_js, resolve_json_module):
-    if src.endswith(".js") or src.endswith(".jsx") or src.endswith(".mjs") or src.endswith(".cjs"):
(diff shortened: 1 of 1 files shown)


ts/private/ts_lib.bzl
```diff
@@ -180,8 +180,9 @@ def _is_typings_src(src):
     return src.endswith(".d.ts") or src.endswith(".d.mts") or src.endswith(".d.cts")
 
 def _is_js_src(src, allow_js, resolve_json_module):
-    if src.endswith(".js") or src.endswith(".jsx") or src.endswith(".mjs") or src.endswith(".cjs"):
-        return allow_js
+    if allow_js:
+        if src.endswith(".js") or src.endswith(".jsx") or src.endswith(".mjs") or src.endswith(".cjs"):
+            return True
 
     if src.endswith(".json"):
```
jbedard, line 187: Same change here just to be consistent?

dzbarsky: sure, done

jbedard (approved):

PR #687, opened 2024-09-08
perf: remove unnecessary tsconfig relative_file computation

jbedard:
The tsconfig is only generated within the BUILD directory and there is an assert that the `srcs` are all within that directory, so the relative file paths are easy to compute.

---

### Test plan

### DIFF
--- a/ts/private/ts_config.bzl
+++ b/ts/private/ts_config.bzl
@@ -137,10 +137,15 @@ def _write_tsconfig_rule(ctx):
 
     filtered_files = _filter_input_files(ctx.files.files, ctx.attr.allow_js, ctx.attr.resolve_json_module)
     if filtered_files:
-        content = content.replace(
-            "\"__files__\"",
-            str([relative_file(f.short_path, ctx.outputs.out.short_path) for f in filtered_files]),
-        )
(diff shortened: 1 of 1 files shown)


ts/private/ts_config.bzl
```diff
@@ -137,10 +137,14 @@ def _write_tsconfig_rule(ctx):
...
     filtered_files = _filter_input_files(ctx.files.files, ctx.attr.allow_js, ctx.attr.resolve_json_module)
     if filtered_files:
-        content = content.replace(
-            "\"__files__\"",
-            str([relative_file(f.short_path, ctx.outputs.out.short_path) for f in filtered_files]),
-        )
+        # Update file paths to be relative to the tsconfig file, including a ./ prefix
+        # to ensure paths are all relative to the config file.
+        filtered_files = [
+            "./" + f.short_path.removeprefix(ctx.label.package + "/")
```
DavidZbarsky-at, line 143: maybe pull `ctx.label.package + "/" out of the comprehension?

DavidZbarsky-at:
This saves us around 300ms

Before: 
image

After: 
image

thesayyn (approved):

PR #688, opened 2024-09-08
perf: remove unnecessary array create+join

jbedard:
---

### Test plan

- New test cases added


thesayyn (approved):
Is this a breaking change?

jbedard:
It's only the internal rule, the macro didn't change although I believe that is internal as well.

PR #690, opened 2024-09-09
refactor: change typechecking validation action to output group

jbedard:
This way tsc is not forced to always run as a validation action in the background and instead invoked only when the `*_typecheck` test is run.

Currently this only effects the `noEmit` case.

In the future this will allow the `isolatedDeclarations` fast-path to avoid running type-checking at all unless `bazel test` is run. This way something such as `ibazel run //devserver` will not be blocked waiting for the validation action to run in the background.

---

### Test plan

### DIFF
--- a/ts/defs.bzl
+++ b/ts/defs.bzl
@@ -327,8 +327,6 @@ def ts_project(
         # we have to produce a few targets
         tsc_target_name = "%s_typings" % name
         transpile_target_name = "%s_transpile" % name
-        typecheck_target_name = "%s_typecheck" % name
-        test_target_name = "%s_typecheck_test" % name
 
         if type(transpiler) == "function" or type(transpiler) == "rule":
             transpiler(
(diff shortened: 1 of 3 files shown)


ts/defs.bzl
```diff
@@ -350,8 +350,7 @@ def ts_project(
         native.filegroup(
             name = typecheck_target_name,
             srcs = [tsc_target_name],
-            # This causes the types to be produced, which in turn triggers the tsc action to typecheck
-            output_group = "types",
+            output_group = "typecheck",
```
jbedard, line 353: This could be considered "breaking" if people are using `[name]_typecheck` to get the dts files :/

dzbarsky: Seems a little fringe, maybe we add it to release notes?

ts/defs.bzl
```diff
@@ -346,12 +344,37 @@ def ts_project(
         else:
             fail("transpiler attribute should be a rule/macro or a skylib partial. Got " + type(transpiler))
 
+        # Default target produced by the macro gives the js and map outs, with the transitive dependencies.
+        js_library(
+            name = name,
+            # Include the tsc target in srcs to pick-up both the direct & transitive declaration outputs so
```
dzbarsky, line 350: Why srcs and not deps? I guess I'm not super clear on when we use which one, it's a little confusing because we collect from both

ts/private/ts_project.bzl
```diff
@@ -241,18 +243,20 @@ This is an error because Bazel does not run actions unless their outputs are nee
...
+        typecheck_outs.append(typecheck_output)
 
-        outputs.append(validation_output)
-        stdout_file = validation_output.path
+        outputs.append(typecheck_output)
+        stdout_file = typecheck_output.path
 
         if supports_workers:
-            arguments.add_all(["--bazelValidationFile", validation_output.short_path])
+            arguments.add_all(["--bazelValidationFile", typecheck_output.short_path])
```
dzbarsky, line 255: Not: you can use the 2-param overload of add, no need to construct an array and call add_all

dzbarsky (approved):

ts/defs.bzl
dzbarsky: I see it's preexisting, so just wondering

ts/private/ts_project.bzl
dzbarsky: Nit*

ts/defs.bzl
jbedard: I think because both js and dts files are the primary output of this target, we don't want to consider them as transitive deps.

gregmagolan (approved):
LGTM

PR #697, opened 2024-09-19
feat: support custom dts transpiler

jbedard:
Support a custom transpiler for producing dts files.

Overview of targets created:
* `{name}` outputs a `JsInfo` containing all transpiled files, no guarantee if this runs tsc, no guarantee it outputs anything such as `no_emit = True`

If any transpiler or no_emit is used then the following are also created:
* `{name}_types` if `declaration: True && not no_emit`: dts files in `DefaultInfo`
* `{name}_tsc` invokes tsc, maybe outputting something, maybe doing type-checking, maybe both
* `{name}_typecheck` outputs something when type-checking passes
* `{name}_typecheck_test` forces `{name}_typecheck` to build and therefor forces type-checking

---

A custom dts transpiler can now be set using `ts_project(declaration_transpiler)`. This aligns dts transpiling with js transpiling allowing fast transpiling of all files, only requiring invoking tsc for a type-checking action.

### Test plan

- New test cases added


ts/defs.bzl
```diff
@@ -167,6 +173,12 @@ def ts_project(
 
             See [docs/transpiler.md](/docs/transpiler.md) for more details.
 
+        declaration_transpiler: A custom transpiler tool to run that produces the TypeScript declaration outputs instead of `tsc`.
```
jbedard, line 176: TODO: names, should be similar to the existing `transpiler` attribute:
* declaration_transpiler
* dts_transpiler
* types_transpiler
* ?

jbedard:
TODO: tests, I think swc isn't ready (https://github.com/swc-project/swc/issues/9512) so we might have to try something else, even just a `js_binary` hiding plain old tsc

ts/private/ts_project.bzl
```diff
@@ -215,81 +215,113 @@ This is an error because Bazel does not run actions unless their outputs are nee
...
 
-        if stdout_file != "":
-            env["JS_BINARY__STDOUT_OUTPUT_FILE"] = stdout_file
+        inputs_depset = inputs if ctx.attr.async_typecheck else transitive_inputs_depset
 
         ctx.actions.run(
             executable = executable,
             inputs = inputs_depset,
-            arguments = [arguments],
+            arguments = [arguments, tsc_emit_arguments],
```
MichaelMitchell-at, line 312: `--emitDeclarationOnly` can appear twice since both `arguments` and `tsc_emit_arguments` can contain `--emitDeclarationOnly`

ts/defs.bzl
```diff
@@ -313,19 +325,43 @@ def ts_project(
...
+    tsc_target_name = name
+    declarations_target_name = None
+    transpile_target_name = None
+
+    # typing outputs
+    tsc_typings_outs = []
+    tsc_typing_maps_outs = []
+    if not declaration_transpiler or declaration_transpiler == "tsc":
+        tsc_typings_outs = _lib.calculate_typings_outs(srcs, typings_out_dir, root_dir, declaration, composite, allow_js, no_emit)
+        tsc_typing_maps_outs = _lib.calculate_typing_maps_outs(srcs, typings_out_dir, root_dir, declaration_map, allow_js, no_emit)
```
MichaelMitchell-at, line 338: I think these are not actually used since they are passed to `ts_project_rule`, however `ts_project_rule` never reads from `ctx.attr.typings_outs` nor `ctx.attr.typing_maps_outs`. It recalculates the outs itself.

ts/private/ts_project.bzl
```diff
@@ -215,81 +215,113 @@ This is an error because Bazel does not run actions unless their outputs are nee
 
     output_types = typings_outs + typing_maps_outs + typings_srcs
 
+    # What tsc will be emitting
+    use_tsc_for_js = len(js_outs) > 0
+    use_tsc_for_dts = len(typings_outs) > 0
```
MichaelMitchell-at, line 220: Due to the reason I mentioned in my comment above, this will be incorrect (non-empty) when using a custom `declaration_transpiler`.

ts/private/ts_project.bzl
MichaelMitchell-at, line 1: (Github doesn't let me comment on unchanged lines)

The logic starting at line 180 will need to account for both `js_outs` and `typings_outs` being empty when both  `transpiler` and `declaration_transpiler` are custom.

MichaelMitchell-at (changes requested):
I tested this out on our repo and I ran into some issues. I think if my comments are addressed we'll be good though!

ts/private/ts_project.bzl
MichaelMitchell-at: I think you can ignore this. Seeing `--emitDeclarationOnly` twice is probably a side-effect of the other issue I commented on.

ts/defs.bzl
jbedard: It's done in the macro here just to predeclare all possible, so the outputs can be referenced as labels.

Then within the rule they are ignored and recalculated like you mentioned, then things like targets are expanded to a full list of files which could not be pre-calculated.

ts/private/ts_project.bzl
jbedard: Yeah, you're correct with this as well.

I think in this case (both transpilers are custom) it can be the same as `no_emit = True` where it will just be a single action with `--noEmit` that outputs the `typecheck_output` file to the `typecheck` output group. Then I think we can actually remove the full block starting at line 180, but I'd like that to be a followup...

ts/private/ts_project.bzl
jbedard: I'll verify that case more when I add proper tests for this 👍

ts/private/ts_project.bzl
jbedard: See https://github.com/aspect-build/rules_ts/pull/698/commits/eb3dbb7dead69f4225eac485ebb03ffb079e9766 which adds support for any case where no outptus are produced (in addition to `--noEmit` which is already supported.

That's what we should do imo...

ts/private/ts_project.bzl
jbedard: I think this was fixed in #679

ts/private/ts_project.bzl
jbedard: That logic (which was starting at line 180) has been dropped now

ts/private/ts_project.bzl
MichaelMitchell-at: I think this is the last thing left. I added this patch on top of the latest version of this branch and everything seems to be working.
```diff
diff --git a/ts/defs.bzl b/ts/defs.bzl
index 63f885d..875f768 100644
--- a/ts/defs.bzl
+++ b/ts/defs.bzl
@@ -481,6 +481,7 @@ def ts_project(
         tsc = tsc,
         tsc_worker = tsc_worker,
         transpile = -1 if not transpiler else int(transpiler == "tsc"),
+        declaration_transpile = -1 if not declaration_transpiler else int(declaration_transpiler == "tsc"),
         supports_workers = supports_workers,
         is_typescript_5_or_greater = is_typescript_5_or_greater,
         validate = validate,
diff --git a/ts/private/ts_lib.bzl b/ts/private/ts_lib.bzl
index 99cab36..1d1fed4 100644
--- a/ts/private/ts_lib.bzl
+++ b/ts/private/ts_lib.bzl
@@ -70,6 +70,18 @@ https://docs.aspect.build/rulesets/aspect_rules_js/docs/js_library#deps for more
         default = -1,
         values = [-1, 0, 1],
     ),
+    "declaration_transpile": attr.int(
+        doc = """\
+        Whether tsc should be used to produce .d.ts outputs
+
+        Values are:
+        - -1: Error if --@aspect_rules_ts//ts:default_to_tsc_transpiler not set, otherwise transpile
+        - 0: Do not transpile
+        - 1: Transpile
+        """,
+        default = -1,
+        values = [-1, 0, 1],
+    ),
     "tsc": attr.label(
         doc = "TypeScript compiler binary",
         mandatory = True,
diff --git a/ts/private/ts_project.bzl b/ts/private/ts_project.bzl
index 0d73be7..c95c3ec 100644
--- a/ts/private/ts_project.bzl
+++ b/ts/private/ts_project.bzl
@@ -64,8 +64,8 @@ def _ts_project_impl(ctx):
     typings_out_dir = ctx.attr.declaration_dir or ctx.attr.out_dir
     js_outs = _lib.declare_outputs(ctx, [] if ctx.attr.transpile == 0 else _lib.calculate_js_outs(srcs, ctx.attr.out_dir, ctx.attr.root_dir, ctx.attr.allow_js, ctx.attr.resolve_json_module, ctx.attr.preserve_jsx, ctx.attr.no_emit, ctx.attr.emit_declaration_only))
     map_outs = _lib.declare_outputs(ctx, [] if ctx.attr.transpile == 0 else _lib.calculate_map_outs(srcs, ctx.attr.out_dir, ctx.attr.root_dir, ctx.attr.source_map, ctx.attr.preserve_jsx, ctx.attr.no_emit, ctx.attr.emit_declaration_only))
-    typings_outs = _lib.declare_outputs(ctx, _lib.calculate_typings_outs(srcs, typings_out_dir, ctx.attr.root_dir, ctx.attr.declaration, ctx.attr.composite, ctx.attr.allow_js, ctx.attr.no_emit))
-    typing_maps_outs = _lib.declare_outputs(ctx, _lib.calculate_typing_maps_outs(srcs, typings_out_dir, ctx.attr.root_dir, ctx.attr.declaration_map, ctx.attr.allow_js, ctx.attr.no_emit))
+    typings_outs = _lib.declare_outputs(ctx, [] if ctx.attr.declaration_transpile == 0 else _lib.calculate_typings_outs(srcs, typings_out_dir, ctx.attr.root_dir, ctx.attr.declaration, ctx.attr.composite, ctx.attr.allow_js, ctx.attr.no_emit))
+    typing_maps_outs = _lib.declare_outputs(ctx, [] if ctx.attr.declaration_transpile == 0 else _lib.calculate_typing_maps_outs(srcs, typings_out_dir, ctx.attr.root_dir, ctx.attr.declaration_map, ctx.attr.allow_js, ctx.attr.no_emit))
 
     validation_outs = []
     if ctx.attr.validate:
@@ -181,6 +181,10 @@ See https://github.com/aspect-build/rules_ts/issues/361 for more details.
     if len(outputs) > 0 and ctx.attr.transpile == -1 and not ctx.attr.emit_declaration_only and not options.default_to_tsc_transpiler:
         fail(transpiler_selection_required)
 
+    # TODO: Determine if this is necessary
+    # if len(outputs) > 0 and ctx.attr.declaration_transpile == -1 and ctx.attr.emit_declaration_only and not options.default_to_tsc_transpiler:
+    #     fail(declaration_transpiler_selection_required)
+
     output_types = typings_outs + typing_maps_outs + typings_srcs
     use_tsc_noemit = ctx.attr.no_emit or (ctx.attr.transpile == 0 and not ctx.attr.declaration)
```

jbedard: That looks about right to me. I think we can just use a boolean though, or even just set `no_emit` from the macro.

I don't think we want the `fail` though, at least for now. That is more important for .js and I don't think we have a recommended dts transpiler at this time like we do with the js transpiler.

MichaelMitchell-at (approved):
Tested it out on our code base. Works great!

ts/defs.bzl
jbedard: Decided on `declaration_transpiler`

ts/private/ts_lib.bzl
```diff
@@ -82,6 +82,9 @@ https://docs.aspect.build/rulesets/aspect_rules_js/docs/js_library#deps for more
         executable = True,
         cfg = "exec",
     ),
+    "declaration_transpile": attr.bool(
```
jbedard, line 85: Note that this is different then the macro `declaration_transpiler`, this is just a flag indicating if the declarations are generated outside the rule so the rule doesn't have to do anything.

ts/defs.bzl
```diff
@@ -172,6 +173,12 @@ def ts_project(
 
             See [docs/transpiler.md](/docs/transpiler.md) for more details.
 
+        declaration_transpiler: A custom transpiler tool to run that produces the TypeScript declaration outputs instead of `tsc`.
+
```
gregmagolan, line 177: Mention that if set, it should be a rule or macro with the signature `(name, srcs, **kwargs)`

ts/defs.bzl
```diff
@@ -318,19 +325,43 @@ def ts_project(
...
-    tsc_typing_maps_outs = _lib.calculate_typing_maps_outs(srcs, typings_out_dir, root_dir, declaration_map, allow_js, no_emit)
+    # Target names for tsc, dts+js transpilers
+    tsc_target_name = name
+    declarations_target_name = None
+    transpile_target_name = None
+
+    # typing outputs
+    tsc_typings_outs = []
+    tsc_typing_maps_outs = []
+    if not declaration_transpiler or declaration_transpiler == "tsc":
```
gregmagolan, line 336: Don't need to have the explicit `tsc` option for `declaration_transpiler` we do for `transpiler`.

```
if not declaration_transpiler:
```

should be sufficient since we're not forcing users to either set `--@aspect_rules_ts//ts:default_to_tsc_transpiler` or an explicit `tsc` string.

ts/defs.bzl
```diff
@@ -172,6 +173,12 @@ def ts_project(
 
             See [docs/transpiler.md](/docs/transpiler.md) for more details.
 
+        declaration_transpiler: A custom transpiler tool to run that produces the TypeScript declaration outputs instead of `tsc`.
+
+            This is only used when `transpiler` is set to a custom transpiler tool.
```
gregmagolan, line 178: Is this true? I don't see it asserted. A user could use tsc for transpilation but configure it to not emit declarations and use a custom declaration_transpiler right?

ts/defs.bzl
```diff
@@ -318,19 +325,43 @@ def ts_project(
...
+    declarations_target_name = None
+    transpile_target_name = None
+
+    # typing outputs
+    tsc_typings_outs = []
+    tsc_typing_maps_outs = []
+    if not declaration_transpiler or declaration_transpiler == "tsc":
+        tsc_typings_outs = _lib.calculate_typings_outs(srcs, typings_out_dir, root_dir, declaration, composite, allow_js, no_emit)
+        tsc_typing_maps_outs = _lib.calculate_typing_maps_outs(srcs, typings_out_dir, root_dir, declaration_map, allow_js, no_emit)
+    else:
```
gregmagolan, line 339: If `declaration_transpiler` is set and `transpiler` is `tsc`, should we assert that `declaration` is `False` and fail if it is not. We don't want the tsc "transpiler" also outputting declarations and trying to overwrite on the ones from the `declaration_transpiler` if not in the sandbox.

ts/defs.bzl
jbedard: That sounds right, I'll remove the magic "tsc" on this line. I'm not sure why I originally did that (this PR has been through a few iterations now...).

ts/defs.bzl
jbedard: Hmm, I don't think so... I think `declaration && declaration_transpiler != None` means we should invoke the `declaration_transpiler`, but add `--declaration false` when invoking tsc.

Just like when using `transpiler` we add `--emitDeclarationOnly` when invoking tsc.

ts/defs.bzl
jbedard: I think you're correct. This is not true, I originally thought it would be but I think this full PR turned out a bit simpler then I originally thought 👍

jbedard:
Updated. See the outline of targets I put into the PR description.

IMO the `{name}_types` target should *always* be created, even if we are using `tsc` for everything (js + dts) this `{name}_types` target extracts the dts files into `DefaultInfo`. I don't see any reason why that should only be set when using a transpiler?

ts/defs.bzl
jbedard: I've updated this...

`declaration: True` means declarations will be outputted... maybe from tsc, maybe from a `declaration_transpiler`

The `--emitDeclarationOnly`, `--declaration false` and `--noEmit` might be added to the `tsc` invocation depending on the transpiler+noemit rule attributes.

docs/transpiler.md
```diff
@@ -111,12 +111,13 @@ then `load()` from your own macro rather than from `@aspect_rules_ts`.
 
 ## Macro expansion
 
-When `transpiler` is used, then the `ts_project` macro expands to these targets:
+When `transpiler` or `declaration_transpiler` is used, then the `ts_project` macro expands to these targets:
```
jbedard, line 114: @gregmagolan I updated this doc since the last review

ts/defs.bzl
```diff
@@ -318,19 +325,43 @@ def ts_project(
...
+    declarations_target_name = None
+    transpile_target_name = None
+
+    # typing outputs
+    tsc_typings_outs = []
+    tsc_typing_maps_outs = []
+    if no_emit or not declaration_transpiler:
+        tsc_typings_outs = _lib.calculate_typings_outs(srcs, typings_out_dir, root_dir, declaration, composite, allow_js, no_emit)
+        tsc_typing_maps_outs = _lib.calculate_typing_maps_outs(srcs, typings_out_dir, root_dir, declaration_map, allow_js, no_emit)
+    elif not no_emit and declaration:
```
jbedard, line 339: Note that we declare no `declarations_target_name` if we don't need the `declaration_transpiler` to output any dts files.

docs/transpiler.md
gregmagolan: Reading the code I think there are cases when the macro does not expand to all of the targets tho the exact matrix is hard to follow. When no_emit is True is one of those cases I see.

```
if no_emit or not declaration_transpiler:
   ...
elif not no_emit and declaration:
   ...
```

is particularly confusing

ts/defs.bzl
gregmagolan: I think `elif declaration:` should be sufficient since if `no_emit` is True you'll fall into the first condition

ts/defs.bzl
```diff
@@ -172,6 +173,12 @@ def ts_project(
 
             See [docs/transpiler.md](/docs/transpiler.md) for more details.
 
+        declaration_transpiler: A custom transpiler tool to run that produces the TypeScript declaration outputs instead of `tsc`.
```
gregmagolan, line 176: It looks like `declaration_transpiler` is ignored if `declaration` is False?

If so, should mention that in this docstring and also `fail` on that condition to help users out since as it will be an easy thing to miss

docs/transpiler.md
gregmagolan: Speaking of no_emit, that flag is particularly confusing if there is a custom `transpiler` or `declaration_transpiler` set since the docstring for that parameter is no longer accurate as Bazel is expecting outputs, just not from the `{name}_tsc` target.

gregmagolan: It looks like `transpiler` and `declaration_transpiler` are ignored if `no_emit` is True so should probably help the user out and `fail` on that condition.

ts/defs.bzl
```diff
@@ -439,6 +494,7 @@ def ts_project(
         tsc = tsc,
         tsc_worker = tsc_worker,
         transpile = -1 if not transpiler else int(transpiler == "tsc"),
+        declaration_transpile = declaration_transpiler != None,
```
gregmagolan, line 497: should be `declarations_target_name != None` since there are cases where `declaration_transpiler` is ignored

ts/defs.bzl
```diff
@@ -439,6 +494,7 @@ def ts_project(
         tsc = tsc,
         tsc_worker = tsc_worker,
         transpile = -1 if not transpiler else int(transpiler == "tsc"),
```
gregmagolan, line 496: Here also, `transpiler` can be ignored if `emit_declaration_only` is True so more correct would be
```
0 if transpile_target_name != None else (1 if transpiler == "tsc" else -1)
```

gregmagolan: And then as a defence in depth, also fail with a friendly error message if `emit_declaration_only` is True and user sets `transpiler` since that case is not valid

ts/defs.bzl
gregmagolan: Actually, maybe this should be
```
elif declaration or emit_declaration_only:
```
if it is valid to have `declaration` False and `emit_declaration_only` True?

jbedard: > if it is valid to have declaration False and emit_declaration_only True?

tsc fails in that case
```
error TS5069: Option 'emitDeclarationOnly' cannot be specified without specifying option 'declaration' or option 'composite'.
```
... then if trying to test that `composite` scenario:
```
examples/transpiler/tsconfig_custom_transpilers-emit_declaration_only.json(1,54): error TS6304: Composite projects may not disable declaration emit.
```
... because we are using `tsc --noEmit` due to the `declaration_transpiler` 🤯

ts/defs.bzl
jbedard: Actually I think I'm going to leave this (and the above line) as-is.

These variables just tell the rule if alternate transpilers exist, not if the current tsconfig will vs will not use them.

docs/transpiler.md
```diff
@@ -111,16 +111,16 @@ then `load()` from your own macro rather than from `@aspect_rules_ts`.
...
```
gregmagolan, line 114: Last nit here. Clarify that `transpiler` and `declaration_transpiler` may be set but _not_ used in some cases, such as when `no_emit` is set. In those cases, the macro will not expand to these targets.

ts/defs.bzl
gregmagolan, line 176: nit: clarify that there are cases where `declaration_transpiler` is set but not used if the rule is configured to not output declarations, such as when `declaration` is False or when `no_emit` is True.

gregmagolan: may as well clarify the same thing for `transpiler` that it is not used when `no_emit` is True or `emit_declarations_only` is True

gregmagolan (approved):

docs/transpiler.md
gregmagolan, line 114: nit: `When 'transpiler' or 'declaration_transpiler' is set` is more accurate since from reading the code the macro expansion is when they are set and not necessarily if they are used
nit: looks like if `no_emit` is set the macro will also expand to these targets (minus the `[name]_types` target

PR #698, opened 2024-09-20
feat: support ts_project with transpiler and no declarations for type-checking

jbedard:
Now that `--noEmit` can be used to typecheck without any output files it can also be used when a custom transpiler is used along with no dts outputs.

---

### Test plan

- New test cases added

### DIFF
--- a/examples/no_emit/BUILD.bazel
+++ b/examples/no_emit/BUILD.bazel
@@ -1,4 +1,5 @@
 load("@aspect_rules_ts//ts:defs.bzl", "ts_project")
+load("@bazel_skylib//rules:build_test.bzl", "build_test")
 load("@bazel_skylib//rules:write_file.bzl", "write_file")
 
 write_file(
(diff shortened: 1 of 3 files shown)


ts/private/ts_lib.bzl
```diff
@@ -87,6 +87,18 @@ https://docs.aspect.build/rulesets/aspect_rules_js/docs/js_library#deps for more
         mandatory = True,
         allow_single_file = [".json"],
     ),
+    "async_typecheck": attr.bool(
```
DavidZbarsky-at, line 90: I'm not in love with this name, though we will stick it in a macro anyway so it's not a huge deal :)

For us we did a single flag `enforce_isolated_declarations` that enabled this mode + isoDecls in the tsconfig. Do you think there's any case where someone would set isoDecls in tsconfigs but not want to enable this?

DavidZbarsky-at (approved):

ts/private/ts_lib.bzl
jbedard: Lets move this discussion here: https://github.com/aspect-build/rules_ts/pull/679#discussion_r1759506820

dzbarsky (approved):
Only skimmed the example, rule change lgtm

examples/no_emit/BUILD.bazel
```diff
@@ -9,12 +10,46 @@ write_file(
...
+write_file(
+    name = "gen_js",
+    out = "b.js",
+    content = [
+        "export const b = 43",
+    ],
+)
+
+# Shows how to run `tsc` producing no outputs at all via no_emit=True.
 # It will use a validation action to type-check the file:
```
gregmagolan, line 22: No longer a validation action after #690 right?

examples/no_emit/BUILD.bazel
```diff
@@ -9,12 +10,46 @@ write_file(
...
 # https://bazel.build/extending/rules#validation_actions
 # Run bazel with --norun_validations to skip type-checking.
```
gregmagolan, line 24: Setting --norun_validations is no longer correct either after #690 right?

examples/transpiler/BUILD.bazel
```diff
@@ -31,7 +40,7 @@ ts_project(
 )
 
 # Runs babel to transpile ts -> js
-# and does not produce any declaration outputs.
+# and does not produce any declaration outputs due to noEmit=True.
 # `tsc` is used for type-check only, as a validation action
 # (run bazel with --norun_validations to skip typechecking)
```
gregmagolan, line 45: Setting --norun_validations is no longer correct either after https://github.com/aspect-build/rules_ts/pull/690 right?

examples/no_emit/BUILD.bazel
```diff
@@ -9,12 +10,46 @@ write_file(
...
     srcs = ["a.ts"],
-    no_emit = True,
+    tsconfig = {
+        "compilerOptions": {
+            "noEmit": True,
+        },
+    },
+)
+
+# Shows how to run `tsc` producing no outputs at all via declarations=False.
```
gregmagolan, line 35: Comment wrong? I don't see `declarations=False` set.

gregmagolan (approved):
lgtm

examples/no_emit/BUILD.bazel
jbedard: Correct, I'll reword a few of these

PR #706, opened 2024-09-30
Support TypeScript workers with isolated declarations and custom TypeScript

MichaelMitchell-at:
I've successfully tested this on Airtable's code base. Memory usage can be a problem when typechecking a lot of targets. Something like what @alexeagle requested in https://github.com/bazelbuild/bazel/issues/12165#issuecomment-769207271 might be useful. I'm still playing around with `--experimental_total_worker_memory_limit_mb` and `--experimental_shrink_worker_pool` to see if they can prevent OOMs.

Also, I don't know why, but I have not encountered https://github.com/aspect-build/rules_ts/issues/361 and I don't believe any changes in this PR would have done anything to fix that issue.

### DIFF
--- a/ts/private/ts_project.bzl
+++ b/ts/private/ts_project.bzl
@@ -84,13 +84,6 @@ def _ts_project_impl(ctx):
 
     typecheck_outs = []
 
-    # In worker mode we need to pass in a single Args object to each action with use_param_file set.
-    # Since the arguments to the TsProjectCheck and TsProject/TsProjectEmit actions are slightly
-    # different we need to have separate Args objects for each action. Since they share many of the
-    # same arguments, but copying an Args object isn't allowed, we lazily accumulate the modifications
-    # and apply them to each separately constructed Args object later.
-    argument_mods = []
-
     execution_requirements = {}
     executable = ctx.executable.tsc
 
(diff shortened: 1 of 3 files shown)


ts/private/ts_project.bzl
```diff
@@ -84,7 +84,13 @@ def _ts_project_impl(ctx):
 
     typecheck_outs = []
 
-    arguments = ctx.actions.args()
+    # In worker mode we need to pass in a single Args object to each action with use_param_file set.
```
jbedard, line 87: Is there a reason you think lambdas is better then just an array of args?

ts/private/ts_project.bzl
```diff
@@ -232,16 +239,20 @@ See https://github.com/aspect-build/rules_ts/issues/361 for more details.
...
+        arguments = ctx.actions.args()
+        for mod in argument_mods:
+            mod(arguments)
+
+        arguments.add("--noEmit")
 
-        typecheck_arguments.add("--noEmit")
+        worker_arguments = ctx.actions.args()
+        if supports_workers:
+            arguments.add("--bazelValidationFile", typecheck_output.short_path)
```
jbedard, line 250: Should this be `worker_arguments`?

ts/private/ts_project.bzl
```diff
@@ -259,23 +270,25 @@ See https://github.com/aspect-build/rules_ts/issues/361 for more details.
...
+            arguments.add("--noResolve")
 
         if not use_tsc_for_js:
             # Not emitting js
-            tsc_emit_arguments.add("--emitDeclarationOnly")
+            arguments.add("--emitDeclarationOnly")
 
         elif not use_tsc_for_dts:
             # Not emitting declarations
             # TODO: why doesn't this work with workers?
```
jbedard, line 289: Does use of a single `ctx.actions.args()` fix this?

MichaelMitchell-at: Works on my machine :)

MichaelMitchell-at:
@jbedard updated!

ts/private/ts_project.bzl
```diff
@@ -232,16 +231,26 @@ See https://github.com/aspect-build/rules_ts/issues/361 for more details.
...
+        env = {
+            "BAZEL_BINDIR": ctx.bin_dir.path,
+        }
+
+        if supports_workers:
+            args.use_param_file("@%s", use_always = True)
+            args.set_param_file_format("multiline")
+            args.add("--bazelValidationFile", typecheck_output.short_path)
+        else:
+            env["JS_BINARY__STDOUT_OUTPUT_FILE"] = typecheck_output.path
```
MichaelMitchell-at, line 248: It took me several hours to finally narrow down to this as the root cause of a bug where running a command with workers enabled was just hanging. The persistent worker protocol uses stdout to communicate with the worker process, but setting this env variable makes stdout of the worker process get captured into a temp file by the `js_binary` wrapper script:
https://github.com/aspect-build/rules_js/blob/f583319639a7875b891ab35701a3d0e630d4f8f3/js/private/js_binary.sh.tpl#L21-L23
So the Bazel process was never receiving the message that the worker was done with the job.
The purpose of capturing stdout was to create the marker file, so when running in worker mode we instead rely on the `--bazelValidationFile` flag (see handling in ts_project_worker.js).

Interesting to note is that the logic was originally structured similarly in the original validation action PR:
https://github.com/aspect-build/rules_ts/pull/574/files#r1744177502

ts/private/ts_project_worker.js
```diff
@@ -467,11 +467,6 @@ function createFilesystemTree(root, inputs) {
 
 
 /** Program and Caching */
-function isExternalLib(path) {
-    return path.includes('external') &&
```
MichaelMitchell-at, line 471: This condition fails if using a custom version of TypeScript. Updated the logic in this file with a different way of checking this.

ts/private/ts_project.bzl
jbedard: Interesting!

Can we maybe add a comment here explaining that a little bit? Even just essentially "The persistent worker protocol uses stdout to communicate with the worker process so stdout can only be redirected in non-worker mode"

ts/private/ts_project.bzl
```diff
@@ -232,16 +231,26 @@ See https://github.com/aspect-build/rules_ts/issues/361 for more details.
         typecheck_output = ctx.actions.declare_file(ctx.attr.name + ".typecheck")
         typecheck_outs.append(typecheck_output)
 
-        typecheck_arguments = ctx.actions.args()
```
jbedard, line 235: Can we keep the existing names just to reduce the diff in this PR?

jbedard (approved):

PR #708, opened 2024-10-06
Add fixture for declaration transpiler test

MichaelMitchell-at:
I haven't quite figured out how or what to test, but I figure adding a dummy transpiler based on what we have at Airtable could be helpful.

---

### Test plan

### DIFF
--- a/ts/test/mock_transpiler.bzl
+++ b/ts/test/mock_transpiler.bzl
@@ -1,5 +1,7 @@
 "Fixture to demonstrate a custom transpiler for ts_project"
 
+load("@aspect_rules_js//js:libs.bzl", "js_lib_helpers")
+load("@aspect_rules_js//js:providers.bzl", "JsInfo")
 load("@aspect_rules_ts//ts/private:ts_lib.bzl", "lib")
 
 _DUMMY_SOURCEMAP = """{"version":3,"sources":["%s"],"mappings":"AAAO,KAAK,CAAC","file":"in.js","sourcesContent":["fake"]}"""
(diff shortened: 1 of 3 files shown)


ts/test/mock_transpiler.bzl
```diff
@@ -21,15 +23,38 @@ def _mock_impl(ctx):
...
-            arguments = [src.path, js_file.path.replace(".ts", ".js")],
+            arguments = [src.path, js_file.path],
         )
 
         ctx.actions.write(
             output = map_file,
             content = _DUMMY_SOURCEMAP % src.short_path,
         )
 
-    return DefaultInfo(files = depset(out_files))
```
jbedard, line 32: Why did this have to switch? The transpiler should just have to output the transpiled js or dts (+maps) as default info?

MichaelMitchell-at: Having `JsInfo` seems typical of actual usage. Some examples:
- https://github.com/aspect-build/rules_swc/blob/9498a51454efc51ec5c2da9bac9fad610a3f4eab/swc/private/swc.bzl#L362-L375
- https://github.com/aspect-build/rules_esbuild/blob/49510f4eaaab95a5c637dbdacc9f7bc9c80406fd/esbuild/private/esbuild.bzl#L446-L458

jbedard: That just means the rule can be used directly for things like `js_library(deps)` I guess, instead of how `ts_project` (the macro) or `js_library(srcs)` uses the `DefaultInfo`.

I guess it does not harm, but also isn't used in these tests 🤷

dzbarsky: Yeah it's probably worth doing in case folks crib this as an example (though we could also provide a production-ready transpiler using oxc as an example ;)

PR #710, opened 2024-10-07
fix: always declare ts_project_rule as primary ts_project target

jbedard:
This makes the `ts_project` macro produce more consistent underlying targets. The `{name}` target will now be a `ts_project_rule` 100% of the time. The only conditional logic is the the same as today: the addition of `{name}_types`, `{name}_typecheck`, `{name}_typecheck_test`.

We can maybe think of adding those 3 extra targets 100% of the time in the future (or maybe only doing `{name}_types` 100% of the time and adding the test only when `{name}` will not run tsc?).

---

### Test plan


ts/private/ts_lib.bzl
```diff
@@ -70,6 +70,12 @@ https://docs.aspect.build/rulesets/aspect_rules_js/docs/js_library#deps for more
         default = -1,
         values = [-1, 0, 1],
     ),
+    "pretranspiled_js": attr.label(
+        doc = "Exertanally transpiled .js to be included in output providers",
```
dzbarsky, line 74: Typo on externally

ts/private/ts_project.bzl
```diff
@@ -202,17 +202,9 @@ See https://github.com/aspect-build/rules_ts/issues/361 for more details.
...
-    default_outputs = []
-
-    # Default outputs (DefaultInfo files) is what you see on the command-line for a built
-    # library, and determines what files are used by a simple non-provider-aware downstream
-    # library. Only the JavaScript outputs are intended for use in non-TS-aware dependents.
-    if use_tsc_for_js:
-        # Special case case where there are no source outputs and we don't have a custom
-        # transpiler so we add output_types to the default outputs
-        default_outputs = output_sources if len(output_sources) else output_types
+    # Special case case where there are no source outputs
```
dzbarsky, line 205: Case case?

dzbarsky (commented):
nice, this looks simpler and makes our life better as well. We're trying this internally, will let you know

jbedard:
TODO: update https://github.com/aspect-build/rules_lint/blob/main/docs/eslint.md#with-ts_project

gregmagolan:
Yeah, this is nice that the macro produce the same rule type as it name. 👍

gregmagolan (approved):

PR #712, opened 2024-10-08
Test case for ts/js interop.

aptenodytes-forsteri:
---

### Test plan

- New test cases added


examples/ts_js_interop/BUILD.bazel
```diff
@@ -0,0 +1,25 @@
...
+
+ts_project(
+    name = "src",
+    srcs = [
+        "a.js",
+        "b.ts",
+    ],
+    allow_js = True,
+    declaration = True,
+    emit_declaration_only = True,
```
aptenodytes-forsteri, line 13: Commenting this out causes the `b.js` to be generated, but also causes a conflict when building all targets under this directory.

examples/ts_js_interop/tsconfig.json
```diff
@@ -0,0 +1,47 @@
...
+    "forceConsistentCasingInFileNames": true,
+    "noImplicitReturns": true,
+    "strict": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "experimentalDecorators": true,
+    "lib": ["dom", "es2022", "dom.iterable"],
+    "baseUrl": "./",
+    "paths": {
+      "wkspc/*": ["*", "bazel-out/k8-opt/bin/*"]
```
jbedard, line 16: I don't think we should have any `bazel-*` references in this file. Can we clean this up to the minimal config required to reproduce the issue?

examples/ts_js_interop/BUILD.bazel
jbedard: Lets comment it out then to reproduce the bug?

aptenodytes-forsteri: Both commented and uncommented are bugs.

Uncommented: I expect this example to build, not fail to produce JavaScript therefore not being able to create the file group.
Commented: I at least expect the targets to build, not result in that conflict error? Are you able to see both errors on your end when pulling these changes?

examples/ts_js_interop/tsconfig.json
aptenodytes-forsteri: Do you suspect the tsconfig is the culprit? I figured it was sort of irrelevant- just copied what I had in my project. We can try to trim it down to minimal, but I think this reproduction case hinges on the settings of the ts_project and the swc transpire step.

examples/ts_js_interop/BUILD.bazel
aptenodytes-forsteri: In other words, I'm between a rock and a hard place. If I don't emit, I don't get any transpiled JavaScript, even though I need the js to pass further downstream. In this example, I have reproduced that with the file group rule. 

If I do emit, I get an error about the same js file being produced by the transpile and type check steps.

My current workaround is to produce my own ts_project macro which calls tsc to produce the .d.ts files and calls swc to produce the transpiles. I can still use ts_project if all sources are typescript.

jbedard: Can we have 2 `ts_project` targets (maybe with different srcs) to reproduce both of those bugs?

I mainly just want the minimum code to reproduce the issue(s) without any extra code mixed any causing confusion.

examples/ts_js_interop/tsconfig.json
jbedard: I assume it's irrelevant but would like to remove it to prove that. I'd like the test case to be the absolute minimum amount of code to make it clear and precise what is being tested.

I don't think `paths` is at all related to the bug for example?

examples/ts_js_interop/BUILD.bazel
aptenodytes-forsteri: Ok, added a second `ts_project` target

examples/ts_js_interop/tsconfig.json
aptenodytes-forsteri: I modified the tsconfig files to be much more minimal.

examples/ts_js_interop/BUILD.bazel
```diff
@@ -0,0 +1,45 @@
...
+    resolve_json_module = True,
+    transpiler = partial.make(
+        swc,
+        swcrc = ":.swcrc",
+    ),
+    tsconfig = ":tsconfig.json",
+)
+
+filegroup(
+    name = "transpiled_a_b",
```
jbedard, line 23: I think these `filegroup` targets are never used.

Can we replace those with `build_test` targets so `bazel test //...` triggers them and they showup in the test results?

jbedard:
Sorry I forgot about this. Just 1 minor comment and a buildifier error now and then I think we can merge this.

examples/ts_js_interop/BUILD.bazel
nweinstein7: I'm not familiar with build_test targets, so probably? Have changes gone in that allow the file groups to build successfully?

aptenodytes-forsteri: I'm not familiar with build_test targets, but probably? Have changes gone in that allow these filegroups to build successfully now?

aptenodytes-forsteri:
> Sorry I forgot about this. Just 1 minor comment and a buildifier error now and then I think we can merge this.

Should we link this PR to the commits that made these test cases pass?

jbedard:
> Should we link this PR to the commits that made these test cases pass?

You can put links to the commits in the PR description 👍

aptenodytes-forsteri:
> > Should we link this PR to the commits that made these test cases pass?
> 
> You can put links to the commits in the PR description 👍

Sorry, I don't know what commits fixed this. I figure you know because you changed something to make these tests pass?

examples/ts_js_interop/BUILD.bazel
jbedard: Search for `build_test` in other tests for examples and try switching these `filegroup` to test targets.

jbedard:
Oh I see, sorry I don't recall either.

If you want you could bisect to determine which one did it, I assume that wouldn't take long (and your tests are new so you can just leave them not checked in while bisect-ing).

jbedard:
Actually it was probably this change: https://github.com/aspect-build/rules_ts/commit/dd832dbf5af2dc636e7a3673626356418c2d2653#diff-fd5b55e792f846b8312680267efd413bf73bfe655630485e1514db77eb7d187cL329-R370 ?

examples/ts_js_interop/BUILD.bazel
aptenodytes-forsteri: https://github.com/aspect-build/rules_ts/blob/3849ce6b8b660e83831c15696f03fd405ffd716e/ts/private/build_test.bzl#L7

I see, makes sense.  It's a trivial test that will always pass as long as it builds.

jbedard: Correct, I've almost exclusively used it for testing rules just like this test you've written.

aptenodytes-forsteri:
Have we confirmed the tests do indeed pass? I haven't returned to this since I made commits a while back.

> Actually it was probably this change: [dd832db#diff-fd5b55e792f846b8312680267efd413bf73bfe655630485e1514db77eb7d187cL329-R370](https://github.com/aspect-build/rules_ts/commit/dd832dbf5af2dc636e7a3673626356418c2d2653#diff-fd5b55e792f846b8312680267efd413bf73bfe655630485e1514db77eb7d187cL329-R370) ?

This timeline doesn't make sense to me. I have this commit in my fork of the repo where these tests fail. https://github.com/aptenodytes-forsteri/rules_ts/commits/ts-js-interop/

I was expecting something like: "The root cause of this test failures is code in file X at line Y. I have resolved this issue by making commit Z. The tests now pass and we can merge in this PR."

I'm not as familiar with this codebase or what the intended behavior is so I was hoping a maintainer could help root cause this issue or explain how I'm using it wrong.

aptenodytes-forsteri:
Basically, my ask since I produced the reproduction repo, opened the bug report, and opened this PR is for a maintainer of this code to pull down these reproduction cases and root cause why I can't seem to use ts and js seamlessly inside a ts_project. I currently have a workaround where I sort of "gut" ts_project and use parts of it in a separate macro when working with ts and js. The hope is that ts_project "just works" and I can get rid of my workaround macro.

aptenodytes-forsteri:
It seems like this run correctly reproduced both errors I see:

https://github.com/aspect-build/rules_ts/actions/runs/11327617927/job/32359980802?pr=712

So presumably the issue still persists and needs a root cause explanation and fix.

jbedard:
If you've reproduced a bug then I can take a look, and a good test case like this is the best way to demonstrate an issue 👍 

I'll try to look more tomorrow.

jbedard:
I'm wondering how much of this is just a limitation of bazel and rules_ts...

With `allowJs` we still need to copy the files into bazel-bin so they can be included for type-checking. That means we can't transpile them to the same location and would need to use `outDir` so the input != output.

examples/ts_js_interop/BUILD.bazel
```diff
@@ -0,0 +1,45 @@
...
+
+ts_project(
+    name = "a_b",
+    srcs = [
+        "a.js",
+        "b.ts",
+    ],
+    allow_js = True,
+    declaration = True,
+    emit_declaration_only = True,
```
jbedard, line 13: When using `emitDeclarationOnly` the transpiler won't run at all, tsc or a custom transpiler like swc, so `b.js` should not be expected at all

NateW-BD:
> I'm wondering how much of this is just a limitation of bazel and rules_ts...
> 
> With `allowJs` we still need to copy the files into bazel-bin so they can be included for type-checking. That means we can't transpile them to the same location and would need to use `outDir` so the input != output.

Right this is sort of what I'm getting at. I know it's a tricky case, but I want to transpile the js files *and* produce the .d.ts files *and* include them for typechecking.

My workaround involves a macro that filters out js sources into a separate swc and tsc macro which eventually get passed to a js library.

Then, if there are ts sources, too, I create a ts_project that depends on the js_library

examples/ts_js_interop/BUILD.bazel
NateW-BD: I figured as much- it seems like emit_declaration_only is not the fix to my issue and this test case can be deleted. I only included it to point out that this is one way to avoid the conflict error

aptenodytes-forsteri:
> I'm wondering how much of this is just a limitation of bazel and rules_ts...
> 
> With `allowJs` we still need to copy the files into bazel-bin so they can be included for type-checking. That means we can't transpile them to the same location and would need to use `outDir` so the input != output.

Right this is sort of what I'm getting at. I know it's a tricky case, but I want to transpile the js files *and* produce the .d.ts files *and* include them for typechecking.

My workaround involves a macro that filters out js sources into a separate swc and tsc macro which eventually get passed to a js library.

Then, if there are ts sources, too, I create a ts_project that depends on the js_library

examples/ts_js_interop/BUILD.bazel
aptenodytes-forsteri: Right, I figured as much. So this test case can be deleted and clearly does not solve my problem. I only included it to point out that this is *one* way to avoid the conflict error, but I need b.js to be produced in my environment.

aptenodytes-forsteri:
> With allowJs we still need to copy the files into bazel-bin so they can be included for type-checking

Could there be some conditional logic where, if the sources are js, you only copy the transpiled files to bazel-bin, not the raw source?

jbedard:
That makes the transpiler case diverge even more though, which I'd like to avoid :/

aptenodytes-forsteri:
> That makes the transpiler case diverge even more though, which I'd like to avoid :/

Another option: check for and explicitly disallow this case (which I guess is custom transpiler with a mix of ts and js sources)? Throw an error explaining that when using a custom transpiler you can't do allowJs and can't pass in js files to the ts_project? And perhaps provide an example of a workaround of a wrapper macro to filter out the js files and pass them to separate swc and tsc rules?

jbedard:
Closing for now, feel free to reopen

PR #714, opened 2024-10-11
feat: add transitive typecheck output group

MichaelMitchell-at:

With isolated declarations in TypeScript, it's no longer necessarily the case that typechecking a project also causes all of its dependencies to be typechecked. Some people prefer/expect the previous behavior when running typechecking locally, so this PR exposes a test target that will always typecheck a project and its dependencies, regardless whether any of them have enabled the `isolated_typecheck` flag.

### Test plan

Works well in our codebase!

### DIFF
--- a/ts/defs.bzl
+++ b/ts/defs.bzl
@@ -370,7 +370,9 @@ def ts_project(
     if not emit_tsc_js or not emit_tsc_dts:
         types_target_name = "%s_types" % name
         typecheck_target_name = "%s_typecheck" % name
+        transitive_typecheck_target_name = "%s_transitive_typecheck" % name
         test_target_name = "%s_typecheck_test" % name
+        transitive_typecheck_test_target_name = "%s_transitive_typecheck_test" % name
 
         # Users should build this target to get typing files
         if not no_emit:
@@ -397,10 +399,26 @@ def ts_project(
             **common_kwargs
         )
 
+        # Users should build this target to get a failed build when typechecking fails
+        native.filegroup(
+            name = transitive_typecheck_target_name,
+            srcs = [name],
+            output_group = "transitive_typecheck",
+            **common_kwargs
+        )
+
         # Ensures the typecheck target gets built under `bazel test --build_tests_only`
         build_test(
             name = test_target_name,
             targets = [typecheck_target_name],
+            tags = common_kwargs.get("tags"),
+            size = "small",
+            visibility = common_kwargs.get("visibility"),
+        )
+
+        build_test(
+            name = transitive_typecheck_test_target_name,
+            targets = [transitive_typecheck_target_name],
             tags = common_kwargs.get("tags"),
             size = "small",
             visibility = common_kwargs.get("visibility"),
--- a/ts/defs.bzl
+++ b/ts/defs.bzl
@@ -404,7 +404,9 @@ def ts_project(
             name = transitive_typecheck_target_name,
             srcs = [name],
             output_group = "transitive_typecheck",
-            **common_kwargs
+            tags = ["manual"] + common_kwargs.get("tags", []),
+            visibility = common_kwargs.get("visibility"),
+            testonly = common_kwargs.get("testonly"),
         )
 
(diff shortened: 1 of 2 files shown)


ts/private/ts_project.bzl
```diff
@@ -29,6 +29,15 @@ def _gather_types_from_js_infos(targets):
     ])
     return depset([], transitive = files_depsets)
 
+def _gather_transitive_typecheck_from_output_group_infos(typecheck_outs, targets):
+    files_depsets = []
+    files_depsets.extend([
```
jbedard, line 34: Can we just inline this?

MichaelMitchell-at, line 34: I was copy-pasting the code above but I guess I can just assign the list directly

ts/defs.bzl
```diff
@@ -406,6 +416,14 @@ def ts_project(
             visibility = common_kwargs.get("visibility"),
         )
 
+        build_test(
```
jbedard, line 419: I'm kind of tempted to not expose this for now, at least as an initial poc.

jbedard:
Can you explain the use case again in the PR description?

MichaelMitchell-at:
> Can you explain the use case again in the PR description?

Done!

ts/defs.bzl
jbedard: I think this means `test //...` will run this which I think we want to avoid, so we should either drop it for now or tag it as manual...

MichaelMitchell-at: Oh good call

ts/defs.bzl
```diff
@@ -397,10 +399,26 @@ def ts_project(
             **common_kwargs
         )
 
+        # Users should build this target to get a failed build when typechecking fails
+        native.filegroup(
+            name = transitive_typecheck_target_name,
+            srcs = [name],
+            output_group = "transitive_typecheck",
+            **common_kwargs
```
jbedard, line 407: Does this also have to be manual to avoid `build //...` triggering all transitive `{name}_transitive_typecheck` targets to build?

ts/defs.bzl
```diff
@@ -397,10 +399,26 @@ def ts_project(
...
+        )
+
         # Ensures the typecheck target gets built under `bazel test --build_tests_only`
         build_test(
             name = test_target_name,
             targets = [typecheck_target_name],
+            tags = ["manual"] + common_kwargs.get("tags", []),
```
jbedard, line 414: Wrong target for the manual tag, should be the `transitive_` one...

MichaelMitchell-at: That's what I get for trying to make a quick edit from my phone 🤦

gregmagolan:
This is cool. Needs to be more discoverable. There are some docs on macro expansion in `docs/transpiler.md` that should be updated if a new expansion is added.

I'll pull it down and experiment with it over the weekend. Feels like it should be possible to set as the default behavior as long as the type check actions are still able to be run in parallel.

jbedard:
@gregmagolan I was thinking of explicitly **not** making it discoverable to start, and marking it as manual so it doesn't show up doing `test //...`. Mainly just so we can try it out before making it public API. Even just so people can test it at HEAD for now 🤷 

I think you are correct that the type-check actions should still be able to run in parallel, we just need to make sure we set this `_test` target is setup correctly and make sure. In that case I think we can do that by default and we only need one `build_test` target? Which would be another reason not to make this new one public to start...?

jbedard:
For the parallelization, assuming we have `isolated_typecheck` enabled:
* each `_typecheck` (slow tsc) target needs the `_types` of each dependency
* `_types` can all be done in parallel by the `declaration_transpiler`
* nothing depends on `_typecheck`, so tsc never depends on another tsc

For this `_transitive_typecheck`... that output_group is basically just a `depset` of tsc actions, but those actions themselves never depend on eachother so I think they will still all be done in parallel? So do we still need the non-transitive one? I feel like "typecheck this target only" should still be an option but 🤷

jbedard (approved):

PR #715, opened 2024-10-11
feat: always declare a ts_project _types target to fetch dts files

jbedard:
Sometimes the dts files need to be fetched, there is no reason this shouldn't be done when a transpiler is not being used.

This makes the declare targets more predictable.

---

`ts_project(name)` declaration files can be fetched using the `{name}_types` target.

### Test plan

- New test cases added


examples/transpiler/BUILD.bazel
```diff
@@ -93,6 +101,27 @@ build_test(
...
 
+# Runs babel to transpile ts -> js while tsc would fail with typecheck errors
+# Tagging as manual should prevent the typecheck from running even when requesting
+# transpiler output files.
+ts_project(
+    name = "typecheck_fail",
+    srcs = ["typecheck_fail.ts"],
+    declaration = True,
+    source_map = True,
+    tags = ["manual"],
```
jbedard, line 112: This test is extra, I just want to make sure failing tsc is not invoked when tagged as manual, even when the non-tsc transpiler is forced to be invoked.

examples/targets/BUILD.bazel
```diff
@@ -35,17 +35,36 @@ ts_project(
...
         ":gen-a-ts",
     ],
+    declaration = True,
     out_dir = "out",
+    source_map = True,
+    tsconfig = "tsconfig.json",
 )
 
-# Ensure all the outputs are pre-declared
+# Ensure all the outputs are pre-declared and in the correct locations including
```
dzbarsky, line 44: unfinished sentence?

dzbarsky (approved):

dzbarsky:
@jbedard disabled auto merge so you an address the comment

gregmagolan (approved):

examples/targets/BUILD.bazel
jbedard: thank you lol

jfirebaugh:
The docs in https://github.com/aspect-build/rules_ts/blob/main/docs/transpiler.md#macro-expansion should be updated to say that the `[name]_types` expansion is always generated.

Issue #716, opened 2024-10-15
[Bug]: #710 changed copy_runfiles_to_bin behavior of ts_project if `transpiler` is set


gzm0:
### What happened?

When upgrading to 3.2.1 (from 3.2.0) the `copy_runfiles_to_bin` behavior of `ts_project` changes when `transpiler` is set:

```
ts_project(
  data = ["my-file.txt"],
  transpiler = swc,
)
```

Up until 3.2.0, `my-file.txt` gets copied to bin. After 3.2.1, it doesn't anymore (due to the refactor in #710).

Somewhat ironically, I expect this will fix #411 (there is more information in that bug for why this happens), but it is a breaking change.

### Version

Development (host) and target OS/architectures:

Output of `bazel --version`: bazel 7.3.1

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file: 3.2.0 / 3.2.1

Language(s) and/or frameworks involved: None.

gzm0:
/cc @jbedard

I'm happy to help here if you need more infos. I think the toughest thing here is to figure out how we want the behavior to be (once that is settled, it's IIUC only flag wiring).

jbedard:
Without any further issues raised I think we'll just keep this change as-is since I believe it is now consistent and "more correct", even if it may have been breaking for some use cases.

Issue #411, opened 2023-08-16
[Bug]: copy_data_files_to_bin (for gather_runfiles) depends on whether transpiler is set


gzm0:
### What happened?

It seems that ts_project doesn't expose this config option yet and the defaults are different.

# Without transpiler

1. data is passed to the `ts_project` rule here:
https://github.com/aspect-build/rules_ts/blob/42094e84545d3420081da240819a123aa02b69ce/ts/defs.bzl#L406-L412

3. The `ts_project` rule invokes `gather_runfiles`
https://github.com/aspect-build/rules_ts/blob/42094e84545d3420081da240819a123aa02b69ce/ts/private/ts_project.bzl#L270-L275

4. [The default for `copy_data_files_to_bin` is `False`](https://github.com/aspect-build/rules_js/blob/0109c77ee53303264631cc03f2cc3da772908c4d/js/private/js_helpers.bzl#L177)

# With transpiler

1. data is passed to `js_library` here:
https://github.com/aspect-build/rules_ts/blob/42094e84545d3420081da240819a123aa02b69ce/ts/defs.bzl#L392-L400

2. [The default for `js_library` is `True`](https://github.com/aspect-build/rules_js/blob/0109c77ee53303264631cc03f2cc3da772908c4d/js/private/js_library.bzl#L99-L102)

# What I'd expect

- The behavior with and without `transpiler` to be the same.
- There be an equivalent option to `js_library` on `ts_project` (the macro).

### Version

Development (host) and target OS/architectures:

Output of `bazel --version`:

```
bazelisk --version
bazel 6.3.2
```

Version of the Aspect rules, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file:

```
rules_js: 1.31.0
rules_ts: 1.4.5
```

Language(s) and/or frameworks involved: None.

### How to reproduce

```shell
Minimal steps, happy to provide more if necessary.

1. Add a filegroup with files in a different package to `data` of a `ts_project` (without `transpiler` set).
2. Build passes.
3. Set `transpiler` to `swc`.
4. Build fails (example failure below)
5. Switch `filegroup` to `js_library`
6. Build passes



Expected to find source file <the file> in '//package-of-ts-project', but instead it is in '//package-of-file'.

All source and data files that are not in the Bazel output tree must be in the same package as the
target so that they can be copied to the output tree in an action.

See https://docs.aspect.build/rules/aspect_rules_js/docs/#javascript for more context on why this is required.
```
```


[...]

PR #717, opened 2024-10-15
fix: re-establish pre-3.2.1 data runfile behavior

gzm0:
WIP fix for #716.

- Needs tests
- Needs decision on whether to expose `copy_data_to_bin` on `ts_project`.
- Needs decision on whether to add `no_copy_to_bin` as well (not required, but related).
---

### Test plan

- New test cases added


gzm0:
Ready for review.

An alternative would be to not include the additional parameter in the macro.

gzm0:
Sorry for the delayed reply, I was on vacation. I've answered here: https://github.com/aspect-build/rules_ts/pull/723#issuecomment-2459315003

gzm0:
Hum... The original issue was about misaligned behavior for copy_data_to_bin if transpiler is set or not. IIUC: #723 aligned the behavior, which was a breaking change, but (IMHO) the better overall behavior.

So at this point, with the change almost over a year old, I'm not sure it's a good idea to revert to the behavior that is technically correct under semver. I would suggest to:
- close this PR
- close #411 as fixed (probably verify before you take my word)
- close #716 as wontfix (probably verify before you take my word).

jbedard:
That sounds best to me since no one else commented on this change in behaviour over the last year 👍

PR #718, opened 2024-10-16
build: upgrade to latest bazel 7.3.2

jbedard:
### Test plan


gregmagolan:
XDG_CACHE_HOME is also problematic for other reasons: https://github.com/bazel-contrib/bazel-lib/pull/914#issuecomment-2299383522

jbedard:
@gregmagolan I have no idea what the latest error is and haven't been able to figure it out 🤷 
```
(22:58:41) ERROR: /home/runner/work/rules_ts/rules_ts/docs/BUILD.bazel:5:23: in deps attribute of starlark_doc_extract rule //docs:rules.extract: missing bzl_library targets for Starlark module(s) @platforms//host:constraints.bzl. Since this rule was created by the macro 'stardoc_with_diff_test', the error might have been caused by the macro implementation
```

gregmagolan:
> @gregmagolan I have no idea what the latest error is and haven't been able to figure it out 🤷
> 
> ```
> (22:58:41) ERROR: /home/runner/work/rules_ts/rules_ts/docs/BUILD.bazel:5:23: in deps attribute of starlark_doc_extract rule //docs:rules.extract: missing bzl_library targets for Starlark module(s) @platforms//host:constraints.bzl. Since this rule was created by the macro 'stardoc_with_diff_test', the error might have been caused by the macro implementation
> ```

Need to update bazel-lib to a minimum of 2.9.0 to pickup https://github.com/bazel-contrib/bazel-lib/pull/914. But can't do that since there is a regression in libarchive in 3.7.5 that breaks the tar toolchain so first have to update bazel-lib to use a pre-build bsdtar using libarchive 3.7.7 :sweat_smile:

So first have to verify that libarchive 3.7.7 fixes the regression in bsdtar-prebuilt (we caught the failure in silo when trying to bump to latest bazel-lib). Then can land an update to bazel-lib and then can update rules_js & friends minimum bazel-lib to green this up.

gregmagolan:
After https://github.com/bazel-contrib/bazel-lib/pull/969 lands and a bazel-lib release is cut then this can be greened up

PR #720, opened 2024-10-18
refactor: only pass --outDir flag when needed

alexeagle:
Reduces the severity of the workaround needed for TS #59036

### DIFF
--- a/ts/private/ts_project.bzl
+++ b/ts/private/ts_project.bzl
@@ -125,23 +125,27 @@ See https://github.com/aspect-build/rules_ts/issues/361 for more details.
     # Add user specified arguments *before* rule supplied arguments
     common_args.extend(ctx.attr.args)
 
-    outdir = _lib.join(
-        ctx.label.workspace_root,
-        _lib.join(ctx.label.package, ctx.attr.out_dir) if ctx.attr.out_dir else ctx.label.package,
-    )
+    if (ctx.attr.out_dir and ctx.attr.out_dir != ".") or ctx.attr.root_dir:
+        # TODO: add validation that excludes is non-empty in this case, as passing the --outDir or --declarationDir flag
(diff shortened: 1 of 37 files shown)


alexeagle:
Note, the failing test is red on `main` as well.

ts/private/ts_project.bzl
```diff
@@ -125,24 +125,28 @@ See https://github.com/aspect-build/rules_ts/issues/361 for more details.
     # Add user specified arguments *before* rule supplied arguments
     common_args.extend(ctx.attr.args)
 
-    outdir = _lib.join(
-        ctx.label.workspace_root,
-        _lib.join(ctx.label.package, ctx.attr.out_dir) if ctx.attr.out_dir else ctx.label.package,
-    )
+    if (ctx.attr.out_dir and ctx.attr.out_dir != ".") or ctx.attr.root_dir:
+        # TODO: add validation that excludes is non-empty in this case, as passing the --outDir or --declarationDir flag
+        # to TypeScript causes it to set a default for excludes such that it won't find our sources that were copied-to-bin.
```
jbedard, line 130: Can we put a reference to https://github.com/microsoft/TypeScript/issues/59036 and the rules_js issue about it?

alexeagle: done

jbedard (approved):

jbedard:
Double check the worker CI failure before any merging 🙏

gzm0:
FWIW: This change causes the `outDir` config in `tsconfig.json` to be used while previously it wasn't.

Maybe it is worth adjusting the config validator to check it isn't there / it is consistent.

alexeagle:
Yes jbedard already noticed that at some customers, I think there's an issue and a fix already.

PR #705, opened 2024-09-27
refactor: simplify ts_project macro conditional logic

jbedard:
Minor var renaming and usage changes to clarify intentions.
Moving common code to util macro.

---

### Test plan

- Covered by existing test cases

PR #721, opened 2024-10-18
fix: do not declare _typecheck[_test] target when using tsc as transpiler

### LINKED ISSUE #719: [Bug]: Superfluous @ts_project_typecheck_test target being generated with v3.2.0
### What happened?

When upgrading from rules_ts v3.1.0 => 3.2.0 or higher, a new `@ts_project_typecheck_test` test target is unexpectedly being generated for some `ts_project`-based targets. I believe that this is being caused by an unintentional logic change in #705 that causes the target to be newly generated when the `declaration` parameter is set to false. https://github.com/aspect-build/rules_ts/pull/705/files#diff-fd5b55e792f846b8312680267efd413bf73bfe655630485e1514db77eb7d187cR399

My understanding is that this target is not needed since tsc will still be invoked as long as no custom transpiler is set and js files are still being generated. The test target fails in some configurations for our build (including the Windows build), but we can't rule out that this is due to an unrelated issue with our build. The issue here is that an unnecessary additional target is being generated.

### Version

macOS Sonoma 14.7 (among others)

Output of `bazel --version`:
bazel 7.3.1
Version of the Aspect rules, or other relevant rules from your `WORKSPACE` or `MODULE.bazel` file:
rules_js v2.0.1
bazel-lib 2.9.2

TypeScript

### How to reproduce

```shell
Declare a dummy ts_project target in a bazel file:
ts_project(
    name = "dummy@ts_project",
    srcs = ["dummy.ts"],
    tsconfig = "tsconfig.json",
)
alongside an empty dummy.ts file and a tsconfig.json file with no content other than `{}`. Then rules_ts will generate a superfluous dummy@ts_project_typecheck_test target starting with v3.2.0 that was not generated before.
```

### Any other information?

_No response_

jbedard:
Declaring the `_typecheck[_test]` targets is only required when tsc is not doing the typechecking while transpiling something.

We did add the `_types` target 100% of the time, transpiler(s) or not. I think we should keep that target 100% of the time. But when should the `_typecheck[_test]` targets be added? This reverts back to be more like before https://github.com/aspect-build/rules_ts/pull/705 where they only exist if tsc will not do type-checking in a different action.

Fix #719

---

### Test plan

- New test cases added

### DIFF
--- a/examples/transpiler/BUILD.bazel
+++ b/examples/transpiler/BUILD.bazel
@@ -84,7 +84,7 @@ build_test(
     targets = [
         # babel ts_project
         ":babel",
-        ":babel_typecheck",
+        ":babel_types",
         # babel outputted js
         "big.js",  # NOTE: does not implement out_dir in this test
         # tsc outputted dts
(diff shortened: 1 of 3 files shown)


jbedard:
So far voters are [in favour of keeping this target](https://github.com/aspect-build/rules_ts/issues/719#issuecomment-2430079938) which I also agree with, so converting to draft for now...

PR #723, opened 2024-10-21
fix: align ts_project(data) runfiles with js_library(data)

jbedard:
After [removing the use of `js_library`](https://github.com/aspect-build/rules_ts/commit/5e25e91420947e3a81938d8eb076803e5cf51fe2) the `ts_project(data)` no longer conditionally goes through `js_library(data)` when the `transpiler` attribute is set. However the `js_library(data)` was creating the runfiles differently then `ts_project_rule(data)`.

This change makes `ts_project(data)` align with the defaults of `js_library(data)` to restore the previous behaviour.

---

### Test plan

- New test cases added


alexeagle (approved):

gzm0:
I'm afraid this is a breaking change for usages of `ts_project` without a custom transpiler:

`copy_data_files_to_bin = True` comes with the caveat that it is source files coming from a different package are not allowed in `data`.

This breaks a basic filegroup:

```bzl
# package //files
filegroup(
  name = "files",
  srcs = [...],
)

# package //code
ts_project(
  data = ["//files"],  # this will break with this change.
)
```

PR #726, opened 2024-10-28
Ensuring package_prefix is correct for external workspaces

menny:
This is a fix over the changes in https://github.com/aspect-build/rules_ts/pull/687.

### Test plan
New unit-tests

- Covered by existing test cases: partial
- New test cases added: added under `e2e` folder.
  - Add a TS repo as an external repository. This should have a target that:
    - is a `ts_project`
    - inside a subfolder
    - has `tsconfig` attribute set to a `dict`.
  - build the `ts_project` target inside this external repository

### DIFF
--- a/e2e/external_dep/BUILD
+++ b/e2e/external_dep/BUILD
@@ -38,3 +38,28 @@ build_test(
         "lib.ts56rc",
     ],
 )
+
+ts_project(
+    name = "external-deps",
+    srcs = ["lib.ts"],
+    composite = True,
+    out_dir = "external-deps",
+    visibility = ["//visibility:public"],
+    deps = [
+        "@sub_external//:lib1",
+        "@sub_external//:tsconfig-dict",
+        "@sub_external//subdir:lib1",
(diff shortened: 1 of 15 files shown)


jbedard:
If we can just get a test reproducing it that would be great 👍

e2e/external_dep/BUILD
```diff
@@ -38,3 +38,26 @@ build_test(
...
+
+ts_project(
+    name = "external-deps",
+    srcs = ["lib.ts"],
+    composite = True,
+    out_dir = "external-deps",
+    visibility = ["//visibility:public"],
+    deps = [
+        "@sub_external//:lib1",
+        "@sub_external//:tsconfig-dict",
```
jbedard, line 50: Maybe add the 2 `subdir` ones here as well? Just to test they still work as `ts_project(deps)` even though the code doesn't load them 🤷).

menny: done

jbedard (approved):

PR #728, opened 2024-10-31
fix: support setting resolveJsonModule with inline ts_project(tsconfig)

jbedard:
I noticed `resolveJsonModule` is treated different then all the other tsconfig attributes which must be aligned between the macro and tsconfig where you can't depend on the tsconfig dict. This allows setting the value only in the tsconfig dict like any other compiler option.

---

### Test plan

### DIFF
--- a/ts/defs.bzl
+++ b/ts/defs.bzl
@@ -260,15 +260,6 @@ def ts_project(
         **kwargs: passed through to underlying [`ts_project_rule`](#ts_project_rule), eg. `visibility`, `tags`
     """
 
-    if srcs == None:
-        include = ["**/*.ts", "**/*.tsx"]
-        exclude = []
-        if allow_js == True:
-            include.extend(["**/*.js", "**/*.jsx"])
-        if resolve_json_module == True:
-            include.append("**/*.json")
-            exclude.extend(["**/package.json", "**/package-lock.json", "**/tsconfig*.json"])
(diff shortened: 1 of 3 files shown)


ts/defs.bzl
```diff
@@ -463,6 +463,19 @@ def ts_project(
         **kwargs
     )
 
+def _default_srcs(allow_js, resolve_json_module):
+    """Returns a list of srcs for ts_project when srcs is not provided."""
+    srcs = []
```
dzbarsky, line 468: Srcs is always redefined below, snip?

ts/defs.bzl
```diff
@@ -463,6 +463,19 @@ def ts_project(
...
+    include = ["**/*.ts", "**/*.tsx"]
+    exclude = []
```
dzbarsky, line 470: Just set exclude to final value below instead?

dzbarsky (approved):

ts/defs.bzl
jbedard: If `glob(exclude)` accepted `None` then I would. But I'm mainly just moving this code and don't want to think about it anymore 😅

(I did the other suggestion re the useless `src = []` though)

dzbarsky: Oh I see it's only set inside the condition. Hard to tell on phone :) you could also override the car instead of extend, that would save some keystrokes :)

Issue #727, opened 2024-10-31
[Bug]: ts_project error when srcs include parent/sibling files


jbedard:
### What happened?

When the `ts_project(srcs)` is passed files that do not belong to the current BUILD odd errors occur.

1. passing `.ts` from parent/sibling packages:

Previous, before https://github.com/aspect-build/rules_ts/commit/85d969ee603f159dda3502deceb15b02cd9d1135, tsc would output an error saying the .ts is outside the `rootDor`.

After https://github.com/aspect-build/rules_ts/commit/85d969ee603f159dda3502deceb15b02cd9d1135 `write_tsconfig` computes the wrong path and tsc outputs an error saying the file doesn't exist instead.

2. passing `.ts` from subpackages

GUESS: tsc will not complain about the inputs but will most likely fail to write the outputs.

3. passing `.d.ts` from parent/sibling packages:

Previously: may have passed depending if the dts is already in the bindir (so no copy action is created)
Now: probably fails because the path is computed wrong

### Version

All, the error changed after https://github.com/aspect-build/rules_ts/commit/85d969ee603f159dda3502deceb15b02cd9d1135

### How to reproduce

https://github.com/jbedard/rules_ts/commit/ebb4759a87e635db1514b5b3d4476a97ea667410

PR #729, opened 2024-10-31
fix: ts_project declaration sources from other directories

jbedard:
Fix https://github.com/aspect-build/rules_ts/issues/727

---

### Test plan

- New test cases added

### DIFF
--- a/examples/global_consumer/tsconfig_gen.json.expected
+++ b/examples/global_consumer/tsconfig_gen.json.expected
@@ -1,0 +1,1 @@
+{"compilerOptions":{"allowJs":false,"declaration":false,"declarationMap":false,"emitDeclarationOnly":false,"noEmit":false,"resolveJsonModule":null,"sourceMap":false},"files":["./index.ts", "./../../examples/global_dts_generated_tsconfig/index.d.ts", "./../../examples/global_dts_generated_tsconfig/global.d.ts"]}
--- a/ts/private/ts_config.bzl
+++ b/ts/private/ts_config.bzl
@@ -114,14 +114,6 @@ extended configuration file as well, to pass them both to the TypeScript compile
     toolchains = COPY_FILE_TO_BIN_TOOLCHAINS,
 )
 
-def _filter_input_files(files, allow_js, resolve_json_module):
-    return [
-        f
-        for f in files
-        # include typescript, json & declaration sources
-        if _lib.is_ts_src(f.basename, allow_js, resolve_json_module) or _lib.is_typings_src(f.basename)
-    ]
-
 def _write_tsconfig_rule(ctx):
     # TODO: is it useful to expand Make variables in the content?
     content = ctx.attr.content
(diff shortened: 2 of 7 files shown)


examples/global_consumer/tsconfig_gen.json.expected
```diff
@@ -0,0 +1 @@
+{"compilerOptions":{"allowJs":false,"declaration":false,"declarationMap":false,"emitDeclarationOnly":false,"noEmit":false,"resolveJsonModule":null,"sourceMap":false},"files":["./index.ts", "./../../examples/global_dts_generated_tsconfig/index.d.ts", "./../../examples/global_dts_generated_tsconfig/global.d.ts"]}
```
jbedard, line 1: While these .expected files are not pretty I think it is important to watch out for these relative paths changing without us noticing. The `./` for files in this package and the `../../` for files outside the package.

I want to prevent us not noticing the `if .startswith(local` breaking and these all becoming the `../../` paths.

ts/private/ts_config.bzl
```diff
@@ -135,22 +127,30 @@ def _write_tsconfig_rule(ctx):
...
-        "./" + f.short_path.removeprefix(package_prefix)
-        for f in filtered_files
-    ]
-
-    content = content.replace("\"__files__\"", str(filtered_files))
+        # If the target is inside another workspace the prefix also contains the navigation to that workspace.
+        local_package_prefix = "../{}/{}".format(ctx.label.repo_name, local_package_prefix)
+
+    # The path to navigate to the root of the workspace
+    path_to_root = "/".join([".." for _ in ctx.label.package.split("/")])
```
dzbarsky, line 137: maybe `path_to_root = "/".join([".."] * ctx.label.package.count("/"))` is faster (I don't recall if we decided "list * const" was good or not). But at least `.count` is cleaner I think?

ts/private/ts_config.bzl
```diff
@@ -135,22 +127,30 @@ def _write_tsconfig_rule(ctx):
...
+
+    # Compute the list of source files with paths relative to the generated tsconfig file.
+    src_files = []
+    for f in ctx.files.files:
+        # Only include typescript source files
+        if not (_lib.is_ts_src(f.basename, ctx.attr.allow_js, ctx.attr.resolve_json_module) or _lib.is_typings_src(f.basename)):
+            continue
+
+        if f.short_path.startswith(local_package_prefix):
+            # Files within this project or subdirs can avoid the ugly ../ prefix
```
dzbarsky, line 147: does it actually matter? This output isn't for humans, it's fed directly to tsc, right? I'd maybe just go for the simpler code

dzbarsky (approved):

ts/private/ts_config.bzl
jbedard: In theory it should be faster, and seems cleaner anyway 👍

ts/private/ts_config.bzl
jbedard: This PR doesn't change that so lets debate it another time

ts/private/ts_config.bzl
jbedard: Just note it has to be `count("/") + 1` to align with split, but seems to be fine 👍

Issue #644, opened 2024-06-26
[Bug]: typescript 5.5.2 cannot find input files with source tsconfig.json


gzm0:
### What happened?

Minimal reproducible example below.

With TS 5.5.2, sources cannot be found:

```
bazel build //...
INFO: Analyzed target //:test (91 packages loaded, 2863 targets configured).
WARNING: /home/tos/.cache/bazel/_bazel_tos/287294dcdaea82c1370bff9bd5064654/external/aspect_rules_ts~~ext~npm_typescript/BUILD.bazel:32:17: input 'package' of @@aspect_rules_ts~~ext~npm_typescript//:.aspect_rules_js/node_modules/typescript is a directory; dependency checking of directories is unsound
ERROR: /home/tos/dh/personal/hack/tsc/BUILD.bazel:3:11: Transpiling TypeScript project @@//:test [tsc -p tsconfig.json] failed: (Exit 2): tsc.sh failed: error executing TsProject command (from target //:test) bazel-out/k8-opt-exec-ST-d57f47055a04/bin/external/aspect_rules_ts~~ext~npm_typescript/tsc.sh --skipLibCheck --project tsconfig.json --outDir . --rootDir .

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
error TS18003: No inputs were found in config file '/home/tos/.cache/bazel/_bazel_tos/287294dcdaea82c1370bff9bd5064654/sandbox/linux-sandbox/984/execroot/_main/bazel-out/k8-fastbuild/bin/tsconfig.json'. Specified 'include' paths were '["**/*"]' and 'exclude' paths were '["/home/tos/.cache/bazel/_bazel_tos/287294dcdaea82c1370bff9bd5064654/sandbox/linux-sandbox/984/execroot/_main/bazel-out/k8-fastbuild/bin"]'.
Target //:test failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 1.016s, Critical Path: 0.30s
INFO: 4 processes: 3 internal, 1 local.
ERROR: Build did NOT complete successfully
```

When downgrading to 5.4.5:

```
bazel build //...
INFO: Analyzed target //:test (90 packages loaded, 2863 targets configured).
[...]

PR #730, opened 2024-11-01
feat: validate tsconfig outDir vs ts_project(out_dir)

Mivr:
### Test plan

- New test cases added

Resolves: https://github.com/aspect-build/rules_ts/issues/644

By adding a validation to the tsconfig file. If a out_dir parameter is set to the rule and the tsconfig file does not have the exclude key set the build will fail


ts/private/ts_project_options_validator.js
```diff
@@ -142,6 +142,15 @@ function main(_a) {
             }
         }
     }
+
+    function check_exclude_and_rootDir() {
+        if (config.exclude === undefined && attrs['root_dir'] !== undefined && attrs['root_dir'] !== '') {
+            console.error('tsconfig validation failed: when root_dir is set, exclude must also be set. See: https://github.com/aspect-build/rules_ts/issues/644 for more details.')
```
jbedard, line 148: Should we say "tsconfig exclude"?
Maybe also "when attribute root_dir is set"?

Just to try to distinguish between attributes in BUILD vs keys in tsconfig.json

examples/declaration_dir/tsconfig.json
```diff
@@ -2,6 +2,8 @@
     "compilerOptions": {
         "sourceMap": true,
         "declaration": true,
-        "declarationMap": true
-    }
+        "declarationMap": true,
+        "declarationDir": "out/types"
+    },
+    "exclude": []
```
jbedard, line 8: Lets add comments above this referencing the issue?

See the comments that were removed in https://github.com/aspect-build/rules_ts/commit/28ea0571a2f32616739cb506398d78ab08d9dde8

ts/defs.bzl
```diff
@@ -294,7 +294,6 @@ def ts_project(
 
         # These options are always passed on the tsc command line so don't include them
         # in the tsconfig. At best they're redundant, but at worst we'll have a conflict
-        out_dir = compiler_options.pop("outDir", out_dir)
```
jbedard, line 297: I think we would still want this so we can do `tsconfig = {... outDir ...}` without also needing `ts_project(tsconfig)`? Just like thigns like `declaration`, so maybe just move it up a few lines?

jbedard: Then in many places where you added `ts_project(tsconfig = {... outDir ... })` we can drop the `ts_project(out_dir)`?

Mivr:
After some discussions with @jbedard, I tested if the exclude is still needed at all for several parameters.  
The tests made are as follows:

## root_dir parameter testing (Latest rules_ts + default TS and node version)
| root_dir | tsconfig rootDir | tsconfig exclude | Result                       |
|----------|------------------|------------------|------------------------------|
| .        | undefined        | undefined        | BUILD PASSED                 |
| ./       | undefined        | undefined        | BUILD PASSED                 |
| ./src    | undefined        | undefined        | BUILD FAILED - cannot find outputs    |
| ./src    | undefined        | []               | BUILD FAILED - cannot find outputs    |
| ./src    | src              | undefined        | BUILD FAILED - cannot find outputs    |
| ./src    | ./src            | undefined        | BUILD FAILED - cannot find outputs    |
| src      | src              | []               | BUILD PASSED                 |
| src      | src              | undefined        | BUILD PASSED                 |
| src      | ./src            | undefined        | BUILD PASSED                 |
| src      | undefined        | undefined        | BUILD PASSED                 |
| src      | undefined        | []               | BUILD PASSED                 |

## out_dir parameter testing (Latest rules_ts + default TS and node version)
| out_dir    | tsconfig outDir | tsconfig exclude | Result                          |
|------------|-----------------|------------------|---------------------------------|
| out        | Undefined       | Undefined        | BUILD PASSED                    |
| out        | out             | Undefined        | BUILD PASSED                    |
| out        | ./out           | Undefined        | BUILD PASSED                    |
| Undefined  | ./out           | Undefined        | BUILD FAILED - cannot find outputs       |
| Undefined  | ./out           | []               | BUILD FAILED - cannot find outputs       |
| Undefined  | out             | Undefined        | BUILD FAILED - cannot find outputs       |
| Undefined  | out             | []               | BUILD FAILED - cannot find outputs       |
| .          | Undefined       | Undefined        | BUILD PASSED                    |
| ./         | Undefined       | Undefined        | BUILD FAILED - ./ messes up outputs paths|
| ./         | Undefined       | []               | BUILD FAILED - ./ messes up outputs paths|
| ./out      | Undefined       | Undefined        | BUILD FAILED - ./ messes up outputs paths|

## declaration_dir parameter testing declaration=True (Latest rules_ts + default TS and node version)
| declaration_dir | tsconfig declarationDir | tsconfig exclude | Result                                |
|-----------------|-------------------------|------------------|---------------------------------------|
| "dec"           | "dec"                  | Undefined        | BUILD PASSED                          |
| "dec"           | Undefined              | Undefined        | BUILD FAILED - cannot find output declarations |
| Undefined       | "dec"                  | Undefined        | BUILD FAILED - cannot find output declarations |

All tests were made before the changes in this branch. No tests show the need for `exclude` to be added at this point.

Is there a test case that still fails with the latest `rules_ts`?

And should we make these errors more friendly?  
@jbedard

ts/private/ts_project_options_validator.js
```diff
@@ -201,6 +212,8 @@ function main(_a) {
             attrs.declaration +
             '\n// declaration_map:       ' +
             attrs.declaration_map +
+            '\n// declaration_dir:       ' +
+            attrs.declaration_dir +
             '\n// incremental:           ' +
             attrs.incremental +
             '\n// source_map:            ' +
```
jbedard, line 219: Should `root_dir` be added to this list?

jbedard (approved):

ts/private/ts_project_options_validator.js
```diff
@@ -65,15 +65,20 @@ function main(_a) {
         }
         return options[option]
     }
-    function check(option, attr) {
+    function check(option, attr, failOnlyIfAttrIsUndefined = false) {
```
jbedard, line 68: Can we just add a `check_outdir()` method instead of `failOnlyIfAttrIsUndefined`? All these booleans and "should fail because" vars seems more complicated then adding one additional method.

Mivr: Added

jbedard (approved):

PR #732, opened 2024-11-05
fix: windows typecheck_test .sh vs .bat error

jbedard:
See https://github.com/bazelbuild/bazel-skylib/blob/1.7.1/rules/build_test.bzl#L19-L33

---

### Test plan

### DIFF
--- a/ts/private/build_test.bzl
+++ b/ts/private/build_test.bzl
@@ -11,12 +11,15 @@ It is used to force Bazel to build some of its dependencies, in the case where
 """
 
 def _build_test_impl(ctx):
-    executable = ctx.actions.declare_file(ctx.label.name + ".sh")
+    # See https://github.com/bazelbuild/bazel-skylib/blob/1.7.1/rules/build_test.bzl#L19-L33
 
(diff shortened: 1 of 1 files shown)


alexeagle (approved):

ts/private/build_test.bzl
```diff
@@ -26,11 +29,22 @@ def _build_test_impl(ctx):
...
+        "is_windows": attr.bool(mandatory = True),
     },
     test = True,
 )
+
+def build_test(name, **kwargs):
+    _build_test(
+        name = name,
+        is_windows = select({
+            "@bazel_tools//src/conditions:host_windows": True,
```
dzbarsky, line 46: shouldn't this depend on the exec platform?

jbedard: I have no idea, I just copied it and it made Windows ci green 🤷‍♂️

dzbarsky: Well if anyone tries to use windows host with Linux RBE I guess they will hit it :)

PR #734, opened 2024-11-07
chore: remove docs and decprecate exclude_types_from_npm_packages

jbedard:
I think this should be removed in the next rules_js major release and we should always recommend `js_library` for local workspace projects. `npm_package` should only be used as a target for building ready-to-publish npm package.

---

### Test plan


jbedard:
Maybe we should discuss this in https://github.com/aspect-build/rules_js/pull/2014 first?

gregmagolan (approved):

PR #735, opened 2024-11-07
docs: change release notes to encourage bzlmod

jbedard:
The previous wording "with Bazel 6" is outdated and makes that option appear to be the "legacy old bazel 6" snippet.

---

### Test plan


alexeagle (approved):
Ideally I'd like all of ours to be downstream of whatever the rules-template creates

PR #533, opened 2024-01-30
test: assets being copied can make a conflicting asset

YvanGuidoin:
Error caused by https://github.com/aspect-build/rules_ts/pull/493

### Type of change

- Bug fix (change which fixes an issue)

### Test plan

- New test cases added

```bash
bazel test //...                                    
ERROR: file 'examples/assets/src/example-copy.json' is generated by these conflicting actions:
Label: //examples/assets:copy_example_json_file, //examples/assets:ts
RuleClass: genrule rule, ts_project rule
JavaActionClass: class com.google.devtools.build.lib.rules.genrule.GenRuleAction, class com.google.devtools.build.lib.analysis.actions.StarlarkAction
Configuration: a71759aba33917e351e9a732687d7f5603e0f139849588af46ae75a677ded0ab
Mnemonic: Genrule, CopyFile
Action key: 06b8d5d60dd630cfa8e986424d5c17575e8ac92c984005d84e01cbeb7ea6e789, 70c8c7c63cbdcb3ed9c0423f204d067dfbe1f9fa4db082fc9526e6061992ef93
Progress message: Executing genrule //examples/assets:copy_example_json_file, Copying file examples/assets/src/example-copy.json
Action describeKey: Executing genrule //examples/assets:copy_example_json_file
  Environment variable: PATH=/bin:/usr/bin:/usr/local/bin
  Argument: /bin/bash
  Argument: -c
  Argument: 'source external/bazel_tools/tools/genrule/genrule-setup.sh; cp ./examples/assets/example.json bazel-out/k8-fastbuild/bin/examples/assets/src/example-copy.json'
, Copying file examples/assets/src/example-copy.json
  Environment variable: PATH=/bin:/usr/bin:/usr/local/bin
  Argument: /bin/bash
  Argument: -c
  Argument: 'cp -f "$1" "$2"'
  Argument: ''
  Argument: bazel-out/k8-fastbuild/bin/examples/assets/src/example-copy.json
  Argument: bazel-out/k8-fastbuild/bin/examples/assets/src/example-copy.json

PrimaryInput: File:[/{...}/rules_ts[source]]examples/assets/example.json, File:[[<execution_root>]bazel-out/k8-fastbuild/bin]examples/assets/src/example-copy.json
PrimaryOutput: File:[[<execution_root>]bazel-out/k8-fastbuild/bin]examples/assets/src/example-copy.json
[...]

PR #742, opened 2024-12-02
ts_project: ensure assets get propagated

iamricard:
https://github.com/aspect-build/rules_ts/pull/533, in fixing the issue caused by https://github.com/aspect-build/rules_ts/pull/493, also omitted adding assets that shouldn't be copied to the `JsInfo` provider. As a result, downstream rules (such as `esbuild`) will fail to find the file because it never collected it.

---

### Test plan

- New test cases added

Easiest way to repro this issue is to pipe the existing test cases into an `esbuild` bundle. The target should fail with `[plugin: bazel-sandbox] Could not resolve "./generated.json"`.

### DIFF
--- a/examples/assets/BUILD.bazel
+++ b/examples/assets/BUILD.bazel
@@ -1,5 +1,6 @@
 load("@aspect_bazel_lib//lib:copy_to_directory.bzl", "copy_to_directory")
 load("@aspect_bazel_lib//lib:testing.bzl", "assert_outputs")
+load("@aspect_rules_js//js:defs.bzl", "js_test")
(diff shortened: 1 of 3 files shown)


jbedard:
Would you be able to add a test? Assuming it's simple... maybe just put a json file in the `assets` and `require` it at runtime? Could run the same test twice (once reproducing this bug, again with `outDir` to test that case) if possible?

iamricard:
> Would you be able to add a test? Assuming it's simple... maybe just put a json file in the `assets` and `require` it at runtime? Could run the same test twice (once reproducing this bug, again with `outDir` to test that case) if possible?

Done! Let me know if that's what you had in mind.

jbedard (approved):

PR #750, opened 2024-12-13
feat: eliminate distinction between `srcs` and `assets`

jfirebaugh:
I believe the distinction between `srcs` and `assets` is unnecessary and the two attributes should be combined. The desired behavior is that `ts_project` should produce at least one output file for every input file in `srcs`. Either the input should produce transpiled/declaration output(s), or it should be copied to the output directory, obeying the path rewriting rules implied by `root_dir`/`out_dir`/`declaration_dir`.

Empirically, users at Figma have found the current behavior of `srcs` confusing. A typical experience is that they put non-TS files in `srcs`, expecting them to get copied to `out_dir`, and then observed some sort of failure, often at runtime, because the file was silently ignored instead.

Some specific gotchas under the current behavior of `srcs` are:
* With `resolve_json_module = True, transpiler = "tsc"`, .json files in `srcs` that are imported by one of the other input files get copied to `out_dir` as expected. But .json files that are listed in srcs but _unreferenced_ by any input file produce an "output was not created" Bazel error. With the change proposed here, all .json files get copied to `out_dir`, regardless of whether they are imported somewhere or not.
* If you take a working `resolve_json_module = True, transpiler = "tsc"` target and change it to transpile with swc, any .json files in `srcs` will no longer get copied to `out_dir`. With the change proposed here, the behavior is consistent between tsc and swc.
* The behavior of `.d.ts` files in `srcs` is unpredictable and changes between rules_ts releases. Prior to 3.3.2, `.d.ts` files did get copied to output, but without `out_dir` applied. In rules_ts 3.3.2, they stopped being output altogether. With the change proposed here, they have a consistent and predictable behavior: they are always outputs, in the `types` output group, with `root_dir`/`out_dir`/`declaration_dir` applied.

---

This would be a breaking change because it removes the `assets` attribute. Generally, the migration action is simply to merge the `assets` value into `srcs`.

### Test plan

- New test cases added


alexeagle:
Ping @jbedard do you have a principled answer to this? I'm persuaded by @jfirebaugh argument

Aghassi:
We have a similar case to @jfirebaugh and also anecdotally having this be separate means each time we setup a new project we have to fail build file generation and go modify all the build files by hand to account for images, json files, and other assets. It's an ergonomic pain and prohibits newer users from just onboarding without hurdles.

jbedard:
My only concern is the confusion of what gets processed by ts vs what simply gets copied. Having `assets` clearly separates those (excluding the json/resolve_json_module confusion).

If we merge the two attributes then `ts_project` would need some logic to decide what is a tsc input ("sources") vs what is simply copied ("assets"). But maybe json already needs that logic due to the issues @jfirebaugh is pointing out? 🤔

jbedard:
See https://github.com/aspect-build/rules_ts/pull/750#issuecomment-2759438250

PR #754, opened 2025-01-08
fix: do not include `tsconfig.json` files in `JsInfo` types

devversion:
It's unclear to me why the `tsconfig.json` files from `ts_config` targets end up contributing to the JSInfo type declarations. Those files shouldn't be necessary for represent the output of a TS compilation in general.

I assume, they could be necessary for composite projects, but even then I'd expect the tsconfig to be retrieved in consuming targets via the dedicated `TsConfigInfo` provider.

Related: https://github.com/angular/angular-cli/pull/29282

### DIFF
--- a/ts/private/ts_config.bzl
+++ b/ts/private/ts_config.bzl
@@ -51,8 +51,7 @@ def _ts_config_impl(ctx):
             transitive_deps.append(dep[TsConfigInfo].deps)
 
     transitive_sources = js_lib_helpers.gather_transitive_sources([], ctx.attr.deps)
-
-    transitive_types = js_lib_helpers.gather_transitive_types(files, ctx.attr.deps)
+    transitive_types = js_lib_helpers.gather_transitive_types([], ctx.attr.deps)
 
     npm_sources = js_lib_helpers.gather_npm_sources(
         srcs = [],


jbedard:
While this does sound like a bug I think the `transitive_x` fields should (almost?) always include the `x` as well.

I was just looking into this exact code for a separate issue and will think about this case as well now. Can you summarize what your targets look like so I can try adding a test? Or maybe one of the tests/examples in rules_ts already reproduce this?

devversion:
> While this does sound like a bug I think the transitive_x fields should (almost?) always include the x as well.

I agree in general, but not sure this is even applicable here. X in this case is the tsconfig file and shouldn't be part of `x` nor `transitive_x`; i.e. not part of this provider at all

>  Can you summarize what your targets look like so I can try adding a test

My target is a simple packaging rule that takes all `.d.ts` and puts it into a tree artifact. i.e. publishing an npm tarball. Right now we accidentally end up with the tsconfig files in there as well

jbedard:
A `ts_config` target should contain all relevant files + packages for the tsconfig though, including the tsconfig.json, what it extends (file or package) and things like `@types/node` packages referenced by `compilerOptions.types` etc.

I think the bug is how `ts_project` propagates those dependencies.

jbedard:
See #755 which I believe fixes this

jbedard:
@devversion can you verify https://github.com/aspect-build/rules_ts/commit/ad6f86cfcef48ca7f93468466949d54e27a1fc1b solved this?

jbedard:
That might also be true actually 🤔 

Essentially https://github.com/aspect-build/rules_ts/pull/761?

PR #755, opened 2025-01-08
fix: do not include tsconfig inputs in ts_project outputs

jbedard:
Transitive tsconfig files/dependencies are propagated as their own `TsConfigInfo` dependency tree. Those tsconfig files/dependencies should never be outputs of `ts_project`.

Close #754

---

Never include tsconfig related files and dependencies in `ts_project` outputs.

### Test plan

- New test cases added

### DIFF
--- a/ts/private/ts_project.bzl
+++ b/ts/private/ts_project.bzl
@@ -37,6 +37,34 @@ def _gather_transitive_typecheck_from_output_group_infos(typecheck_outs, targets
     ]
     return depset(typecheck_outs, transitive = files_depsets)
 
+def _gather_tsconfig_deps(ctx):
+    tsconfig = copy_file_to_bin_action(ctx, ctx.file.tsconfig)
+
+    # Gather TsConfig info from both the direct (tsconfig) and indirect (extends) attribute
+    tsconfig_inputs = [tsconfig] + copy_files_to_bin_actions(ctx, ctx.files.extends)
+
+    deps = []
+
+    # Direct TsConfigInfo deps
+    if TsConfigInfo in ctx.attr.tsconfig:
+        deps.append(ctx.attr.tsconfig[TsConfigInfo].deps)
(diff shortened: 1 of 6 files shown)


ts/test/ts_config_test.bzl
```diff
@@ -0,0 +1,153 @@
...
+    target_under_test = analysistest.target_under_test(env)
+
+    jsinfo = target_under_test[JsInfo]
+
+    output_types = sorted([f.path for f in jsinfo.types.to_list()])
+    asserts.equals(env, 1, len(output_types))
+    asserts.true(env, output_types[0].find("/src_ts.d.ts") != -1)
+
+    output_transitive_types = sorted([f.path for f in jsinfo.transitive_types.to_list()])
+    asserts.equals(env, 1, len(output_transitive_types))
```
jbedard, line 140: This is the assertion that previously failed

ts/test/ts_project_test.bzl
```diff
@@ -13,6 +13,7 @@ def _dir_test_impl(ctx):
 
     # assert the inputs to the tsc action are what we expect
     action_inputs = target_under_test[OutputGroupInfo]._action_inputs.to_list()
+    action_inputs = sorted(action_inputs)
```
jbedard, line 16: The depset order should not be relied on and changed as a result of this fix. Most tests in this file already do this sorting.

ts/private/ts_project.bzl
```diff
@@ -316,9 +329,9 @@ See https://github.com/aspect-build/rules_ts/issues/361 for more details.
             },
         )
 
-    transitive_sources = js_lib_helpers.gather_transitive_sources(output_sources, srcs_tsconfig_deps)
```
jbedard, line 319: This is the main change for the fix, not gathering sources+types from the `tsconfig` target as outputs.

gregmagolan (approved):
🌮

Issue #421, opened 2023-08-23
remove TsConfigInfo


alexeagle:
### What happened?

There's a TODO

https://github.com/aspect-build/rules_ts/blob/71ecd6343a7e3d4af842ebe1b42b7b0aa2d1acd0/ts/private/ts_config.bzl#L45-L46

Maybe we can make it TODONE for 2.0

### Version

HEAD

alexeagle:
I cut the 2.0 RC without this, so now it has to wait for the next major.

PR #761, opened 2025-01-10
refactor: move ts_config src to outputted JsInfo.sources

jbedard:
The `JsInfo.[transitive_]sources` from `ts_config` should contain the sources required to compile, while `JsInfo.[transitive_]types` should contain any additional files required to type-check.

---

### Test plan

### DIFF
--- a/ts/private/ts_config.bzl
+++ b/ts/private/ts_config.bzl
@@ -50,9 +50,9 @@ def _ts_config_impl(ctx):
         if TsConfigInfo in dep:
             transitive_deps.append(dep[TsConfigInfo].deps)
 
-    transitive_sources = js_lib_helpers.gather_transitive_sources([], ctx.attr.deps)
+    transitive_sources = js_lib_helpers.gather_transitive_sources(files, ctx.attr.deps)
 
-    transitive_types = js_lib_helpers.gather_transitive_types(files, ctx.attr.deps)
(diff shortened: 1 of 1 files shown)


devversion (approved):
💯

ts/private/ts_config.bzl
```diff
@@ -50,9 +50,9 @@ def _ts_config_impl(ctx):
         if TsConfigInfo in dep:
             transitive_deps.append(dep[TsConfigInfo].deps)
 
-    transitive_sources = js_lib_helpers.gather_transitive_sources([], ctx.attr.deps)
+    transitive_sources = js_lib_helpers.gather_transitive_sources(files, ctx.attr.deps)
```
devversion, line 53: I would even consider dropping this completely and only exposing things in `JsInfo` that would be needed for compilation. The tsconfig may be needed in composite projects, yeah, but for that you have the whole concept of the `TsConfigInfo` provider?

jbedard: A `ts_config` target needs to expose a tsconfig file and all it's dependencies (more tsconfig files/targets such as `extends` or `composite`, sometimes npm packages referenced by things like `compilerOptions.types` etc).

Today that is awkwardly duplicated in `TsConfigInfo` and `JsInfo` with a TODO to potentially drop the extra `TsConfigInfo` (#421). IMO we should drop one or the other from here and remove the confusing duplication 🤷

dzbarsky (approved):

PR #765, opened 2025-01-21
feat: add generate_trace option to run TypeScript with `--generateTrace`

ericrafalovsky:
This pull request adds a new feature for generating trace files to analyze TypeScript compiler performance (`tsc --generateTrace`)

---

Allow generating a TypeScript compiler trace (`tsc --generateTrace`)

### Test plan

- New basic test cases added
- Manual testing of this change in a private Bazel repo


ts/private/ts_project.bzl
```diff
@@ -196,6 +196,11 @@ See https://github.com/aspect-build/rules_ts/issues/361 for more details.
         common_args.extend(["--tsBuildInfoFile", to_output_relative_path(ctx.outputs.buildinfo_out)])
         outputs.append(ctx.outputs.buildinfo_out)
 
+    tsc_trace_dir = None
+    if options.generate_tsc_trace or ctx.attr.generate_trace:
+        tsc_trace_dir = ctx.actions.declare_directory(ctx.attr.name + "_trace")
+        common_args.extend(["--generateTrace", to_output_relative_path(tsc_trace_dir)])
```
jbedard, line 202: Does putting it into `common_args` mean two actions will both be outputting this file if `ctx.attr.isolated_typecheck and (use_tsc_for_js or use_tsc_for_dts)`?

I think if we are invoking tsc twice (once for transpiling and once for type-checking) both invocations will try to output the same file...

jbedard (approved):

PR #766, opened 2025-01-22
fix: ensure ts_project _types target has the right visibility

menny:
If this is not passed, the _types target will get the package's default visibility.

In out case, we have a package with a very restrictive visibility, but one of the targets (`ts_project`) in the package is  explicitly set to `public`. We want to depend on the `_types` target, but that is not visible due the default visibility.
NOTE: This is relevant only when an `dts` emitter is provided.

---

### Test plan
Create a target with a `dts` emitter. Ensure the `_types` target gets the same visibility of the main target.

### DIFF
--- a/ts/defs.bzl
+++ b/ts/defs.bzl
@@ -385,6 +385,7 @@ def ts_project(
             native.alias(
                 name = types_target_name,
                 actual = declarations_target_name,
+                visibility = common_kwargs.get("visibility"),
             )
         else:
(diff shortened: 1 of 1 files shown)


jbedard (approved):

jbedard:
Thanks for the fix 👍

menny:
wow. that was quick :)

jbedard:
The notification basically popped up under my mouse, and lgtm lol 👍

PR #775, opened 2025-03-04
chore: remove workspace (non-bzlmod) unit testing

jbedard:
WORKSPACE is still tested in e2e tests, but development of rules_js is now only bzlmod and unit tests are only run with bzlmod.

### Test plan


jbedard:
This also requires the rules_lint update so moving to https://github.com/aspect-build/rules_ts/pull/774

PR #776, opened 2025-03-10
Disallow remote cache with workers

MichaelMitchell-at:
This is a patch that we've had at Airtable for a while, but didn't get around to upstreaming until now.

### Changes are visible to end-users: no

Doubt this would be a visible change by most.

### Test plan


jbedard:
Thanks @MichaelMitchell-at! 👍 

Have a sec to fix the buildifier error?

jbedard (approved):

MichaelMitchell-at:
> Thanks @MichaelMitchell-at! 👍
> 
> Have a sec to fix the buildifier error?

Yep, done!

jbedard (approved):

PR #782, opened 2025-03-13
fix: isolated_typecheck and transitive tsconfig deps

### LINKED ISSUE #781: [Bug]: isolated_typecheck does not include transitive ts_config deps when tsc invoked
`ts_project(isolated_typecheck)` reduces the inputs to tsc transpiling, including removal of `ts_config(deps)` which tsconfig `extends` or `types` might require.

jbedard:
Fix #781

---

Include tsconfig deps when invoking tsc even when `ts_project(isolated_typecheck)`.

### Test plan

- New test cases added

### DIFF
--- a/ts/private/ts_project.bzl
+++ b/ts/private/ts_project.bzl
@@ -51,6 +51,8 @@ def _gather_tsconfig_deps(ctx):
     if ctx.attr.extends and TsConfigInfo in ctx.attr.extends:
         deps.append(ctx.attr.extends[TsConfigInfo].deps)
 
+    deps.append(_gather_types_from_js_infos([ctx.attr.tsconfig]))
+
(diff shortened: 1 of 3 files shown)


ts/private/ts_project.bzl
```diff
@@ -84,6 +86,15 @@ def _ts_project_impl(ctx):
 
     srcs = [_lib.relative_to_package(src.path, ctx) for src in srcs_inputs]
 
+    srcs_deps = ctx.attr.srcs + ctx.attr.deps
+
+    tsc_inputs = copy_files_to_bin_actions(ctx, srcs_inputs + tsconfig_inputs)
```
jbedard, line 91: I think we can probably still improve the readability of this a bit. Let me know if you have any new ideas/names etc.

Issue #772, opened 2025-02-15
[FR]: Support depending on ts_projects in other repositories


aschleck:
### What is the current behavior?

In an example where target `//:lib` depends on `@dev_april_corgi//js/common`, we see errors like:

```
ERROR: /private/var/tmp/_bazel_april/2fd9b4127fda8e5b6f99113937b79956/external/dev_april_corgi+/js/common/BUILD:5:13: in ts_project rule @@dev_april_corgi+//js/common:common: 
Traceback (most recent call last):
	File "/private/var/tmp/_bazel_april/2fd9b4127fda8e5b6f99113937b79956/external/aspect_rules_ts+/ts/private/ts_project.bzl", line 188, column 45, in _ts_project_impl
		asset = ctx.actions.declare_file(a_out)
Error in declare_file: the output artifact 'external/dev_april_corgi+/js/dev_april_corgi+/js/common/arrays.ts' is not under package directory 'external/dev_april_corgi+/js/common' for target '@@dev_april_corgi+//js/common:common'
```

You can see this by cloning https://github.com/aschleck/external_ts_example/tree/main and removing this patch line: https://github.com/aschleck/external_ts_example/blob/main/MODULE.bazel#L8 . Including the patch makes this compilation succeed.

### Describe the feature

TypeScript is generally distributed via npm, but I have some code that I wrote (predominantly https://github.com/aschleck/corgi) and am using in several of my own projects that is not worth pushing to npm and is not ready for public consumption. The easiest thing for me to do is to just depend on it as a Bazel module from all my projects, however `ts_project` rules in dependency repositories fail to build from the parent project.

I have found a super simple apparent fix (https://github.com/aschleck/external_ts_example/blob/main/aspect_rules_ts.patch) which works in my testing. The only important thing in my experience is that the dependency project must use relative imports when referring to code in its own project (for example https://github.com/aschleck/corgi/blob/main/js/corgi/binder.ts#L1)

Is this something that could be supported in rules_ts directly? I believe my patch is safe but am not certain if there are other considerations I'm missing

```
diff -Naur ts/private/ts_lib.bzl ts/private/ts_lib.bzl
--- ts/private/ts_lib.bzl
+++ ts/private/ts_lib.bzl
@@ -197,7 +197,7 @@ def _join(*elements):
 def _relative_to_package(path, ctx):
     path = path.removeprefix(ctx.bin_dir.path + "/")
-    path = path.removeprefix("external/")
[...]

PR #784, opened 2025-03-18
fix: assets from ts_projects in other repositories

Mivr:
Fix https://github.com/aspect-build/rules_ts/issues/772

---

### Test plan

- New test cases added


jbedard (approved):

aschleck:
Thank you for fixing on this! I was totally lost

PR #786, opened 2025-03-21
Resolve `@npm_typescript` relative to `rules_ts`

fmeum:
Fixes an error with Bzlmod if the main module doesn't `use_repo` `npm_typescript`.

---

### Test plan

- Manual testing (don't `use_repo` `npm_typescript`)


alexeagle (approved):
THanks!

fmeum:
This fix can't work as is since the `rules_ts_ext` usage is a dev dependency only and even if it wasn't, it wouldn't be clear which of the repos should be picked. This would either require limiting the flexibility in the extension (preventing the user from specifying a custom name and instantiating a single repo only) or switching to a toolchain setup that doesn't require referencing repos by name.

PR #796, opened 2025-04-05
refactor: don't build protoc in docs either, it's the slowest CI step

alexeagle:
follow-up to #792 

I made an incredibly simple GHA at https://github.com/alexeagle/stardoc-prebuilt to give the JAR file we need


jbedard (approved):

jbedard:
Can you verify all this protoc avoidance will still work with bazel 8? I just want to make sure we won't have to revert all this.

alexeagle:
Yes, toolchains_protoc is built and tested with Bazel 8 already, and the docs/ folder here is now on Bazel 8 too

PR #804, opened 2025-04-29
fix: source dts files outputted to outDir

jbedard:
https://github.com/aspect-build/rules_ts/pull/746/ changed this logic to align with outputted .js files which I think makes sense. However js also has the `allowJs` logic while dts files do not.

---

### Test plan

- New test cases added

### DIFF
--- a/ts/private/ts_project.bzl
+++ b/ts/private/ts_project.bzl
@@ -224,8 +224,19 @@ See https://github.com/aspect-build/rules_ts/issues/361 for more details.
         for s in srcs_inputs:
             if _lib.is_js_src(s.path, ctx.attr.allow_js, ctx.attr.resolve_json_module):
                 output_sources.append(s)
-            if _lib.is_typings_src(s.path):
+            elif _lib.is_typings_src(s.path):
                 output_types.append(s)
+    else:
+        # Add DTS inputs that are not transpiled by tsc yet should be copied to the out_dir
+        # See https://github.com/microsoft/TypeScript/issues/39231 for tsc request to add this feature
+        for s in srcs_inputs:
+            if _lib.is_typings_src(s.path):
+                dts_path = _lib.relative_to_package(s.path, ctx)
+                dts_out_path = _lib.to_out_path(dts_path, ctx.attr.out_dir, ctx.attr.root_dir)
+                if s.is_source or dts_path != dts_out_path:
+                    dts_out = ctx.actions.declare_file(dts_out_path)
+                    copy_file_action(ctx, s, dts_out)
+                    output_types.append(dts_out)
 
     # Make sure the user has acknowledged that transpiling is slow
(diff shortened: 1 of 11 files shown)


ts/private/ts_project.bzl
```diff
@@ -224,8 +224,19 @@ See https://github.com/aspect-build/rules_ts/issues/361 for more details.
         for s in srcs_inputs:
             if _lib.is_js_src(s.path, ctx.attr.allow_js, ctx.attr.resolve_json_module):
                 output_sources.append(s)
-            if _lib.is_typings_src(s.path):
+            elif _lib.is_typings_src(s.path):
                 output_types.append(s)
+    else:
+        # Add DTS inputs that are not transpiled by tsc yet should be copied to the out_dir
+        # See https://github.com/microsoft/TypeScript/issues/39231 for tsc request to add this feature
```
jbedard, line 231: If tsc doesn't support this... should rules_ts? Overall I feel like this workflow makes sense so it's odd tsc seems to not support it. Does the use of ts within bazel make this issue more common which would justify adding this feature despite tsc not supporting it? (imo the answer is yes)

As I understand it the scenario would be:
```
src/
   a.ts
   types.d.ts
```

Then `ts_project(root_dir="src", out_dir="dist")`.

If the types within `types.d.ts` are public then the `ts_project` output really does need them (because `a.d.ts` probably imports `types.d.ts`).

alexeagle: Generally the principle here is that we behave the same as `tsc` as much as possible.
If someone needs to do this, can't they do it in userspace?

alexeagle: That's a lot of "probably" and asking questions, are you asking me to answer those questions?

jbedard: I'm looking for feedback and opinions because I'm uncertain about this change. I don't like adding functionality, but I don't think such a standard use case should require an ugly workaround in userspace.

This also partially (but incorrectly) worked in the past before https://github.com/aspect-build/rules_ts/pull/746, and it still (correctly) works if `out_dir == root_dir`.

jbedard: So tsc explicitly doesn't not copy `.d.ts` files, unlike `.js` when `allowJs`. This was intentional from the beginning: https://github.com/microsoft/TypeScript/issues/5112#issuecomment-145633791

Outside bazel this normally isn't an issue because a) no sandboxing (so you can access the `.d.ts` even if it isn't outputted to `outDir`) and b) pnpm symlinks directly into the source for references between projects so these `.d.ts` are available there as well.

In bazel the non-copied `.d.ts` are not accessible at all: they are not in the sandbox, they are not in npm packages. Yet they would be copied if in `js_library(srcs)` which makes this even more odd...?

jbedard:
Lets only do this if https://github.com/microsoft/TypeScript/issues/39231 is done

PR #811, opened 2025-05-30
fix: avoid accidental re-use of incorrect cached versions

devversion:
Prevents accidental re-use of cached versions that would otherwise be used purely based on the "integrity" value. E.g. someone forgot to update the integrity but the `ts_version` is already different.

### DIFF
--- a/ts/private/npm_repositories.bzl
+++ b/ts/private/npm_repositories.bzl
@@ -42,9 +42,15 @@ def _http_archive_version_impl(rctx):
             If this is a semver range you must specify an exact version instead.
             See documentation on rules_ts_dependencies.""".format(version))
 
+    urls = [u.format(version) for u in rctx.attr.urls]
+
     rctx.download_and_extract(
-        url = [u.format(version) for u in rctx.attr.urls],
+        url = urls,
         integrity = integrity,
+        # Prevents accidental re-use of cached versions that would otherwise
+        # be used purely based on the "integrity" value. E.g. someone forgot
+        # to update the integrity but the `ts_version` is already different.
+        canonical_id = get_default_canonical_id(rctx, urls),
     )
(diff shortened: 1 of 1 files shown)


ts/private/npm_repositories.bzl
```diff
@@ -45,6 +45,10 @@ def _http_archive_version_impl(rctx):
     rctx.download_and_extract(
         url = [u.format(version) for u in rctx.attr.urls],
         integrity = integrity,
+        # Prevents accidental re-use of cached versions that would otherwise
+        # be used purely based on the "integrity" value. E.g. someone forgot
+        # to update the integrity but the `ts_version` is already different.
+        canonical_id = "%s_%s" % (version, integrity),
```
jbedard, line 51: Should we use https://bazel.build/rules/lib/repo/cache#get_default_canonical_id instead of inventing our own scheme?

jbedard: Should we at least use the same implementation as [get_default_canonical_id](https://bazel.build/rules/lib/repo/cache#get_default_canonical_id)? Also put a TODO to use it once we drop bazel6...

jbedard: > You might also ask, why does Bazel not call this default helper by default.

That's a really good question 😢

jbedard: But I'm just looking at `get_default_canonical_id` as an example of what `canonical_id` should normally be used for.

I think we basically want to invalidate the cache if the URL changes, even if the `integrity` did not change? That's also what `get_default_canonical_id` does... 🤷

devversion: I'm good using that, so I'll try switch to it;

jbedard (approved):

PR #812, opened 2025-06-06
ts_project_worker: default-initialize tsc outDir

malt3:
Some time ago, the tsc_project macro changed behavior. Before, `out_dir` would always be set, whereas now it may be `None`. If `out_dir` is unset, this expression will result in an incorrect output directory being used:

```
host.optionsToExtend.outDir = `${host.optionsToExtend.outDir}/${SYNTHETIC_OUTDIR}`;
```

(Would result in `'undefined/__st_outdir__'`).

I understand that the persistent worker mode is mostly unused in upstream rules_ts (since it is also disabled by default on TS 5), so it is not surprising to see small breakages.
I still hope we can fix the worker here, and don't have to carry downstream patches.

---

### Test plan

Use a custom ts_worker and ensure you have a ts_project with `outDir` unset (or set to `"."`).


ts/private/ts_project_worker.js
```diff
@@ -704,7 +704,7 @@ function createProgram(args, inputs, output, exit) {
     }
 
     function applySyntheticOutPaths() {
-        host.optionsToExtend.outDir = `${host.optionsToExtend.outDir}/${SYNTHETIC_OUTDIR}`;
+        host.optionsToExtend.outDir = `${host.optionsToExtend.outDir || host.optionsToExtend.rootDir || '.'}/${SYNTHETIC_OUTDIR}`;
```
jbedard, line 707: Why use `rootDir`? That seems wrong imo... 🤔

jbedard: Or is this `rootDir` different then `compilerOptions.rootDir`?

malt3: Yes, `outDir || BUILD-dir` would be ideal. It's just that the long-lived worker process can only infer the BUILD directory from the work request. Right now this doesn't contain the BUILD directory. I just chose root rootDir since it happens to match BUILD-dir by default.

I can add a test, but more importantly: should we add the current package (aka `BUILD-dir`) to the work request explicitly instead of misusing rootDir?

jbedard: Note that I don't know this worker code well and I'm going off assumptions here. So maybe this `SYNTHETIC_OUTDIR` logic is different then what I'm assuming.

Can we simply avoid this `SYNTHETIC_OUTDIR` thing when `outDir` is not configured?

PR #825, opened 2025-07-05
fix: update build_test() fork for compatibility with Bazel 8.3.0

### LINKED ISSUE #824: [Bug]: _build_test rule depends on deprecated target '@@bazel_tools//src/conditions:host_windows_x64_constraint
### What happened?

The build_test macro in aspect_rules_ts calls the _build_test rule with a select involving `@bazel_tools//src/conditions:host_windows`.  Bazel 8.3.0 warns about this:
```
target '<your-target-here>' depends on deprecated target '@@bazel_tools//src/conditions:host_windows_x64_constraint': No longer used by Bazel and will be removed in the future. Migrate to toolchains or define your own version of this setting.
```
It appears that this code in aspect_rules_ts is forked from bazel_skylib.  Note that release 1.8.0 of bazel_skylib corrects this issue.

### Version

Development (host) and target OS/architectures: Ubuntu 24.04 / amd64

Output of `bazel --version`: bazel 8.3.1

`WORKSPACE` or `MODULE.bazel` file: 3.5.0

### How to reproduce

```shell
Use ts_project() in conjunction with Bazel 8.3.1.  It creates a build_test() target and this leads to the warning.
```

### Any other information?

Though we are currently using rules_ts 3.5.0, I checked the code in your main branch and it still uses `@bazel_tools//src/conditions:host_windows`.

dws:
We here imitate the changes in bazel_skylib 1.8.0 to avoid referring to `@bazel_tools//src/conditions:host_windows`.  As of Bazel 8.3.0 referring to `@bazel_tools//src/conditions:host_windows` yields a deprecation warning.

In concert with this, we here bump the minimum required version of bazel_skylib to 1.8.0.  Per the comments in ts/repositories.bzl this may be a BREAKING CHANGE.

Fixes #824

---

### Test plan

- New test cases added


fmeum:
The error message with Bazel 6.5.0 is unexpected and doesn't make sense (it mentions targets in the same package). The same kind of issue doesn't appear in skylib's own tests.

I sent https://github.com/bazelbuild/bazel-skylib/pull/584, could you check whether that fixes the problem?

dws:
> This seems to break bazel 6 though?
> 
> Do you have a link to the change in bazel_skylib that we can compare to? Maybe that handles bazel6 still?

@jbedard I'm here imitating the changes to `rules/build_test.bzl` in https://github.com/bazelbuild/bazel-skylib/commit/8ba7f197a9acd67eb0921dec6f8e8e24645431b8

dws:
> The error message with Bazel 6.5.0 is unexpected and doesn't make sense (it mentions targets in the same package). The same kind of issue doesn't appear in skylib's own tests.
> 
> I sent [bazelbuild/bazel-skylib#584](https://github.com/bazelbuild/bazel-skylib/pull/584), could you check whether that fixes the problem?

@fmeum I added an experimental commit here that patches your https://github.com/bazelbuild/bazel-skylib/pull/584 into the bazel_skylib used here, and that does seem to resolve the problem seen with Bazel 6.5.0.

dws:
> This seems to break bazel 6 though?
> 
> Do you have a link to the change in bazel_skylib that we can compare to? Maybe that handles bazel6 still?

@jbedard The fixes in bazel_skylib 1.8.0 for `build_test()` by themselves pass CI here.  So, I've gotten CI to pass here by simply dropping the version bump to bazel_skylib.  We can thus consider that version bump separately if/when bazel_skylib has a new release that includes https://github.com/bazelbuild/bazel-skylib/pull/584 from @fmeum.

fmeum:
The 1.8.1 release is out now.

PR #832, opened 2025-07-18
chore: enable aspect workflows for more Bazel versions

alexeagle:
After this lands, I'll remove some GHA testing that overlaps.


jbedard (approved):
Can we also delete these from github ci?

PR #835, opened 2025-07-25
docs: remove resolved known issues from README

michael-pplx:
:)


jbedard (approved):
Thanks for both noticing and fixing this! 👍

PR #868, opened 2025-12-10
feat: support bazel path-mapping

jbedard:
Ref https://github.com/aspect-build/rules_js/issues/2574

Support bazel path mapping across compilation modes etc.

### Test plan

### DIFF
--- a/ts/private/ts_proto_library.bzl
+++ b/ts/private/ts_proto_library.bzl
@@ -131,6 +131,9 @@ def _protoc_action(ctx, proto_info, outputs):
         ),
         env = {"BAZEL_BINDIR": ctx.bin_dir.path},
         use_default_shell_env = True,
+        execution_requirements = {
+            "supports-path-mapping": "1",
+        },
     )
 
 def _declare_outs(ctx, info, ext):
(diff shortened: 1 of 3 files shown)


ts/private/ts_proto_library.bzl
```diff
@@ -131,6 +131,9 @@ def _protoc_action(ctx, proto_info, outputs):
         ),
         env = {"BAZEL_BINDIR": ctx.bin_dir.path},
```
dzbarsky: hmm can you check with `-s` what gets sent here? I'm not convinced this `.path` won't break it

dzbarsky (approved):
did you check that `to_output_relative_path` works as we expect?

ts/private/ts_proto_library.bzl
jbedard: Does that mean this PR has essentially no effect?

dzbarsky: yes, i believe so

jbedard:
Blocked on https://github.com/aspect-build/rules_js/issues/2581

jbedard:
Done elsewhere/differently

PR #869, opened 2025-12-10
chore: remove bazelisk aspect-cli config

jbedard:
Let users decide if/when to use aspect-cli.

### Test plan


alexeagle:
could you do a text search as well? for example https://github.com/aspect-build/rules_ts/blob/main/.github/workflows/ci.yaml#L169

alexeagle (approved):
awesome deletes

PR #877, opened 2026-01-13
refactor: upgrade bzlmod aspect_bazel_lib v2 to bazel-lib v3

jbedard:
---

### Test plan

- New test cases added


jbedard:
This won't be possible until rules_ts drops bazel 6 and <7.6 so `repo_name = None` can be used, or we explicitly keep the `aspect_bazel_lib` v2 dependency to ensure a minimum version that is compatible with v3

PR #887, opened 2026-01-23
chore: move all linux testing to aspect workflows

alexeagle:


alexeagle:
Yes, intentional - but I see that it's breaking as they're not published to the registry. I'll restore for now and add a broken link checker here

PR #898, opened 2026-02-26
Add an example where the srcs of a ts_project is a directory of generated code

MatrixFrog:
I believe this might be a bug, or at least a surprising behavior. If you pass as the `srcs` of a `ts_project`, a directory of generated TS code, without specifying the exact filenames of all the TS files inside it, you get a target that essentially does nothing:

```sh
$ cd examples
$ bazel build //gen_srcs:ts_proj_typecheck_test
$ cat bazel-bin/gen_srcs/ts_proj_typecheck_test.sh # The entire test is just `exit 0`
```

Instead, it would be great if we got a warning or error at build-time: `Input target ":code_generation" is an opaque directory and will be ignored` or something like that.

### DIFF
--- a/e2e/test/dir_as_source.bats
+++ b/e2e/test/dir_as_source.bats
@@ -1,0 +1,32 @@
+load "common.bats"
+
+setup() {
+	cd $BATS_FILE_TMPDIR
+}
+
+teardown() {
+	bazel shutdown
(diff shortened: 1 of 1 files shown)


jbedard:
Actually, sorry the `test.sh` thing doesn't exist in this repo (unlike others like rules_js), but rules_ts has something better... checkout e2e/test and the bats tests.

PR #915, opened 2026-04-09
test: fix examples module with Bazel 7

acozzette:
Before this change, Bazel 7 could not build under examples/ due to the use of `ignore_directories()` in REPO.bazel. This change fixes the problem by deleting that file and going back to the older .bazelignore solution instead. I also updated the CI config to start testing the examples with Bazel 7.

---

### Test plan

### DIFF
--- a/.aspect/workflows/config.yaml
+++ b/.aspect/workflows/config.yaml
@@ -9,9 +9,8 @@ workspaces:
                   without: true
     examples:
         tasks:
-            # No Bazel7 in root/examples
             - bazel-7:
-                  without: true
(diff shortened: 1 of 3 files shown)


jbedard:
We still need to test bazel7 so I'd rather change the examples to work with bazel7...

acozzette:
That sounds good, I updated my PR to do that instead. I'm not 100% sure if I made the right change to the Aspect workflow config, but hopefully I'll find out in a few minutes once the tests run.

jbedard (approved):

PR #919, opened 2026-04-26
test: add tests and examples for package.json support in ts_config

kirkobyte:
Adds tests and examples validating that `package.json` can be provided to rules via `ts_config(deps)`, this is important for TS module type detection behavior.

Typescript's module format detection behavior for a given file:
https://www.typescriptlang.org/docs/handbook/modules/theory.html#module-format-detection
> the format of each file is determined by its file extension and the type field of the first package.json file found in a search of the file’s directory and all ancestor directories

This makes it important that `ts_config` can propagate `package.json`, and that an existing `ts_config` can be extended with `package.json` in a subdirectory as needed.

---

### Test plan

- This PR only adds tests to validate behavior


examples/package_json_usage/colocated/BUILD.bazel
```diff
@@ -0,0 +1,18 @@
...
+
+# Demonstrates that a package.json can be passed to ts_config via `deps`,
+# making it visible to tsc alongside the tsconfig.json. With
+# `verbatimModuleSyntax: true` and `module: "nodenext"`, TypeScript needs to
+# see a sibling package.json with `"type": "module"` to classify index.ts as
+# ESM — otherwise the `export` statement fails to type-check.
+ts_config(
+    name = "tsconfig",
+    src = "tsconfig.json",
+    deps = ["package.json"],
```
jbedard, line 13: So commenting this out should cause a test to fail.. does it?

ts/test/ts_config_test.bzl
```diff
@@ -151,3 +190,50 @@ def _ts_project_outputs_only_srcs_types_test_impl(ctx):
     return analysistest.end(env)
 
 _ts_project_outputs_only_srcs_types_test = analysistest.make(_ts_project_outputs_only_srcs_types_test_impl)
+
+def _ts_config_propagates_package_json_dep_test_impl(ctx):
```
jbedard: This really just asserts that `ts_config(deps)` are available via `target_under_test[TsConfigInfo].deps` - is there any value to that?

ts/test/ts_config_test.bzl
```diff
@@ -151,3 +190,50 @@ def _ts_project_outputs_only_srcs_types_test_impl(ctx):
...
+        env,
+        any([p.endswith("/package.json") for p in dep_paths]),
+        "expected package.json to be propagated through TsConfigInfo.deps, got: {}".format(dep_paths),
+    )
+
+    return analysistest.end(env)
+
+_ts_config_propagates_package_json_dep_test = analysistest.make(_ts_config_propagates_package_json_dep_test_impl)
+
+def _ts_config_wrapping_ts_config_test_impl(ctx):
```
jbedard: IDK if this test really adds much value either?

examples/package_json_usage/colocated/BUILD.bazel
jbedard: Something other then the 2 analysis tests that is. Can we have a `js_test` that does `import 'index.mjs'` or something like that to ensure `.mjs` was outputted? Is that what we're verifying?

kirkobyte: Yeah, I should test this. I think I got a bit turned around from the test file in this project, and got focused on testing the file propagation

ts/test/ts_config_test.bzl
kirkobyte: I thought that was one of the things we wanted to validate here 😅 but happy to omit if you think it's unhelpful

jbedard: Well it sort of is, but it's basically just doing `x = [y]; assert y in x`? That doesn't really show _why_ and doesn't add enough value to be worth the code imo.

If I think it would be better to test at the `ts_project(tsconfig)` level to assert that the `package.json` makes it to the tsc sandbox. I assume that is easiest as just a `ts_proejct(tsconfig)` in a BUILD and then asserting that the outputs are correct, but if you find a more direct way of testing it with an analysis test that could also be valuable.

ts/test/ts_config_test.bzl
```diff
@@ -151,3 +223,37 @@ def _ts_project_outputs_only_srcs_types_test_impl(ctx):
     return analysistest.end(env)
 
 _ts_project_outputs_only_srcs_types_test = analysistest.make(_ts_project_outputs_only_srcs_types_test_impl)
+
+def _ts_project_sees_package_json_in_tsc_inputs_test_impl(ctx):
```
jbedard, line 227: This is better and worth keeping IMO 👍

examples/linked_lib/tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
     "compilerOptions": {
-        "declaration": true
+        "declaration": true,
+        "types": ["node"]
```
jbedard: Why change this example?

kirkobyte: I was testing out things in the other packages, and found that the `.ts` file in this one had an error because without `node`, `require` can't be resolved. 

There's also no `build_test` in this folder, which might be why it wasn't picked up?

jbedard: But `node_modules/@types/node` is in the `ts_project(deps)` so this line shouldn't be required. By default `node_modules/@types/*` should be included and this tsconfig option just restricts that glob but should never be required (and it's just duplicating info already in `ts_project(deps)` so I generally don't recommend it with bazel).

I think this directory is just for the `pkg` target so it gets consumed in a `:.aspect_rules_js/node_modules/...` target somewhere and this directory itself shouldn't need any test.

What were you doing that caused an error?

kirkobyte: It was in the IDE that there was an error: 

Image

I just ran the build now and I see that it does pass, which I didn't quite expect 😅 Does `ts_project` add entries from `deps` into the ts `types` option? 🤔

jbedard: The tsconfig `types` is completely optional and is basically just a filter of `node_modules/@types/*`. Unspecified means "use all `@types/*` packages".

With bazel sandboxing imo the tsconfig `types` is pointless because bazel sandboxing and precise `ts_projecct(deps)` is already precise enough and needs no more filtering...

jbedard (approved):

jbedard:
@codex

PR #936, opened 2026-06-15
fix: include js_outs in transitive_typecheck when declaration=False

### LINKED ISSUE #935: [Bug]: _transitive_typecheck silently skips type-checking for targets with declaration = False
### What happened?

When a ts_project target has `declaration = False` building its `_transitive_typecheck` target does nothing — no type-checking is performed for that target or anything downstream of it.

Root cause: The transitive_typecheck output group is built from typecheck_outs, which is populated either by a dedicated `--noEmit` action (when tsc isn't emitting at all) or by the dts outputs (when tsc emits dts files). When `declaration = False`, tsc emits only js files  so typecheck_outs ends up empty. The `transitive_typecheck` depset has nothing to pull on, meaning the tsc action is never triggered.

### Version

Output of `bazel --version`: 8.7.0

`WORKSPACE` or `MODULE.bazel` file: `MODULE.bazel`

Language(s) and/or frameworks involved: typescript

### How to reproduce

```shell
In a dependency chain A → B → C where B has declaration = False, building A_transitive_typecheck skips B entirely. If all targets in a chain have declaration = False, _transitive_typecheck is a no-op across the whole chain.
```

### Any other information?

_No response_

menny:
When a ts_project has declaration=False and no isolated_typecheck, typecheck_outs was left empty, causing _transitive_typecheck to produce no artifacts and silently skip type-checking for that target and anything downstream of it.

Fix by falling back to js_outs as typecheck evidence when output_types is empty — the js emit action performs a full type-check as a side effect, making js_outs a valid proof that tsc ran successfully.

Fixes: https://github.com/aspect-build/rules_ts/issues/935

### Test plan
adds two analysis tests covering the declaration=False path: one verifying js_outs appear in transitive_typecheck for a single target, and one verifying the chain propagates correctly through deps.

### DIFF
--- a/ts/private/ts_project.bzl
+++ b/ts/private/ts_project.bzl
@@ -301,7 +301,8 @@ See https://github.com/aspect-build/rules_ts/issues/361 for more details.
             env = env,
         )
     else:
-        typecheck_outs.extend(output_types)
+        # When tsc emits js but no dts, js_outs are the only artifact proving tsc ran and type-checked.
+        typecheck_outs.extend(output_types if output_types else js_outs)
 
     if use_tsc_for_js or use_tsc_for_dts:
         tsc_emit_arguments = ctx.actions.args()
--- a/ts/test/ts_project_test.bzl
+++ b/ts/test/ts_project_test.bzl
@@ -109,6 +109,39 @@ _TSCONFIG = {
     },
 }
 
+_TSCONFIG_JS_ONLY = {
+    "compilerOptions": {
+        "declaration": False,
+    },
+}
+
+# no_decl_lib_transitive_typecheck_test
+# Verifies that when declaration=False, js_outs appear in transitive_typecheck (the fix).
+def _no_decl_lib_transitive_typecheck_test_impl(ctx):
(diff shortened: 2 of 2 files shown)


DeepakG123 (approved):
LGTM!

ts/private/ts_project.bzl
```diff
@@ -301,7 +301,8 @@ See https://github.com/aspect-build/rules_ts/issues/361 for more details.
             env = env,
         )
     else:
-        typecheck_outs.extend(output_types)
+        # When tsc emits js but no dts, js_outs are the only artifact proving tsc ran and type-checked.
+        typecheck_outs.extend(output_types if output_types else js_outs)
```
jbedard, line 305: Does the `if ctx.attr.isolated_typecheck or not (use_tsc_for_js or use_tsc_for_dts):` path up above already handle the scenario where we have `declarations = False, transpiler = swc`? Meaning this line you're changing is already guaranteed that tsc is producing the `js_outs`? Does a test cover that?

jbedard (approved):

PR #937, opened 2026-06-27
feat: support TypeScript 7 RC

longlho:
Summary
- handle native TypeScript versions that ship platform-specific triple packages like @typescript/typescript-darwin-arm64
- keep native TS prerelease metadata separate from TOOL_VERSIONS so LATEST_TYPESCRIPT_VERSION stays stable-only
- link TS 7 native packages into the generated npm repo
- add the TS 7 RC fixture alongside the existing 5.6 RC fixture in bzlmod and WORKSPACE

Platform note
- locally tested on macOS darwin-arm64
- CI should exercise Linux x64 for e2e/external_dep via the Aspect Workflows matrix
- existing CI does not run this e2e workspace on Windows

Test plan
- ./ts/private/mirror_versions.sh
- buildifier ts/private/versions.bzl ts/private/npm_repositories.bzl ts/repositories.bzl ts/BUILD.typescript e2e/external_dep/MODULE.bazel e2e/external_dep/WORKSPACE e2e/external_dep/BUILD
- bazel test //... (from e2e/external_dep)
- bazel test --enable_bzlmod=false //... (from e2e/external_dep)
- bazel test //ts/...

### DIFF
--- a/e2e/external_dep/MODULE.bazel
+++ b/e2e/external_dep/MODULE.bazel
@@ -25,7 +25,13 @@ rules_ts_ext.deps(
     ts_integrity = "sha512-E3b2+1zEFu84jB0YQi9BORDjz9+jGbwwy1Zi3G0LUNw7a7cePUrHMRNy8aPh53nXpkFGVHSxIZo5vKTfYaFiBQ==",
     ts_version = "5.6.1-rc",
 )
-use_repo(rules_ts_ext, "npm_typescript", "npm_typescript-56rc", "npm_typescript3")
+
+# Example of a native rc version
+rules_ts_ext.deps(
+    name = "npm_typescript-7rc",
+    ts_version = "7.0.1-rc",
+)
+use_repo(rules_ts_ext, "npm_typescript", "npm_typescript-56rc", "npm_typescript-7rc", "npm_typescript3")
 
 bazel_dep(name = "sub_external", version = "0.0.0", dev_dependency = True)
 local_path_override(
--- a/ts/private/npm_repositories.bzl
+++ b/ts/private/npm_repositories.bzl
@@ -1,7 +1,9 @@
 """Runtime dependencies fetched from npm"""
 
 load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")
-load("//ts/private:versions.bzl", "TOOL_VERSIONS")
+load("//ts/private:versions.bzl", "NATIVE_TYPESCRIPT_VERSIONS", "TOOL_VERSIONS")
+
+_TYPESCRIPT_NATIVE_PACKAGE_PREFIX = "@typescript/"
 
 def _http_archive_version_impl(rctx):
     integrity = None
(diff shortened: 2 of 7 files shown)


e2e/external_dep/MODULE.bazel
```diff
@@ -21,11 +21,10 @@ rules_ts_ext.deps(
 
 # Example of an rc version
 rules_ts_ext.deps(
-    name = "npm_typescript-56rc",
```
jbedard, line 24: Lets still test 56, or update to v6 if you want. But make your 7rc an addition instead of replacement for this test

ts/private/versions.bzl
```diff
@@ -7,7 +7,41 @@ Automatically mirrored using mirror_versions.sh and automated on GitHub Actions
...
+    for i in range(len(versions) - 1, -1, -1):
+        version = versions[i]
+        if "-" not in version:
+            return version
+    fail("No stable TypeScript versions are mirrored.")
+
+# Versions should be ascending order. Prerelease versions may appear after the
+# latest stable version so they can be mirrored without becoming the default.
+TYPESCRIPT_NATIVE_PACKAGE_INTEGRITIES = {
+    "7.0.1-rc": {
```
jbedard, line 11: Would it be any simpler to just put all the v7+ data into a new format? Don't touch `TOOL_VERSIONS` or `LATEST_TYPESCRIPT_VERSION` etc and just have a separate v7+/beta map?

Just a thought/idea 🤷

jbedard:
What platforms has this been tested on?

ts/private/versions.bzl
longlho: Done. Moved the TS 7 RC wrapper/native package data into a separate NATIVE_TYPESCRIPT_VERSIONS map, and TOOL_VERSIONS/LATEST_TYPESCRIPT_VERSION are back to stable releases only.

e2e/external_dep/MODULE.bazel
longlho: Done. Restored the 5.6 RC fixture and added TS 7 RC as an additional repo/target in both bzlmod and WORKSPACE e2e coverage.

longlho:
Platform testing update: I tested locally on macOS darwin-arm64 with both bzlmod and WORKSPACE e2e/external_dep runs, plus //ts/.... The Aspect Workflows matrix should exercise e2e/external_dep on Linux x64 after this push. The current repo CI does not run that e2e workspace on Windows.

ts/private/npm_repositories.bzl
```diff
@@ -45,12 +50,51 @@ def _http_archive_version_impl(rctx):
         # to update the integrity but the `ts_version` is already different.
         canonical_id = get_default_canonical_id(rctx, urls),
     )
+
+    native_package_integrities = native_typescript_version.get("native_package_integrities", {})
+    native_packages = _declared_native_packages(rctx)
```
jbedard, line 85: Can we avoid a lot of this if `not native_typescript_version`? I don't want to introduce the `rctx.read(rctx.path("package/package.json"))` unless we know we're using native packages...

jbedard: @longlho I'd like this stuff fixed up before merging, so unless you're using ts7 this PR should have zero side-effects.

Are you able to finish this up or should I take it from here?

longlho: Yeah feel free to take it!

jbedard (approved):

PR #946, opened 2026-07-09
perf: use tuple form of endswith in hot src-classification paths

jbedard:
_is_typings_src/_is_js_src/_is_ts_src run for every src of every ts_project during loading and analysis; a Starlark profile of a large monorepo shows endswith alone at ~0.5s of Starlark CPU. Use the tuple form of endswith so each check is one builtin call instead of up to four.

### Test plan

### DIFF
--- a/ts/private/ts_lib.bzl
+++ b/ts/private/ts_lib.bzl
@@ -227,11 +227,15 @@ def _relative_to_package(path, ctx):
         path = path.removeprefix(ctx.label.package + "/")
     return path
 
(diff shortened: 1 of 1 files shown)


acozzette (approved):

ts/private/ts_lib.bzl
```diff
@@ -227,11 +227,15 @@ def _relative_to_package(path, ctx):
         path = path.removeprefix(ctx.label.package + "/")
     return path
 
+_TYPINGS_EXTS = (".d.ts", ".d.mts", ".d.cts")
```
dzbarsky, line 230: nice one

PR #950, opened 2026-07-10
refactor!: validate ts_project attributes with `tsc --showConfig`

jbedard:
This expands the rules_ts 3.x typescript 7 `tsc --showConfig` logic (https://github.com/aspect-build/rules_ts/pull/969) to also apply to typescript <7 so we have a single implementation.

Currently (typescript 7.0.2) typescript 7 is not as strict as typescript 5,6 so an extra generated `{extends: "./realConfigjson", files: []}` is required for 5,6 - but we're doing it 100% of the time just for simplicity. A lot of the test cases added are to ensure this generated tsconfig works when in parent/children dirs, different repos etc.

Some previously unsupported/undetected tsconfig validation (config vs `ts_project` attribute) issues will now be caught such as invalid `ts_project` attributes based on the a tsconfig.json which extends from an npm package. The extended npm package may now be correctly taken into account.

### Test plan

- New test cases added

### DIFF
--- a/ts/private/ts_project_options_validator.cjs
+++ b/ts/private/ts_project_options_validator.cjs
@@ -1,75 +1,9 @@
 'use strict'
 exports.__esModule = true
 var path_1 = require('path')
-// Resolve the tsconfig "extends" chain to a single config, equivalent to what
-// `tsc --showConfig -p <tsconfig>` prints. Classic TypeScript (< 7) exposes the
-// compiler API to node so this runs in-process; native TypeScript (>= 7) ships no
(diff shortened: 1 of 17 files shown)


ts/private/ts_project_options_validator.cjs
```diff
@@ -142,7 +103,7 @@ function main(_a) {
     // The resolved config, in `tsc --showConfig` shape. The checks below read this
     // JSON only, so they run identically whether it was resolved in-process or by
```
acozzette: I think this comment is stale now that we no longer do the resolution in-process.

acozzette (approved):

PR #953, opened 2026-07-11
refactor!: drop WORKSPACE support, rules_ts is bzlmod-only

jbedard:
Deletes ts/repositories.bzl (rules_ts_dependencies and rules_ts_bazel_dependencies), all WORKSPACE / WORKSPACE.bzlmod files, and the bazel-7-wksp CI matrix leg. LATEST_TYPESCRIPT_VERSION moves to ts/extensions.bzl, the only symbol the bzlmod extension needed from repositories.bzl. The e2e/test bats harness now generates a MODULE.bazel scratch module instead of a WORKSPACE.

BREAKING CHANGE: rules_ts_dependencies / rules_ts_bazel_dependencies WORKSPACE setup is removed. Use bzlmod (MODULE.bazel).

BREAKING CHANGE: remove support for loading `aspect_rules_ts` via WORKSPACE

### Test plan


Aghassi:
Just to confirm this will make it a major version and not go into the 3.x?

We aren't on bzlmod yet sadly so just planning the upgrade path

jbedard:
Yes. I've officially created a 3.x branch, and this is the first thing for 4.x... I thought I'd tag you to encourage that bzlmod migration of yours... 😅

jbedard:
Note that 3.9.0 was tagged with initial tsgo/ts7 support, I also just tagged 3.9.1 before branching 3.x off.

Real tsgo support needs too many changes so I decided to start 4.x for that...

Aghassi:
Perfect thanks :D. I gotta got enable bzlmod and see what breaks

e2e/test/dir_as_source.bats
```diff
@@ -11,6 +11,9 @@ teardown() {
 
 @test 'ts_project with a directory in srcs should fail to build' {
 	workspace
+	cat >>MODULE.bazel <<EOF
```
acozzette, line 14: Is there any particular reason not to just handle this in the `workspace` function in common.bats?

acozzette (approved):

e2e/test/dir_as_source.bats
jbedard: I think because most tests shouldn't directly depend on bazel-lib?

PR #962, opened 2026-07-14
fix: validate native TypeScript with TypeScript 6

longlho:

Native TypeScript builds now run `ts_project` option validation through a companion repository containing an exact classic TypeScript 6 release. The native compiler repository keeps its public `:validator` label and delegates to that companion. `validator_ts_version` and `validator_ts_integrity` let consumers pin the validator without adding a second package to the compiler repository's `node_modules` tree.

### Test plan

- `bazel test --lockfile_mode=error --distdir=/private/tmp //ts/...`
- `bazel test --lockfile_mode=error --distdir=/private/tmp //:validator_version_test` from `e2e/external_dep`
- `bazel build --lockfile_mode=error --distdir=/private/tmp @npm_typescript-7_validator//:validator` from `e2e/external_dep`
- Full Bazel 7/8/9 workspace matrix in CI

### DIFF
--- a/MODULE.bazel.lock
+++ b/MODULE.bazel.lock
@@ -260,7 +260,7 @@
   "moduleExtensions": {
     "//ts:extensions.bzl%ext": {
       "general": {
-        "bzlTransitiveDigest": "WV+DuevmiFUeiSXIrDC1IEJov1WdDAy3pSFgQHHEFA0=",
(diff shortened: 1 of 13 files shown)


jbedard:
I hope this was accomplished with #969 - please let me know if anything is missing

PR #964, opened 2026-07-15
test: cover tsconfig options implied as computed defaults of other options

jbedard:
tsc materializes some tsconfig options as computed defaults of other options
(composite implies declaration/incremental, checkJs implies allowJs, and
moduleResolution bundler or module nodenext/preserve imply resolveJsonModule
depending on the TypeScript version). The ts_project attributes only mirror
what the tsconfig configures explicitly, so validation must accept these
without the corresponding attribute being set.

Also cover the rootDir/exclude guard accepting an "exclude" inherited through
the tsconfig extends chain, including with declaration_dir set, where tsc
materializes an implicit exclude entry for declarationDir whenever the user
does not author one.

### Test plan

- New test cases added


jbedard:
These are test cases uncovered while looking into making the tsconfig validation work with tsgo (https://github.com/aspect-build/rules_ts/pull/950 or https://github.com/aspect-build/rules_ts/pull/952 still TBD)

acozzette (approved):

PR #965, opened 2026-07-15
fix: declare extension_metadata for the ts module extension

jbedard:
See diffs, including MODULE.bazel.lock files, for what the the metadata accomplishes.

Declare module as `reproducible` and report `root_module_direct_*deps` bzlmod extension info.

### Test plan


e2e/bzlmodules/MODULE.bazel.lock
```diff
@@ -144,43 +144,6 @@
   },
   "selectedYankedVersions": {},
   "moduleExtensions": {
-    "@@aspect_rules_ts~//ts:extensions.bzl%ext": {
```
jbedard, line 147: This deletion is the expected result

e2e/bzlmodules/MODULE.bazel.lock
```diff
@@ -192,3558 +155,28 @@
...
-          ],
-          [
-            "aspect_tools_telemetry~",
-            "bazel_skylib",
-            "bazel_skylib~"
-          ]
-        ]
-      }
-    },
-    "@@rules_jvm_external~//:extensions.bzl%maven": {
```
jbedard, line 221: This and everything below I have no idea...

acozzette: I find that if I just remove this file and regenerate it on main, all this same stuff disappears. It's kind of surprising to me that this doesn't get updated for me unless I delete the file and thereby force the regeneration. But at least it seems unrelated to your change.

ts/extensions.bzl
```diff
@@ -57,6 +64,16 @@ To resolve this conflict, the root module should explicitly specify the desired
...
         )
 
+    # A repo requested by both dev and non-dev usages of the root module is a non-dev dep.
+    for name in root_direct_deps.keys():
+        root_direct_dev_deps.pop(name, None)
+
+    return module_ctx.extension_metadata(
+        root_module_direct_deps = root_direct_deps.keys(),
+        root_module_direct_dev_deps = root_direct_dev_deps.keys(),
+        reproducible = True,
```
acozzette, line 75: I think it would be worth adding a short comment saying why this is reproducible, which is that we require a checksum to be set either in `ts_integrity` or in `ts/private/versions.bzl`.

acozzette (approved):

e2e/bzlmodules/MODULE.bazel.lock
jbedard: If I remove it and regenerate it it is the exact same each time 🤷

Issue #483, opened 2023-11-12
[Q]: How to use rules_ts within local bazel_deb ?


farcop:
### What happened?

I have root workspace

root MODULE.bazel
```
bazel_dep(name = "submodule")
local_path_override(
    module_name = "submodule",
    path = "./submodule",
)

bazel_dep(name = "aspect_rules_ts", version = "1.4.5")

rules_ts_ext = use_extension("@aspect_rules_ts//ts:extensions.bzl", "ext")

rules_ts_ext.deps()

use_repo(rules_ts_ext, "npm_typescript", dev_dependency = True)
```

and submodule MODULE.bazel
```
module(name = "submodule")

bazel_dep(name = "aspect_rules_ts", version = "1.4.5")

rules_ts_ext = use_extension("@aspect_rules_ts//ts:extensions.bzl", "ext")

rules_ts_ext.deps()

use_repo(rules_ts_ext, "npm_typescript")
```

I cant build root ts
```
% bazel build //e2e-tests/...          
ERROR: Analysis of target '//e2e-tests:e2e-tests-compiled' failed; build aborted: no such package '@[unknown repo 'npm_typescript' requested from @]//': The repository '@[unknown repo 'npm_typescript' requested from @]' could not be resolved: No repository visible as '@npm_typescript' from main repository
INFO: Elapsed time: 5.504s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (2 packages loaded, 1 target configured)
```

if I set dev_dependency = True at submodule then it unable to build submodule from root module.

### Version

bazel 6.4.0
"aspect_rules_ts", version = "1.4.5"

### How to reproduce

```shell
See above
```

mvgijssel:
I'm having a similar problem with local repositories. When trying to build a ts_project rule from a nested repository from the root repository I'm asked the following:

```
You must choose exactly one of the following flags:

1. To choose the faster performance put this in /.bazelrc:

    # passes an argument `--skipLibCheck` to *every* spawn of tsc
    build --@aspect_rules_ts//ts:skipLibCheck=always
    fetch --@aspect_rules_ts//ts:skipLibCheck=always
    query --@aspect_rules_ts//ts:skipLibCheck=always

2. To choose more correct typechecks, put this in /.bazelrc:

    # honor the setting of `skipLibCheck` in the tsconfig.json file
    build --@aspect_rules_ts//ts:skipLibCheck=honor_tsconfig
[...]

PR #967, opened 2026-07-15
refactor!: validate ts_project attributes with `tsc --showConfig` via shell

jbedard:
A single validation action runs a shell script that writes a wrapper tsconfig next to the project's tsconfig (so relative paths and ${configDir} resolve exactly as the compile action does), resolves the extends chain with the default tsc (--project <wrapper> --showConfig), and compares the flattened result against the ts_project attributes with jq. No TypeScript compiler API involved, so the resolver can be swapped for a native binary such as tsgo (typescript >= 7).

Options tsc materializes as implied by other options (composite -> declaration/incremental, checkJs -> allowJs, module nodenext/preserve and moduleResolution bundler -> resolveJsonModule) are accepted without the matching attribute being set. Directory options that the attribute overrides on the tsc command line (outDir, declarationDir, tsBuildInfoFile, rootDir) may still be written relative to the tsconfig directory, while un-overridden ones must agree with the attribute in package-relative terms.

BREAKING CHANGES:
- ts_project(validator) and the @npm_typescript//:validator target are removed. Validation always resolves the tsconfig with the default tsc (a private attribute), so custom compilers such as ngc and workspaces that cannot resolve @npm_typescript themselves (#483) work unchanged.
- The rules_ts module extension now always creates @npm_typescript (lazily fetched) so rules_ts itself can reference it as a non-dev dependency, and rules_ts depends on jq.bzl for its jq toolchain.
- npm-package tsconfig extends now actually resolve during validation, which can surface real attribute mismatches that previously passed silently, and the extends target must be an input (ts_config(deps) or the tsconfig attribute).

### Test plan


jbedard:
With all validation now using `tsc --showConfig` and an eventual plan to invoke `tsc` without node I don't think this idea is worth it anymore.

PR #968, opened 2026-07-15
refactor: remove the underlying rule ts_project(extends) attribute and always pass ts_config()

jbedard:
### Test plan

### DIFF
--- a/ts/defs.bzl
+++ b/ts/defs.bzl
@@ -139,9 +139,13 @@ def ts_project(
 
             See [docs/tsconfig.md](/docs/tsconfig.md) for detailed information.
 
-        extends: Label of the tsconfig file referenced in the `extends` section of tsconfig
-            To support "chaining" of more than one extended config, this label could be a target that
-            provdes `TsConfigInfo` such as `ts_config`.
+        extends: Label of the tsconfig file referenced in the `extends` section of tsconfig.
+
+            Syntax sugar for a `ts_config` target: the `tsconfig` is wrapped in one declaring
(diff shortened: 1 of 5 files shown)


jbedard:
@codex

ts/defs.bzl
```diff
@@ -139,9 +139,14 @@ def ts_project(
...
-        extends: Label of the tsconfig file referenced in the `extends` section of tsconfig
-            To support "chaining" of more than one extended config, this label could be a target that
-            provdes `TsConfigInfo` such as `ts_config`.
+        extends: Label of the tsconfig file referenced in the `extends` section of tsconfig.
+
+            Syntax sugar for when `tsconfig` is a file or a dictionary: a file is wrapped in a
+            `ts_config` target with this label in its `deps`, while a dictionary additionally
+            gains an `extends` clause pointing at it.
+
+            When `tsconfig` is already a `ts_config` target, declare the extended configs in the
```
acozzette: I find this sentence a bit confusing, because I thought at first that it was telling me as the user what to do, in other words: "If tsconfig is already a ts_config target, then you're going to need to update its deps directly." But I think what it's really saying is: "If tsconfig is already a ts_config target, then this macro will wrap it in another tsconfig target so that it can declare the extended configs as deps."

acozzette (approved):

ts/defs.bzl
jbedard: Is the updated one any better?

acozzette: Yeah, that looks good!

PR #972, opened 2026-07-18
feat!: clearer bzlmod extension API for TypeScript toolchains

### LINKED ISSUE #941: use_extension second argument naming (ext) feels semantically unclear — is it required or just convention?
Hi maintainers 👋

While working with aspect_rules_ts, I noticed the following usage pattern in MODULE.bazel:

```starlark
rules_ts_ext = use_extension(
    "@aspect_rules_ts//ts:extensions.bzl",
    "ext",
    dev_dependency = True,
)

rules_ts_ext.deps(ts_version = "5.6.2")

use_repo(rules_ts_ext, "npm_typescript")
```

I have a question about the second argument "ext" in use_extension.

Is this "ext" value:

- a required convention enforced by Bazel or rules_ts
- or just an arbitrary identifier chosen by the user?

Specifically:

- Can this value be renamed to something more semantic like "ts", "typescript" or "rules_ts"?
- Or is "ext" required by design or internal implementation?
- Does changing it affect module resolution, repository naming, or use_repo behavior?

From a readability perspective, "ext" feels too generic in larger MODULE.bazel files, and a more explicit name like ts_ext = use_extension(...) would improve clarity.

However, I'm not sure whether this is safe or discouraged.

jbedard:
Rename the module extension surface so a MODULE.bazel reads self-documenting:
```
    ts = use_extension("@aspect_rules_ts//ts:extensions.bzl", "typescript")
    ts.toolchain(version = "5.6.2")   # or version_from = "//:package.json"
    use_repo(ts, "npm_typescript")
```
- Extension symbol `ext` -> `typescript`, matching rules_js's convention of naming the extension after its domain concept rather than a generic "ext".
- Tag class `deps` -> `toolchain`: the tag selects the TypeScript compiler version backing the toolchain; it does not declare dependencies.
- Attrs lose their redundant `ts_` prefix inside the `typescript.toolchain`
  context: `ts_version`/`ts_version_from`/`ts_integrity` ->
  `version`/`version_from`/`integrity`.

The default repo name `npm_typescript` and the multi-toolchain `name` escape hatch are unchanged. The internal `npm_dependencies` repo rule (shared with the WORKSPACE path) is untouched.

Fix #941

BREAKING CHANGES:

The bzlmod module extension API has changed. Update MODULE.bazel:

before
```
    rules_ts_ext = use_extension("@aspect_rules_ts//ts:extensions.bzl", "ext")
    rules_ts_ext.deps(ts_version = "5.6.2")
    use_repo(rules_ts_ext, "npm_typescript")
```
after
```
    ts = use_extension("@aspect_rules_ts//ts:extensions.bzl", "typescript")
    ts.toolchain(version = "5.6.2")
    use_repo(ts, "npm_typescript")
```

Renames:
- `use_extension(..., "ext")` -> `use_extension(..., "typescript")`
- `.deps(...)` -> `.toolchain(...)`
- `ts_version` -> `version`
- `ts_version_from` -> `version_from`
- `ts_integrity` -> `integrity`

The local variable bound to `use_extension` (commonly `rules_ts_ext`) is arbitrary and may be named anything; the examples now use `typescript`. The `npm_typescript` repo name and the `name` attribute for registering additional TypeScript versions are unaffected.

### Test plan


ts/extensions.bzl
```diff
@@ -71,18 +71,18 @@ To resolve this conflict, the root module should explicitly specify the desired
...
             "name": attr.string(default = "npm_typescript"),
-            "ts_version": attr.string(),
-            "ts_version_from": attr.label(),
-            "ts_integrity": attr.string(),
+            "version": attr.string(),
+            "version_from": attr.label(),
+            "integrity": attr.string(),
         }),
     },
 )
```
acozzette, line 88: Could we leave a line with `ext = typescript` here as an alias so that MODULE.bazel files can refer to either `ext` or `typescript`?

jbedard: When also dropping the `ts_` attribute prefixes I'm not sure if that's worth it? Especially when this is for 4.x and we can make breaking changes.

I did debate doing `typescript = ext` in 3.x though, just so people can start transitioning?

acozzette: Good point, I guess this won't help much by itself since we're also dropping the `ts_` prefixes.

I do think it would be good if possible to accept both the new and old syntaxes on 3.x. I've found in the past that if I require users to fix something simultaneously with an upgrade then it usually comes back to bite me. It would be nice if we could provide a way to be compatible with both 3.x and 4.x, so that modules could add support for the new version without breaking anyone who's still on the old one.

acozzette (approved):

ts/extensions.bzl
jbedard: We could try supporting both APIs on 3.x with a WARNING printed when using the deprecated one? Or at the very least we can do `ext = typescript` so they can migrate ONE part of the breaking change now in 3.x...

e2e/nested_repos/module_b/MODULE.bazel
```diff
@@ -2,10 +2,10 @@ module(name = "module_b")
 
 bazel_dep(name = "aspect_rules_ts", version = "0.0.0")
 
-rules_ts_ext = use_extension("@aspect_rules_ts//ts:extensions.bzl", "ext")
-rules_ts_ext.deps(
-    ts_version_from = "//:package.json",
+typescript = use_extension("@aspect_rules_ts//ts:extensions.bzl", "typescript")
+typescript.toolchain(
```
thesayyn: does this mean we have a typescript toolchain now?

jbedard: I guess it's not actually a toolchain, so we could potentially (for now) stick with the `.deps()` name? That also reduces the breaking change.

Regarding toolchains... see https://github.com/aspect-build/rules_ts/pull/951 - we can debate the correctness there, but maybe we should also wait till then to do this rename?

thesayyn: eh i am fine with breaking changes, i was just curious. i'd be happy to see https://github.com/aspect-build/rules_ts/pull/951 land honestly. so that we can use configurability and platforms and toolchains to configure typescript in the graph.

thesayyn (approved):

ts/extensions.bzl
acozzette: That sounds good to support both APIs on 3.x with a warning. I guess I don't see that much value if we do just `ext = typescript` by itself, though, because then there's still no way to be compatible with both 3.x and 4.x.

jbedard: How about https://github.com/aspect-build/rules_ts/pull/975 to start? I can then even merge that into `main` before landing this PR that makes the breaking change....

acozzette: #975 looks great to me, though I did leave one comment that I'm not sure `ext` and `typescript` will be treated as aliases the way they're currently set up in that PR.

acozzette (approved):

PR #975, opened 2026-07-22
feat: add `typescript` bzlmod extension API, deprecate `ext`

jbedard:
Introduce the clearer module extension API and plans for 4.x:
```
    typescript = use_extension("@aspect_rules_ts//ts:extensions.bzl", "typescript")
    typescript.deps(version = "5.6.2")   # or version_from = "//:package.json"
    use_repo(typescript, "npm_typescript")
```
- Add the `typescript` extension symbol alongside the retained `ext`.
- The `deps` tag gains unprefixed `version`/`version_from`/`integrity` attrs; the `ts_version`/`ts_version_from`/`ts_integrity` attrs stay as aliases. A tag may not mix the two sets.
- Using the `ext` symbol or any `ts_*` attr from the root module now prints a deprecation warning. The tag class stays named `deps`.

Migrate all in-repo usages (MODULE.bazel, examples, e2e, release_prep.sh) to the new API, and add the e2e/deprecated_ext workspace as backward-compat coverage for the deprecated names.

The old names will be removed in the next major release.

Ref #941

Users should migrate from the deprecated `ext` extension to the modern versions:
```
    typescript = use_extension("@aspect_rules_ts//ts:extensions.bzl", "typescript")
    typescript.deps(version = "5.6.2")   # or version_from = "//:package.json"
```

### Test plan

- New test cases added


ts/extensions.bzl
```diff
@@ -81,14 +123,32 @@ To resolve this conflict, the root module should explicitly specify the desired
...
+    }),
+}
+
+typescript = module_extension(
+    implementation = _typescript_impl,
+    tag_classes = _TAG_CLASSES,
+)
+
+# Deprecated: use `typescript` instead. Retained for backward compatibility.
 ext = module_extension(
```
acozzette, line 151: I wonder how this will work, because I would think that `typescript` and `ext` would now be treated as two different extensions rather than aliases for the same extension. E.g. if one module uses `ext` and another uses `typescript` then won't they be separate? I wonder if it would be better to just do `ext = typescript` and accept the downside that we won't be able to print a deprecation message.

acozzette (approved):

ts/extensions.bzl
jbedard: I think under the hood they both just do `npm_dependencies()` though?

jbedard: I guess there is no deprecation message printed when using `ext` instead of `typescript` though 🤔

acozzette: I'm still a bit confused about how this would work. If project A uses `typescript` and project B uses `ext` and they're both dependencies of the same root module, then couldn't they both refer to `@npm_typescript` but get different TypeScript versions?

jbedard: If they both have the exact `npm_typescript` name it would be an error, if they use different names... it's the same `npm_dependencies()` repo rules, just invoked via a different path. I think?

acozzette: Hmm, I guess I don't totally understand this but the fact that it's already an error to have multiple versions with the same name makes me think this should be ok. LGTM

Just out of curiosity, what is the recommended approach for a module to use if it contains a TypeScript library meant to be used by other modules? It seems like you have to specify a version, but if you use the default name `npm_typescript` then things will break if any other libraries choose a different version. Would the safe approach be to just choose a unique name (e.g. `npm_typescript_myproject`) and then it's not a big deal if different modules use different TypeScript versions?

acozzette (approved):

ts/extensions.bzl
jbedard: I think the safe thing is to create your own unique name, which does get scoped to your MODULE, then override `ts_project` to explicitly use that one so it won't fallback to the "default".

PR #977, opened 2026-07-30
refactor: capture isolated typecheck stderr via spawn_binary

acozzette:
Bazel requires every action to have at least one output, so the isolated typcheck action captures tsc's stderr into a marker file. This previously relied on JS_BINARY__STDOUT_OUTPUT_FILE, a private env var understood only by js_binary's launcher. Use bazel_lib's spawn_binary toolchain instead, the same mechanism run_binary's `stdout` and `stderr` attributes use. Requires bumping the bazel_lib minimum version to 3.5.0.

The reason we start capturing `stderr` instead of `stdout` is that `spawn_binary` will not print out any output that it is capturing. If type checking fails then we want to be able to know the reason, so it's important for `stdout` to printed out. `stderr` is not especially interesting, but Bazel requires every action to output something.

---

### Test plan

### DIFF
--- a/ts/private/ts_project.bzl
+++ b/ts/private/ts_project.bzl
@@ -13,6 +13,8 @@ load(":ts_config.bzl", "TsConfigInfo")
 load(":ts_lib.bzl", "COMPILER_OPTION_ATTRS", "OUTPUT_ATTRS", "STD_ATTRS", _lib = "lib")
 load(":ts_validate_options.bzl", _validate_lib = "lib")
 
+SPAWN_BINARY_TOOLCHAIN_TYPE = "@bazel_lib//lib:spawn_binary_toolchain_type"
+
 # Forked from js_lib_helpers.js_lib_helpers.gather_files_from_js_providers to not
 # include any sources; only transitive types & npm sources
 def _gather_types_from_js_infos(targets):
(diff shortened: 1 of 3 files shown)


ts/private/ts_project.bzl
```diff
@@ -401,5 +416,7 @@ ts_project = rule(
     """,
     implementation = lib.implementation,
     attrs = lib.attrs,
-    toolchains = COPY_FILE_TO_BIN_TOOLCHAINS,
+    toolchains = COPY_FILE_TO_BIN_TOOLCHAINS + [
+        config_common.toolchain_type(SPAWN_BINARY_TOOLCHAIN_TYPE, mandatory = False),
```
jbedard, line 473: Isn't it mandatory?

jbedard:
This should probably target `3.x` if we actually want people to use it anytime soon? `main` is currently `4.x` alpha and no one will use it for a while...

ts/private/ts_project.bzl
```diff
@@ -248,23 +250,36 @@ def _ts_project_impl(ctx):
...
+        # All we care about is whether this action succeeded or not, but Bazel
+        # requires every action to have an output. Wrap the invocation in
+        # bazel_lib's spawn_binary to capture stderr into an output file. We
+        # capture stderr rather than stdout so that tsc's diagnostics still
+        # reach the console when type-checking fails.
+        spawn_binary_toolchain = ctx.toolchains[SPAWN_BINARY_TOOLCHAIN_TYPE]
+        if not spawn_binary_toolchain:
+            fail("ts_project target {} requires the spawn_binary toolchain.".format(ctx.label))
+        wrapper_arguments = ctx.actions.args()
+        wrapper_arguments.add("--stderr", typecheck_output)
```
jbedard, line 316: Previously we captured stdout?

ts/private/ts_project.bzl
```diff
@@ -248,23 +250,36 @@ def _ts_project_impl(ctx):
...
+        wrapper_arguments.add("--")
+        wrapper_arguments.add(executable)
+
         ctx.actions.run(
-            executable = executable,
+            executable = spawn_binary_toolchain.spawn_binary_info.bin,
```
jbedard, line 321: It looks like the bazel-lib `spawn_binary` does spawn an entire extra process which we are trying to avoid... how else can we do this? :/

ts/private/ts_project.bzl
acozzette: Right, the reason I switched to stderr is that spawn_binary will not print out anything it's capturing in the event the action fails. So if type-checking fails and we're capturing stdout, the error will get swallowed. By capturing stderr instead, we can make sure a useful error message gets printed out.

ts/private/ts_project.bzl
acozzette: Claude seems to have set it up this way following the example of run_binary: https://github.com/bazel-contrib/bazel-lib/blob/1292cec119703023b452fb82087ebccb9a082e67/lib/private/run_binary.bzl#L181 We only run this action if we're doing just type checking without compiling, so depending on how you use `ts_project` you might not need the toolchain.

ts/private/ts_project.bzl
acozzette: I think the only way to avoid an extra process in the short term would be to keep handling this in our own launcher script. In the longer term this extra process should go away, though, because if we can move our stdout/stderr/exit_code/etc handling out of our launcher script and rely on spawn_binary instead, then that should allow us to drop our bash launcher and subtract one process there.

ts/private/ts_project.bzl
jbedard: Does that issue only exist when using `spawn_binary`?

jbedard: Don't we care about the type-check errors being outputted to the console though? Was this a bug previously where nothing was outputted and we should switch to stderr before this refactor?

acozzette: Yes, we don't seem to have it in our launcher script because it will apparently print out stdout if the action fails (even if it's capturing stdout).

PR #979, opened 2026-08-03
feat: support path mapping in ts_project when possible

acozzette:
Use `js_binary_lib.run_binary_action` when available so that the tsc emit action advertises `supports-path-mapping` and can share its cache across compilation modes.

We have to be prepared for the absence of `run_binary_action`, because it will only be available starting from rules_js v3.4.0. We cannot just upgrade our minimum rules_js dependency since that would be a breaking change.

---

Support path mapping in ts_project when possible

### Test plan

- New test cases added

### DIFF
--- a/e2e/path_mapping/MODULE.bazel
+++ b/e2e/path_mapping/MODULE.bazel
@@ -8,15 +8,7 @@ local_path_override(
     path = "../..",
 )
 
-# js_binary_lib.run_binary_action, which ts_project opportunistically uses to support path
-# mapping, has not yet been released. This pins to a commit on an unmerged aspect_rules_js
-# branch that has it, for testing purposes only.
-bazel_dep(name = "aspect_rules_js", version = "3.0.2")
-git_override(
-    module_name = "aspect_rules_js",
-    commit = "821502f7b994e8dd9c8fb614121478d1c25f97d9",
-    remote = "https://github.com/aspect-build/rules_js.git",
-)
+bazel_dep(name = "aspect_rules_js", version = "3.4.0")
 
 bazel_dep(name = "bazel_skylib", version = "1.8.1", dev_dependency = True)
 
--- a/e2e/path_mapping/MODULE.bazel
+++ b/e2e/path_mapping/MODULE.bazel
@@ -1,0 +1,27 @@
+"Basic e2e test for ts_project path mapping support"
+
+module(name = "e2e_path_mapping")
+
+bazel_dep(name = "aspect_rules_ts", version = "0.0.0")
+local_path_override(
+    module_name = "aspect_rules_ts",
+    path = "../..",
+)
+
+# js_binary_lib.run_binary_action, which ts_project opportunistically uses to support path
+# mapping, has not yet been released. This pins to a commit on an unmerged aspect_rules_js
+# branch that has it, for testing purposes only.
+bazel_dep(name = "aspect_rules_js", version = "3.0.2")
+git_override(
+    module_name = "aspect_rules_js",
+    commit = "821502f7b994e8dd9c8fb614121478d1c25f97d9",
+    remote = "https://github.com/aspect-build/rules_js.git",
+)
+
+bazel_dep(name = "bazel_skylib", version = "1.8.1", dev_dependency = True)
+
+typescript = use_extension("@aspect_rules_ts//ts:extensions.bzl", "typescript")
+typescript.deps(
+    version_from = "//:package.json",
+)
+use_repo(typescript, "npm_typescript")
--- a/ts/private/ts_lib.bzl
+++ b/ts/private/ts_lib.bzl
@@ -1,6 +1,8 @@
 "Utilities functions for selecting and filtering ts and other files"
 
+load("@aspect_rules_js//js:libs.bzl", "js_binary_lib")
 load("@aspect_rules_js//js:providers.bzl", "JsInfo")
+load("@bazel_skylib//lib:dicts.bzl", "dicts")
 
 # Attributes common to all TypeScript rules
 STD_ATTRS = {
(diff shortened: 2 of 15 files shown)


e2e/path_mapping/MODULE.bazel
```diff
@@ -0,0 +1,27 @@
...
+
+module(name = "e2e_path_mapping")
+
+bazel_dep(name = "aspect_rules_ts", version = "0.0.0")
+local_path_override(
+    module_name = "aspect_rules_ts",
+    path = "../..",
+)
+
+# js_binary_lib.run_binary_action, which ts_project opportunistically uses to support path
```
acozzette: Need to fix this before merging. Hopefully we can get the necessary feature out in rules_js v3.4.0.

ts/private/ts_lib.bzl
```diff
@@ -420,6 +422,26 @@ def _declare_outputs(ctx, paths):
         for path in paths
     ]
 
+# js_binary_lib.run_binary_action is only available in aspect_rules_js 3.4.0
+# and higher. Opportunistically use it when present, since bumping our minimum
+# supported aspect_rules_js version would be a breaking change.
+def _run_binary_action(ctx, env = None, execution_requirements = None, **kwargs):
```
jbedard, line 428: In the future (`main` branch / rules_ts 4.x) this method will be gone and we'll just have `js_binary_lib.run_binary_action(...)` in `ts/private/ts_project.bzl` and `ts/private/ts_validate_options.bzl`?

I guess putting this util here makes the diff in those 2 files align with `main` the most? 👍

jbedard:
LGTM but lets get a rules_js release out before landing this 👍

ts/private/ts_lib.bzl
acozzette: > In the future (`main` branch / rules_ts 4.x) this method will be gone and we'll just have `js_binary_lib.run_binary_action(...)` in `ts/private/ts_project.bzl` and `ts/private/ts_validate_options.bzl`?

Yes, that's what I was thinking. On 4.x things will be simpler since we can just rely on `js_binary_lib.run_binary_action()` being there.

> I guess putting this util here makes the diff in those 2 files align with `main` the most? 👍

Yes, I think so.

ts/private/ts_lib.bzl
```diff
@@ -420,6 +422,26 @@ def _declare_outputs(ctx, paths):
...
+    env = env or {}
+    execution_requirements = execution_requirements or {}
+    if hasattr(js_binary_lib, "run_binary_action"):
```
thesayyn, line 431: i like this. lets leave a TODO: in here to clean this up in the future when we bump the minimum version.

jbedard: FYI this PR is going into the rules_ts `3.x` branch. The cherry-pick into `main` will not have this. So I don't think we need to cleanup this PR at all.

But @acozzette please verify, and please do the cherry-pick right after merging this.

acozzette: Yep, once this PR lands I will cherry-pick it to main and make to sure to clean this part up.

ts/private/ts_project.bzl
```diff
@@ -281,37 +295,61 @@ See https://github.com/aspect-build/rules_ts/issues/361 for more details.
...
             typecheck_arguments.use_param_file("@%s", use_always = True)
             typecheck_arguments.set_param_file_format("multiline")
             typecheck_arguments.add("--bazelValidationFile", typecheck_output.short_path)
+
+            run_executable = executable
+            run_arguments = [typecheck_arguments]
+            run_env = {"BAZEL_BINDIR": ctx.bin_dir.path}
         else:
-            env["JS_BINARY__STDOUT_OUTPUT_FILE"] = typecheck_output.path
+            # Set BAZEL_BINDIR and JS_BINARY__STDOUT_OUTPUT_FILE using the hermetic `env`
```
acozzette: @jbedard I currently have a solution here using `env` to pass the environment variables, but based on your feedback on my other PR I'm guessing this isn't going to be your favorite solution.

Here are the different ways I think we could get this working:
1. Use `env` to pass the environment variables. This works and should be fairly efficient, but there's a fair amount of boilerplate. It also keeps relying on `JS_BINARY__STDOUT_OUTPUT_FILE` which we would eventually like to get rid of.
2. Use `spawn_binary`, which is what `run_binary` does to implement this feature. Possibly less efficient than env since it currently forks a new process, but it removes the dependency on the internal `JS_BINARY__` variable.
3. Add a new `--stdout` flag to the `js_binary` launcher and use that, perhaps combined with a `stdout` attribute on `js_binary_lib.run_binary_action()`.
4. Update this action to output an empty directory instead of a file. This is very simple because then we don't need to do anything, since an empty directory is a valid output. The only disadvantage I have found is that Google Cloud RBE apparently has a bug preventing it from handling empty output directories: https://github.com/bazelbuild/bazel/issues/28286

I think ideally we should remove the stdout-capturing functionality from our launcher script, so that makes me think options 1 and 3 are not great except maybe as a temporary solution. My favorite is option 4, since that one is the simplest and probably most efficient, and doesn't require anything from the launcher script. It's unfortunate that Google Cloud RBE has a bug affecting it, but presumably they're going to fix it at some point since it's affecting other projects.

jbedard: For 4... doesn't _something_ still need to create the directory though? Even if empty something has to create it...?

acozzette: No, experimentally I have found that you don't have to do anything. I would have to double check but I believe Bazel creates the directory before even running the action.

acozzette: I updated this to get rid of the use of `coreutils env` and  just create an empty directory.

jbedard (approved):

PR #984, opened 2026-08-28
fix: emit source dts files in ts_project output

jbedard:

Source `ts_project(srcs)` dts files are now passed through to the resulting `JsInfo` as-is with no modifications or outDir/rootDir applied just like `tsc` does not emit them while emitted dts may still depend on them.

### Test plan

- New test cases added

### DIFF
--- a/examples/dts_srcs/BUILD.bazel
+++ b/examples/dts_srcs/BUILD.bazel
@@ -1,0 +1,65 @@
+"""ts_project(srcs) containing .d.ts files with various outDir/rootDir settings.
+
+A .d.ts src is never transpiled or relocated by tsc, so it is passed through to
+JsInfo(types) (the `types` output group) at its source location.
+"""
+
+load("@aspect_rules_ts//ts:defs.bzl", "ts_project")
+load("@bazel_lib//lib:testing.bzl", "assert_outputs")
+
+ts_project(
+    name = "default",
+    srcs = [
+        "index.ts",
+        "types.d.ts",
+    ],
+    declaration = True,
+)
+
+assert_outputs(
+    name = "default_types_test",
+    actual = ":default_types",
+    expected = [
+        "dts_srcs/index.d.ts",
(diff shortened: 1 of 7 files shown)


jbedard (commented):
Reviewed the `.d.ts` passthrough and output-path handling. No blocking issues found.

Verified with:
- `bazel test //dts_srcs:all`
- `bazel test //dts_pkg-outDir:all`

Both pass from the `examples` workspace.

jbedard:
Summary:
> Consequences match tsc too:
>- dist/index.d.ts re-exporting ./types dangles — identical to tsc (verified with both compilers earlier).
>- A consumer importing the .d.ts at its source path works — identical to tsc.
>- npm_package ships the .d.ts at its source path — same as npm publish with files covering it.

acozzette (approved):

jbedard:
@acozzette WDYT of this? Technically I think this is "more aligned with tsc", but it sure is odd 🤷

jbedard:
Where things like `.js` when `allowJson` or `.json` when `resolveJsonModule` do have rootDir+outDir applied and get fully copied/outputted to bin...

jbedard:
FWIW I'm trying to make `oxc_transpile` do _similar_ to what rules_ts/tsc does: https://github.com/aspect-build/rules_oxc/pull/48

acozzette:
I feel like I'm not familiar enough with the situation to have a very informed opinion. But if the choice is between copying the file and not copying it, doing less for now seems like a good call.

jbedard:
`tsc` does not copy it, so I think we should do the same. However outside bazel that original `.d.ts` in your source is still available - which is where this PR comes in, should the original `.d.ts` source file be put into the `JsInfo.types`? That's what this PR adds...

acozzette:
I think that makes sense. If we're already doing it in the common case where `out_dir` matches `root_dir` then it seems reasonable to do that when they're not equal as well.

PR #986, opened 2026-09-19
ci: connect CI to Aspect Cloud

joshcbarnes:
Points both CI workflows at Aspect Cloud's remote cache and BES. They run on stock GitHub-hosted runners, which have no cache of their own.

Follows the pattern from rules_lint [#1003](https://github.com/aspect-build/rules_lint/pull/1003) and [#1006](https://github.com/aspect-build/rules_lint/pull/1006): `aspect setup bazelrc` generates the config for vanilla `bazel`, `.aspect/config.json` supplies it for aspect-cli builds.

### The two halves

**aspect-cli jobs** (`ci-workflows.yaml`) get `--remote`. That needs endpoints, and neither side supplies them on its own: setup-aspect's `aspect auth login --with-api-token` records a credential but no endpoints, and Aspect Cloud's built-in entry ships with none. They come from a committed `.aspect/config.json`, found via `$ASPECT_WORKSPACE` — pinned to the checkout root, since the test matrix builds from the nested workspaces. That variable selects only the deployment list, never which `.aspect/config.axl` a task loads.

**vanilla-bazel jobs** (`ci-vanilla-bazel.yaml`) get `--config=aspect-cloud` from a committed `.aspect/bazelrc`, which the root and all seven nested workspaces `try-import` at their own `../` depth. `aspect` goes on PATH there too — the rc authenticates through the `aspect` credential helper and stamps the invocation via `aspect setup workspace-data`.

Both flags are empty on a fork PR, which gets no org secrets: the login never runs, so the remote would be reached with no credential.

### Why `.aspect/generated/` appears in eight places

The rc's `aspect-exec-log` group writes `%workspace%/.aspect/generated/exec.log.zstd`. The nested workspaces try-import the root rc, so `%workspace%` is *theirs*, and Bazel fails the build rather than create the directory:

```
ERROR: Error initializing execution log: .../exec.log.zstd (No such file or directory)
```

So each workspace gets the `.gitignore` that `aspect setup bazelrc` generates at the root. rules_lint doesn't hit this — its pinned rc predates the group.

Note this group only ever serves vanilla `bazel`. An `aspect <task>` invocation runs with `--ignore_all_rc_files` and passes its own `--execution_log_compact_file=/tmp/aspect-bazel-<uuid>/exec.log.zstd`, so both halves upload a compact exec log, by two different mechanisms.

### Why the version bump

The pin was 2026.28.11, which predates the deployment feature — no `--remote`, no `aspect auth status`/`configure`. Deployments landed in 2026.38.14 and `aspect setup workspace-data` in 2026.38.24. `--task-key` is deprecated there, so the three call sites move to `--task:name`.

The secret moves to `ASPECT_OSS_API_TOKEN`, matching rules_lint — that's the account with cache/BES write access for the org.

### Deltas from generated output

`.aspect/bazelrc` is `aspect setup bazelrc` output with two edits, both of which need to survive a regeneration:

- a local single-tenant dev deployment's section removed
- `--workspace_status_command="aspect setup workspace-data"` added to the `aspect-cloud` section, so a vanilla invocation is attributed in the Web UI the way `aspect <task>` is. It names a command rather than carrying data, so it's the same string for every job and commit — a constant, which is what an rc is for — while Bazel re-runs it every invocation.

### Verified locally

From `e2e/path_mapping` with `ASPECT_WORKSPACE` at the repo root — the CI shape:

- `--announce_rc` shows `--config=aspect-cloud` resolving out of the root `.aspect/bazelrc`, with the quoted status command parsed as a single flag.
- Status keys land in `volatile-status.txt`, not `stable-` — a `STABLE_` key would invalidate every stamped action whenever the commit moves.
- Exec log written to `e2e/path_mapping/.aspect/generated/exec.log.zstd` (10,377 bytes) and **uploaded**, per the BEP's `buildToolLogs`:
  ```
  execution_log.binpb.zst -> bytestream://cache.aspect.build/blobs/06cdc9be90a11985ea4ec...
  command.profile.gz      -> bytestream://cache.aspect.build/blobs/609d017f074e418763b5a...
  ```
- `aspect test --remote` on a clean output base: 3.0s, 31 actions, 12 remote cache hits, 56 BES events streamed.
- `aspect format` passes under the bumped CLI, confirming `.aspect/config.axl` still loads on 2026.38.24.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


.aspect/generated/.gitignore
```diff
@@ -0,0 +1,2 @@
+*
```
jbedard, line 1: Can't we just put `**/.aspect/generated` in only the root `.gitignore`?

joshcbarnes: This is actually kind of load-bearing. By putting this inside the generated directory, it ensures the directory exists in git so that exec logs can be put there

jbedard: Oh I see, the exec log flag must point to a pre-existing directory? Aren't these temp files that we want cleaned up and not persisted though? I'd think `$TMP` would be a better place to put them...

.aspect/bazelrc
```diff
@@ -0,0 +1,40 @@
...
+common:aspect-bes --grpc_keepalive_timeout=30s
+common:aspect-bes --generate_json_trace_profile
+common:aspect-bes --noslim_profile
+common:aspect-bes --experimental_profile_include_target_label
+common:aspect-bes --experimental_profile_include_primary_output
+common:aspect-bes --legacy_important_outputs
+common:aspect-bes --remote_build_event_upload=all
+
+# The compact execution log, uploaded with the build diagnostics for per-action data in the Web UI. Slow on a thin link: remove this group and the --config lines naming it to skip it.
+common:aspect-exec-log --execution_log_compact_file=%workspace%/.aspect/generated/exec.log.zstd
```
jbedard: Where else does this path get referenced from? I always thought the aspect-cli would do this and then we wouldn't need all the `.gitignore` stuff here - so is this just so we can get execlog data without using the cli? That would also be a reason we can't hide this within the cli...

jbedard (approved):

